PR #8, opened 2017-10-11
Initial translations support

ilkkao:
Adds the files needed to enable translations. The actual translation JSON files will be added separately when they are ready.


manifest.json
```diff
@@ -3,46 +3,47 @@
   "author": "Zendesk",
   "version": "1.0.1",
   "api_version": 1,
+  "default_locale": "en-us",
   "settings": [{
-    "label": "Colors",
+    "label": "$colors_group_label",
```
ilkkao: value is considered as a translation key if it starts with a `$` character.

PR #15, opened 2018-01-18
Updatign the article template to use the new edited_at property

rmcgrew:
Updating the article template for the new edited_at property

@ilkkao @aliciacatalina

### DIFF
--- a/templates/article_page.hbs
+++ b/templates/article_page.hbs
@@ -44,10 +44,10 @@
             {{/link}}
 
             <ul class="meta-group">
-              {{#is article.created_at article.updated_at}}
+              {{#is article.created_at article.edited_at}}
                 <li class="meta-data">{{date article.created_at timeago=true}}</li>
               {{else}}
-                <li class="meta-data">{{date article.updated_at timeago=true}}</li>
+                <li class="meta-data">{{date article.edited_at timeago=true}}</li>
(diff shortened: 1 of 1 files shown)


aliciacatalina:
Do we need to update documentation too? https://developer.zendesk.com/apps/docs/help-center-templates/objects#article-object

luis-almeida:
I can see documentation has been updated in the same PR where we added support for `edited_at`:
https://github.com/zendesk/help_center/pull/15229/files#diff-081d6c4a1938144617b42c40ed5fe354R49

I believe this hasn't been reviewed and published in the developer website though.
Definitely a good catch by @aliciacatalina. Would be nice to have this new prop visible in the docs.

aliciacatalina:
So before, with the updated_at an updated article would show like this:
screen shot 2018-01-18 at 10 37 58
 The date it was updated at, and the string updated, because we have a conditional in the template, if I apply the change in this PR to the template, the result is:
screen shot 2018-01-18 at 10 42 43
Because the `edited_at` and the `created_at` are the same, is this what we want to display?
@rmcgrew

rmcgrew:
If the translation hasn't been edited, then yes. So I would guess the update 1 year ago was something like a vote, etc. that wouldn't be considered an edit and we'd still show that original date.

Here's what I see for an article that I've edited. 
screen shot 2018-01-18 at 10 48 47

PR #22, opened 2018-05-17
Add a script to update theme translations files

ilkkao:
A simple script to update the translated strings in the theme. Uses Zendesk locales API: https://developer.zendesk.com/rest_api/docs/core/locales#content

@zendesk/delta


bin/update-translations.js
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env node
+
+const fetch = require('node-fetch');
+const path = require('path');
+const fs = require('fs');
+
+const LOCALE_ENDPOIINT = `https://support.zendesk.com/api/v2/locales`;
```
luis-almeida: Nit typo in `LOCALE_ENDPOINT`.

ilkkao: thanks!

PR #32, opened 2018-07-12
Fix typos in translation keys

vecerek:
This PR marks the affected keys and creates new key with the fix.

@zendesk/delta @zendesk/i18n


siblackzen:
👍 LGTM!

PR #49, opened 2018-09-07
Listen to input on textarea to enable request submit button

clong-:
### Description
In the /requests/{id} interface, when pasting text into the comment text area using the mouse (via context menu or menu bar), the submit button does not enable. Likewise, when cutting or deleting all text from the comment box using the mouse, the submit button does not disable.

This is due to the current listener on the field only binding on `keyup`. Since mouse actions don't fire a keyup event, the interface doesn't respond to the text being entered/removed as we might intend.

This PR adjusts the default listener to bind on `input`, which should fire whenever the value of the field changes, not just on keyboard events.

### Tasks
- [x] :green_book:
- [x] :+1:

### Risks
- Low

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.4.2",
+  "version": "1.4.3",
   "api_version": 1,
   "default_locale": "en-us",
   "settings": [{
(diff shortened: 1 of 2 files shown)


clong-:
Thanks @vecerek!

clong-:
Oh, it looks like I am not authorized to merge:


How shall I proceed?

vecerek:
@clong- we can take over from here and merge it for you tomorrow morning and we'll propagate the change accordingly to TC and HC (which I see you already opened a PR for). If it's fine by you 🙂

clong-:
@vecerek Sounds good, thanks!

PR #50, opened 2018-09-18
Remove checks for internal sections

jbdietrich:
These checks no longer make sense with the advent of article view permissions (they will always return `false`).

@zendesk/delta

### DIFF
--- a/style.css
+++ b/style.css
@@ -922,38 +922,41 @@ ul {
 
 .promoted-articles-item {
   flex: 1 0 auto;
+  border-bottom: 1px solid #ddd;
+  padding: 15px 0;
 }
 
 @media (min-width: 1024px) {
   .promoted-articles-item {
     align-self: flex-end;
     flex: 0 0 auto;
-    padding-right: 30px;
+    margin-right: 30px;
     width: 33%;
     /* Three columns on desktop */
   }
   [dir="rtl"] .promoted-articles-item {
-    padding: 0 0 0 30px;
+    margin: 0 0 0 30px;
   }
 }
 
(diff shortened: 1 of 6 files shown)


jbdietrich:
@luis-almeida this change was prompted by another PR to change the documentation: https://github.com/zendesk/help_center/pull/16703. I think we still need the CSS etc... because it is possible to mark an article as `internal`.

luis-almeida:
Perfect! Sorry I missed that PR.

augustocravosilva:
You can actually remove this CSS https://github.com/zendesk/copenhagen_theme/blob/master/styles/_category.scss#L29-L31

Don't forget to https://github.com/zendesk/copenhagen_theme#requirements

jbdietrich:
@augustocravosilva I think it is still possible to check `internal` on the articles from within the section tree. Do we still want to remove that?

augustocravosilva:
I think we should remove it, and when https://zendesk.atlassian.net/browse/PUB-1102 is implemented it might be re-added if needed. The style might not apply when the icon is placed next to an article title vs. section title.

csk157:
I added a couple updates in the last commit:
- added lock icons to the articles in home, category and section pages
- moved the border for the promoted articles from the anchor tag to the list item (so that it can contain the lock icon)

screen shot 2018-10-05 at 12 31 25
screen shot 2018-10-05 at 12 31 34
screen shot 2018-10-05 at 12 31 05

csk157:
@luis-almeida @augustocravosilva could you take a look again? I got around to marking the articles with a lock (latest commit).

PR #52, opened 2018-10-09
 Mark optional form fields in the New Post form

yara-simargil:
Following https://github.com/zendesk/help_center/pull/16923

In order to support a11y (for cognitive disability in this case) we need to replace asterisk with marking optional fields.

@zendesk/delta
cc/ @zendesk/piratos

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.4.4",
+  "version": "1.4.5",
   "api_version": 1,
   "default_locale": "en-us",
   "settings": [{
--- a/styles/_forms.scss
+++ b/styles/_forms.scss
@@ -69,7 +69,7 @@
 }
 
 .form-field .optional {
-  color: #bbb;
+  color: $secondary-text-color;
   margin-left: 4px;
 }
 
--- a/styles/_forms.scss
+++ b/styles/_forms.scss
@@ -68,6 +68,11 @@
   margin-left: 2px;
 }
 
+.form-field .optional {
+  color: #bbb;
+  margin-left: 4px;
+}
+
 .form-field p {
   color: $secondary-text-color;
   font-size: 12px;
--- a/style.css
+++ b/style.css
@@ -364,6 +364,11 @@ ul {
   margin-left: 2px;
 }
 
+.form-field .optional {
+  color: lighten($text_color, 20%);
+  margin-left: 4px;
+}
+
 .form-field p {
   color: lighten($text_color, 20%);
   font-size: 12px;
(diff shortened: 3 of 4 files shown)


styles/_forms.scss
```diff
@@ -68,6 +68,11 @@
   margin-left: 2px;
 }
 
+.form-field .optional {
+  color: #bbb;
```
anpa: not sure if this should be `$border-color`?

https://github.com/zendesk/copenhagen_theme/blob/master/styles/_variables.scss#L23

yara-simargil: why should it? that's a different colour - `#bbb` not `#ddd`. and I'd say it's pretty hard to read the text when using lighter one.

anpa: Argh, sorry I am not reading properly today. You're right! I thought I read `$border-color: #bbb` 🙈

liambutler: Does this mean that the text for 'optional' will be #BBB? If so, it won't have sufficient contrast to meet the accessibility requirement.

According to https://webaim.org/resources/contrastchecker/, #BBB against a white background gives a contrast of 1.92. For text of this size, it would need to have a contrast of at least 4.5

yara-simargil: Thanks for bringing this! It somehow fell out of my sight. It's indeed painful to read it with `#bbb`.
We could use `#68737D` from Garden here, its the closest one passing the 4.5 limit: 
screen shot 2018-10-11 at 11 50 51
Though in this case it is almost indistinguishable from label text. How do you think?
@milenaalarsen

milenaalarsen: From our typography guidelines, we are encouraged to use #2F3941 for primary/ standard and #68737D for secondary. 

When a field is disabled and the user can't write inside the form, the font color inside is #C2C8CC. But we don't have anything in place for disable titles.  Does this help?

yara-simargil: Thank you @augustocravosilva, I didn't know that! I think we should go with `$secondary-text-color`. And maybe we should consider switching Help Center to use proposed Garden colours but that's out of scope for this PR. I'll update the colour in all forms.

augustocravosilva: Sounds good :)
Please run `bin/compile.rb` to produce the final CSS

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.4.3",
+  "version": "1.4.4",
```
augustocravosilva: this version already exists

yara-simargil: Rebased and bumped to `v1.4.5`. Thanks!

yara-simargil:
@augustocravosilva @anpa Is this good to go or do we need to change anything else?

PR #57, opened 2018-10-18
Replace show comment container and mark as resolved with native buttons to support a11y

yara-simargil:
Following https://github.com/zendesk/help_center/pull/17006

Users could not leave comments on requests using keyboard. In order to provide native keyboard navigation we should replace `div` and `a` tags with `button`.

@anpa @luis-almeida @zendesk/delta 

Jira: https://zendesk.atlassian.net/browse/HCJ-1715

#### Risks
Low - small ui update

### DIFF
--- a/templates/request_page.hbs
+++ b/templates/request_page.hbs
@@ -75,11 +75,9 @@
         </div>
 
         <div class="comment-container">
-          <div class="comment-show-container {{#validate 'body'}}hidden{{/validate}}">
-            <span class="comment-show-container-content">
-              {{t 'add_to_conversation'}}
-            </span>
-          </div>
+          <button type="button" class="comment-show-container {{#validate 'body'}}hidden{{/validate}}">
+            {{t 'add_to_conversation'}}
+          </button>
 
           <div class="comment-fields {{#validate 'body'}}shown{{/validate}}">
             {{#if help_center.request_ccs_enabled}}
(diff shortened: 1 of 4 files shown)


templates/request_page.hbs
```diff
@@ -75,11 +75,9 @@
         </div>
 
         <div class="comment-container">
-          <div class="comment-show-container {{#validate 'body'}}hidden{{/validate}}">
-            <span class="comment-show-container-content">
-              {{t 'add_to_conversation'}}
-            </span>
-          </div>
+          <button type="button" class="comment-show-container {{#validate 'body'}}hidden{{/validate}}">
```
anpa: just noticed, why `type="button"` if it's already a `<button />`?

yara-simargil: Good question! Because by default type will be set to `submit` which will.. well, submit the form and therefore reload the page.

yara-simargil:
Yes, thank you! 🙏  I'll update it.

templates/request_page.hbs
anpa: Ahhh... interesting!

templates/request_page.hbs
```diff
@@ -101,11 +99,11 @@
             {{checkbox 'mark_as_solved'}}
 
             {{#if request.can_be_marked_as_solved}}
-              <a role="button" class="button-large button-secondary mark-as-solved"
+              <button type="button" class="button-large button-secondary mark-as-solved"
```
anpa: wait, but this one is a submit button, right?

yara-simargil: No, we do submission manually in the handler after we set `checked` and `disabled`. I didn't change that logic.

anpa: gotcha! thanks!

PR #59, opened 2018-11-15
Fix promoted articles styling

csk157:
Changes in my previous PR didn't behave exactly as expected for promoted articles. This should fix the columns.

screenshot 2018-11-15 at 09 55 12

screenshot 2018-11-15 at 10 17 06

It switches to using padding instead of margin (`33%` width doesn't behave well with margins) and wraps the content of the item in a `div`, which would actually have the border (I was considering adding a special class name for it, but we don't have it for `a` either 🤷‍♂️)

@zendesk/delta

### Edit
Updated removed the `div` and made `a` to wrap the icon. Also applied the similar styling to the `a` itself, also made it a block.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.4.8",
+  "version": "1.4.9",
   "api_version": 1,
   "default_locale": "en-us",
   "settings": [{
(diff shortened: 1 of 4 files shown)


csk157:
Makes sense, I will try some other approach.

P.S. it doesn't seem to behave that way in category or section pages where articles are listed. Only the text is clickable. I will take a note to revisit those pages

csk157:
Updated the PR to use `a` as a container, now the whole row is clickable

PR #63, opened 2019-01-14
Refactor jQuery to vanilla JS

lovelovedokidoki:
We want to not be dependent on jQuery anymore, therefore we must refactor to vanilla JS.
Please comment if you can see that the proposed code can be simplified further without using browser support.

* JIRA: [GG-249: Refactor script.js file in CPH theme to use pure vanilla JS](https://zendesk.atlassian.net/browse/GG-249)

cc @zendesk/guide-growth 

#### Risks:
The script is responsible for a number of important interactions in the UI. In order to move away from jQuery, a special version of Element.prototype.closest is added to support IE11 which doesn't have that prototype yet. (.closest() used to be just a jQuery feature, but was added to most modern browsers https://caniuse.com/#feat=element-closest)

Due to these risks, following features should be manually tested on all supported browsers before a merge

##### Features on article pages:

_Social Share buttons_
Script contains logic to open an iframe

_Comments area_
Script expands comment textarea and shows relevant buttons to submit comments when user clicks on comment textbox

##### Features on requests pages:

_Submit requests_
 Script sets marked as solved form data before submitting when ‘Mark as solved’ button is clicked or ‘Marked as solved and submit’ button is clicked

_Request comments area_
Script expands comment textarea and shows relevant buttons to submit comments when user clicks on comment textbox. Also changes text of ‘Marked as solved’ button and shows/disables ‘Submit’ button dependent on whether input is empty or not

_Reopen request_
When a request has been solved, end-user should still be able to submit a new comment (thus reopening the request). Script is responsible for enabling the submit button when input field is not empty. If script fails at doing this, the user can't reopen the request.

_Collapsible sidebar (Ticked details sidebar)_
This sidebar is by default collapsed when render size is mobile screen sizes. Script toggles it upon clicking.

_Multiple organizations_
When the end-user is a member of multiple organisations (and this has to be enabled in Support), they see a dropdown in the tickets details sidebar. Script is responsible for catching changes to this drowdown and submit them.

_Search bar in ‘My activities’ / ‘My requests’ page_
 Script listens for enter key keypress to filter results based on the search string

##### Features in navbar menu (all pages):

Accessibility features such as listening for enter key and escape key to open / collapse menus. This applies to the hamburger menu which appears on responsive layout for mobile screen sizes.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,5 +1,21 @@
-
 document.addEventListener('DOMContentLoaded', function() {
+
+  function matches (element, selector) {
+    return Element.prototype.matches && element.matches(selector)
+      || Element.prototype.msMatchesSelector && element.msMatchesSelector(selector)
+      || Element.prototype.webkitMatchesSelector && element.webkitMatchesSelector(selector);
+  }
+
+  function closest (element, selector) {
+    if (Element.prototype.closest) {
+      return element.closest(selector);
+    }
+    do {
+      if (matches(element, selector)) return element;
+      element = element.parentElement || element.parentNode;
+    } while (element !== null && element.nodeType === 1);
+    return null;
+  }
 
   // social share popups
   [].forEach.call(document.querySelectorAll('.share a'), function(anchor) {
@@ -50,10 +66,8 @@ document.addEventListener('DOMContentLoaded', function() {
     requestMarkAsSolvedCheckbox.setAttribute('checked', true);
     requestCommentSubmitButton.disabled = true;
     this.setAttribute('data-disabled', true);
-    // Element.closest is supported in modern browsers and has a polyfill available for still in use outdated versions
-    // See https://caniuse.com/#feat=element-closest
-    // TODO: Should polyfill be installed? 
-    this.closest('form').submit();
+    // Element.closest is not supported in IE11
+    closest(this, 'form').submit();
   });
 
   // Change Mark as solved text according to whether comment is filled
--- a/script.js
+++ b/script.js
@@ -1,129 +1,159 @@
-/*
- * jQuery v1.9.1 included
- */
 
-$(document).ready(function() {
+document.addEventListener('DOMContentLoaded', function() {
 
   // social share popups
-  $(".share a").click(function(e) {
-    e.preventDefault();
-    window.open(this.href, "", "height = 500, width = 500");
+  [].forEach.call(document.querySelectorAll('.share a'), function(anchor) {
+    anchor.addEventListener('click', function(e) {
+      e.preventDefault();
+      window.open(this.href, '', 'height = 500, width = 500');
+    });
   });
 
   // show form controls when the textarea receives focus or backbutton is used and value exists
-  var $commentContainerTextarea = $(".comment-container textarea"),
-    $commentContainerFormControls = $(".comment-form-controls, .comment-ccs");
+  var commentContainerTextarea = document.querySelector('.comment-container textarea'),
+    commentContainerFormControls = document.querySelector('.comment-form-controls, .comment-ccs');
 
-  $commentContainerTextarea.one("focus", function() {
-    $commentContainerFormControls.show();
-  });
+  if (commentContainerTextarea) {
+    commentContainerTextarea.addEventListener('focus', function focusCommentContainerTextarea() {
+      commentContainerFormControls.style.display = 'block';
+      commentContainerTextarea.removeEventListener('focus', focusCommentContainerTextarea);
+    });
 
-  if ($commentContainerTextarea.val() !== "") {
-    $commentContainerFormControls.show();
+    if (commentContainerTextarea.value !== '') {
+      commentContainerFormControls.style.display = 'block';
+    }
   }
 
   // Expand Request comment form when Add to conversation is clicked
-  var $showRequestCommentContainerTrigger = $(".request-container .comment-container .comment-show-container"),
-    $requestCommentFields = $(".request-container .comment-container .comment-fields"),
-    $requestCommentSubmit = $(".request-container .comment-container .request-submit-comment");
+  var showRequestCommentContainerTrigger = document.querySelector('.request-container .comment-container .comment-show-container'),
+    requestCommentFields = document.querySelectorAll('.request-container .comment-container .comment-fields'),
+    requestCommentSubmit = document.querySelector('.request-container .comment-container .request-submit-comment');
 
-  $showRequestCommentContainerTrigger.on("click", function() {
-    $showRequestCommentContainerTrigger.hide();
-    $requestCommentFields.show();
-    $requestCommentSubmit.show();
-    $commentContainerTextarea.focus();
-  });
+  if (showRequestCommentContainerTrigger) {
+    showRequestCommentContainerTrigger.addEventListener('click', function() {
+      showRequestCommentContainerTrigger.style.display = 'none';
+      [].forEach.call(requestCommentFields, function(e) { e.style.display = 'block'; });
+      requestCommentSubmit.style.display = 'inline-block';
+
+      if (commentContainerTextarea) {
+        commentContainerTextarea.focus();
+      }
+    });
+  }
 
   // Mark as solved button
-  var $requestMarkAsSolvedButton = $(".request-container .mark-as-solved:not([data-disabled])"),
-    $requestMarkAsSolvedCheckbox = $(".request-container .comment-container input[type=checkbox]"),
-    $requestCommentSubmitButton = $(".request-container .comment-container input[type=submit]");
+  var requestMarkAsSolvedButton = document.querySelector('.request-container .mark-as-solved:not([data-disabled])'),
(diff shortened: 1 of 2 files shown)


lovelovedokidoki:
ℹ️Manual testing has been done with the newest version of Chrome so far, but still in progress for other browsers

script.js
```diff
@@ -1,129 +1,159 @@
...
+    requestMarkAsSolvedCheckbox = document.querySelector('.request-container .comment-container input[type=checkbox]'),
+    requestCommentSubmitButton = document.querySelector('.request-container .comment-container input[type=submit]');
+
+  requestMarkAsSolvedButton && requestMarkAsSolvedButton.addEventListener('click', function () {
+    requestMarkAsSolvedCheckbox.setAttribute('checked', true);
+    requestCommentSubmitButton.disabled = true;
+    this.setAttribute('data-disabled', true);
+    // Element.closest is supported in modern browsers and has a polyfill available for still in use outdated versions
+    // See https://caniuse.com/#feat=element-closest
+    // TODO: Should polyfill be installed?
```
luis-almeida: I don't think we need this comment here.
Instead, above the `closest` function declaration we could simply put a `Element.closest not yet supported in IE11`.

script.js
```diff
@@ -1,129 +1,159 @@
...
+  [].forEach.call(document.querySelectorAll('#request-status-select, #request-organization-select'), function(el) {
+    el.addEventListener('change', function() {
       search();
     });
+  });
 
   // Submit requests filter form in the request list page
-  $("#quick-search").on("keypress", function(e) {
-    if (e.which === 13) {
+  const quickSearch = document.querySelector('#quick-search')
```
luis-almeida: Missing semicolon for consistency.

script.js
```diff
@@ -1,129 +1,159 @@
...
+  function closest (element, selector) {
+    if (Element.prototype.closest) {
+      return element.closest(selector);
+    }
+    do {
+      if (matches(element, selector)) return element;
+      element = element.parentElement || element.parentNode;
+    } while (element !== null && element.nodeType === 1);
+    return null;
+  }
```
luis-almeida: Both `matches` and `closest` could be moved to the top not to rely on hoisting.

script.js
```diff
@@ -1,129 +1,159 @@
...
-  $(".header .icon-menu").on("keyup", function(e) {
+  document.querySelector('.header .icon-menu').addEventListener('keyup', function(e) {
     if (e.keyCode === 13) { // Enter key
       e.stopPropagation();
       toggleNavigation(this);
     }
   });
 
-  $("#user-nav").on("keyup", function(e) {
+  document.querySelector('#user-nav').addEventListener('keyup', function(e) {
```
luis-almeida: Same selector used multiple times.

script.js
```diff
@@ -1,129 +1,159 @@
...
     if (e.keyCode === 27) { // Escape key
       e.stopPropagation();
-      this.setAttribute("aria-expanded", false);
-      $(".header .icon-menu").attr("aria-expanded", false);
+      this.setAttribute('aria-expanded', false);
+      document.querySelector('.header .icon-menu').setAttribute('aria-expanded', false);
```
luis-almeida: Same selector used multiple times.

script.js
```diff
@@ -1,129 +1,159 @@
...
     }
   });
 
-  if ($("#user-nav").children().length === 0) {
-    $(".header .icon-menu").hide();
+  if (document.querySelector('#user-nav').children.length === 0) {
+    document.querySelector('.header .icon-menu').style.display = 'none';
```
luis-almeida: Same here. We can 'cache' this elements in variables instead just like above.

script.js
```diff
@@ -1,129 +1,173 @@
-/*
- * jQuery v1.9.1 included
- */
+document.addEventListener('DOMContentLoaded', function() {
 
-$(document).ready(function() {
+  function matches (element, selector) {
```
augustocravosilva: since this is only used in case `closest` is not defined, would it make sense to have this logic inlined in the `closest` function?

script.js
```diff
@@ -1,129 +1,173 @@
...
+      element = element.parentElement || element.parentNode;
+    } while (element !== null && element.nodeType === 1);
+    return null;
+  }
 
   // social share popups
-  $(".share a").click(function(e) {
-    e.preventDefault();
-    window.open(this.href, "", "height = 500, width = 500");
+  [].forEach.call(document.querySelectorAll('.share a'), function(anchor) {
```
augustocravosilva: could we use `Array.prototype` instead of `[]`? Looks a bit less hacky to me
Also in line 51, 96, 166

script.js
```diff
@@ -1,129 +1,173 @@
...
-
-  $requestMarkAsSolvedButton.on("click", function () {
-    $requestMarkAsSolvedCheckbox.attr("checked", true);
-    $requestCommentSubmitButton.prop("disabled", true);
-    $(this).attr("data-disabled", true).closest("form").submit();
+  var requestMarkAsSolvedButton = document.querySelector('.request-container .mark-as-solved:not([data-disabled])'),
+    requestMarkAsSolvedCheckbox = document.querySelector('.request-container .comment-container input[type=checkbox]'),
+    requestCommentSubmitButton = document.querySelector('.request-container .comment-container input[type=submit]');
+
+  requestMarkAsSolvedButton && requestMarkAsSolvedButton.addEventListener('click', function () {
```
augustocravosilva: this should probably be consistent with the previous block and use `if(requestMarkAsSolvedButton)`

script.js
```diff
@@ -1,129 +1,173 @@
...
   // Change Mark as solved text according to whether comment is filled
-  var $requestCommentTextarea = $(".request-container .comment-container textarea");
-
-  $requestCommentTextarea.on("input", function() {
-    if ($requestCommentTextarea.val() !== "") {
-      $requestMarkAsSolvedButton.text($requestMarkAsSolvedButton.data("solve-and-submit-translation"));
-      $requestCommentSubmitButton.prop("disabled", false);
+  var requestCommentTextarea = document.querySelector('.request-container .comment-container textarea');
+
+  requestCommentTextarea && requestCommentTextarea.addEventListener('input', function() {
```
augustocravosilva: same here?

script.js
```diff
@@ -1,129 +1,173 @@
...
+    if (requestCommentTextarea.value !== '') {
```
augustocravosilva: this if statement blocks could be inverted for simplicity `if (requestCommentTextarea.value == '') {`

script.js
```diff
@@ -1,129 +1,173 @@
...
+    el.addEventListener('change', function() {
       search();
     });
+  });
 
   // Submit requests filter form in the request list page
-  $("#quick-search").on("keypress", function(e) {
-    if (e.which === 13) {
+  const quickSearch = document.querySelector('#quick-search')
+  quickSearch && quickSearch.addEventListener('keypress', function(e) {
```
augustocravosilva: we should also use a `if` here for consistency

script.js
augustocravosilva: still missing?

script.js
```diff
@@ -1,129 +1,173 @@
...
-  var $commentContainerTextarea = $(".comment-container textarea"),
-    $commentContainerFormControls = $(".comment-form-controls, .comment-ccs");
+  var commentContainerTextarea = document.querySelector('.comment-container textarea'),
+    commentContainerFormControls = document.querySelector('.comment-form-controls, .comment-ccs');
 
-  $commentContainerTextarea.one("focus", function() {
-    $commentContainerFormControls.show();
-  });
+  if (commentContainerTextarea) {
+    commentContainerTextarea.addEventListener('focus', function focusCommentContainerTextarea() {
```
luis-almeida: No need to name this function.

script.js
luis-almeida: Nit: Always `===`.

script.js
luis-almeida: Also, we can use `const` or any `es6` features here we're not transpiling this this file.

lovelovedokidoki: `const` is supported in the supported browsers but simple treated as `var`. Do you suggest I replace all occurrences with `var`?

script.js
lovelovedokidoki: That's odd. Either I missed a lot of these or I've lost them during rebase acrobatics to make the commit stack look nice. I'll use fixup's less from now on. Hopefully they're alle fixed now

script.js
```diff
@@ -1,129 +1,176 @@
...
+      }
+    });
+  }
 
   function search() {
+    var quickSearch = document.querySelector('#quick-search');
+    var requestStatusSelect = document.querySelector('#request-status-select');
+    var requestOrganizationSelect = document.querySelector('#request-organization-select');
+
     window.location.search = $.param({
```
augustocravosilva: jquery here 😭

lovelovedokidoki: Yikes.
https://github.com/zendesk/copenhagen_theme/pull/66

PR #66, opened 2019-01-21
Refactor jQuery to vanilla JS

lovelovedokidoki:
We want to not be dependent on jQuery anymore, therefore we must refactor to vanilla JS.

I missed a spot on my last PR, so here's another attempt at removing jQuery completely from the code.

* JIRA: [GG-249: Refactor script.js file in CPH theme to use pure vanilla JS](https://zendesk.atlassian.net/browse/GG-249)

cc @zendesk/guide-growth 

#### Risks:
Change touches logic that handles search queries and filtering in My activities / My requests.
It's tested manually in IE11 and newest version of Chrome already.

### DIFF
--- a/script.js
+++ b/script.js
@@ -95,8 +95,9 @@ document.addEventListener('DOMContentLoaded', function() {
 
   // Submit requests filter form in the request list page
   Array.prototype.forEach.call(document.querySelectorAll('#request-status-select, #request-organization-select'), function(el) {
-    el.addEventListener('change', function() {
-      search();
+    el.addEventListener('change', function(e) {
+      e.stopPropagation();
+      closest(this, 'form').submit();
     });
   });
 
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -115,11 +115,14 @@ document.addEventListener('DOMContentLoaded', function() {
...
-      organization_id: requestOrganizationSelect && requestOrganizationSelect.value
-    });
+    window.location.search = String.prototype.concat(
+      'query=',
+      quickSearch && quickSearch.value,
+      'status=',
+      requestStatusSelect && requestStatusSelect.value,
+      'organization_id=',
+      requestOrganizationSelect && requestOrganizationSelect.value
+    );
```
luis-almeida: Does this solution work when multiple params are selected? Do they not need to joined with `&`?

A couple more questions:
- Do we need to encode any of these values?
- Could `requestOrganizationSelect.value` or `requestStatusSelect.value` be `null` or `undefined`?

lovelovedokidoki:
Decided to remove code instead, since search filtering in requests lists is now handled by regular form submit. Now tested in newest versions of Chrome, Firefox and Safari, and IE11, working fine and hopefully no more jQuery this time.

PR #67, opened 2019-01-21
Fix styling breaking on recently viewed and related articles

vecerek:
cc: @zendesk/guide-growth 

#### At 768px:
screen shot 2019-01-21 at 22 31 18

#### At 767px:
screen shot 2019-01-21 at 22 31 50

#### When there are no recently viewed articles at 767px:
screen shot 2019-01-21 at 22 32 21

#### When there are no recently viewed articles at 768px:
screen shot 2019-01-21 at 23 19 55

#### Risks
Low

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.4.13",
+  "version": "1.4.14",
   "api_version": 1,
   "default_locale": "en-us",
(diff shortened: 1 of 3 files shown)


vecerek:
@lovelovedokidoki Nice catch! Totally missed it 🙂 Addressed in 80461c40d06e31ff9f835a4362e7b99e24df9c75 by using padding instead of margin. @luis-almeida what do you think of this solution?

PR #69, opened 2019-02-05
[GS-454] Highlight using colour rather than italics for CJK searches

ellimist:
@zendesk/guide-search 
@zendesk/guide-growth 

Not sure if this is the right place to add this, but CJK languages should not display search results matches in italics, they should instead use a normal style and a coloured background.

Example from google:
image

Before this change:
image

After this change:
image

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.4.16",
+  "version": "1.4.17",
   "api_version": 1,
   "default_locale": "en-us",
   "settings": [{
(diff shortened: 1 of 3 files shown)


ronan-mch:
@vecerek - how does this work with existing customers? Will the change be propagated to them automatically provided that they haven't customised their templates? What about customers who have modified their templates, do we have any way of encouraging them to adopt this change?

libo:
That should also apply for other languages without italic as Hebrew, Arabic, ...

ellimist:
@vecerek bumped the version + compiled the sass

@libo I'll add a task in our backlog for those too. We don't currently fully support those languages in Guide Search right now + I think those need a bit different handling, Google, does bold instead of yellow background for both arabic and hebrew.

image

PR #70, opened 2019-02-19
Adding support for Flexible Hierarchies in section page

aliciacatalina:
~For the upcoming feature of flexible hierarchies we are exposing the sections that are direct children of a section, as well as its articles.
I'm taking the section tree html from the category page and added it to the section page.~

Update:
After talking to product and design, the template design has changed, adding the subsections at the top, therefore keeping the pagination at the bottom.
It also adds an interaction with the subsections list that displays only the first 6 subsections and displays a "See all sections..." trigger to show them all.
There are many ways to do this, and I'm a little bit rusty on semantic HTML and a11y, so please let me know if you have any concerns.

The new designs also removed the border from the article list.

 Some decisions that were made:
- Using an inline SVG for the chevron icon. Reference: https://css-tricks.com/can-make-icon-system-accessible/
- Hiding the sections after the 6th one with CSS

Impacted templates:
- Category
Screenshot 2019-03-22 at 10 19 14

- Section with more than 6 subsections
Screenshot 2019-03-22 at 10 19 20
subsections

- Section with less than 6 subsections
Screenshot 2019-03-22 at 10 21 46

- Section without subsections
Screenshot 2019-03-22 at 10 40 58

### DIFF
--- a/script.js
+++ b/script.js
@@ -162,7 +162,7 @@ document.addEventListener('DOMContentLoaded', function() {
 
     seeAllTrigger.addEventListener("click", function(e) {
       subsectionsList.classList.remove("section-list--collapsed");
-      seeAllTrigger.remove();
+      seeAllTrigger.parentNode.removeChild(seeAllTrigger);
     });
   }
 });
--- a/script.js
+++ b/script.js
@@ -152,4 +152,17 @@ document.addEventListener('DOMContentLoaded', function() {
       this.setAttribute('aria-expanded', !isExpanded);
     });
   });
+
+  // If a section has more than 6 subsections, we collapse the list, and show a trigger to display them all
+  const seeAllTrigger = document.querySelector("#see-all-sections-trigger");
+  const subsectionsList = document.querySelector(".section-list");
+
+  if (subsectionsList && subsectionsList.children.length > 6) {
+    seeAllTrigger.setAttribute("aria-hidden", false);
+
+    seeAllTrigger.addEventListener("click", function(e) {
+      subsectionsList.classList.remove("section-list--collapsed");
+      seeAllTrigger.remove();
+    });
+  }
 });
--- a/styles/_section.scss
+++ b/styles/_section.scss
@@ -66,6 +66,7 @@
 }
 
 .see-all-sections-trigger {
+  cursor: pointer;
   display: block;
   padding: 15px;
   text-align: center;
--- a/styles/_section.scss
+++ b/styles/_section.scss
@@ -13,6 +13,64 @@
   }
 
   &-subscribe {
-    .dropdown-toggle::after { display: none; }
+    .dropdown-toggle::after {
+      display: none;
+    }
   }
+}
+
+.section-list {
+  margin: 40px 0;
+}
+
+.section-list-item:first-child {
+  border-top: 1px solid $border-color;
+}
+
+.section-list-item {
+  border-bottom: 1px solid $border-color;
+  font-size: $font-size-bigger;
+  padding: 15px 0;
+}
+
+.section-list-item a {
+  align-items: center;
+  color: $text_color;
+  display: flex;
+  justify-content: space-between;
+}
+
+.section-list {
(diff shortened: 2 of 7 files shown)


templates/section_page.hbs
```diff
@@ -41,6 +41,43 @@
...
+                    {{/if}}
+                  </li>
+                {{/each}}
+              </ul>
+              {{#if more_articles}}
+                <a href="{{url}}" class="see-all-articles">
+                  {{t 'show_all_articles' count=article_count}}
+                </a>
+              {{/if}}
+            {{else}}
```
vecerek: May sections with no articles contain nested sections? If yes, should this `else` branch be removed like the one in the category page?

aliciacatalina: Yep, you are right! Removed that `else` statement as well :)

augustocravosilva:
Let’s wait on merging this until we get closer to the release date :)

styles/_section.scss
```diff
@@ -13,6 +13,65 @@
...
+  border-top: 1px solid $border-color;
+}
+
+.section-list-item {
+  border-bottom: 1px solid $border-color;
+  font-size: $font-size-bigger;
+  padding: 15px 0;
+}
+
+.section-list-item a {
```
augustocravosilva: could some of these be nested to keep the style we use in other `scss` files?

templates/section_page.hbs
```diff
@@ -19,6 +19,22 @@
         {{/if}}
       </header>
 
+      {{#if section.sections}}
+        <ul class="section-list section-list--collapsed">
+          {{#each section.sections}}
+          <li class="section-list-item">
+              <a href="{{url}}">
+                <span>{{name}}</span>
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16" aria-hidden="true">
```
augustocravosilva: can this be added as a separate file similar to https://github.com/zendesk/copenhagen_theme/blob/master/assets/dropdown-arrow.svg ?

aliciacatalina: I'd lean in favour of inline `svg`s, this way, it will take in the color of the parent, in this case `text-color`. I'd like @luis-almeida's opinion on this.

styles/_section.scss
aliciacatalina: Done in https://github.com/zendesk/copenhagen_theme/pull/70/commits/d7a8440c098057ba6d682e8b3a8642a715d58b6d

script.js
```diff
@@ -152,4 +152,17 @@ document.addEventListener('DOMContentLoaded', function() {
...
+  // If a section has more than 6 subsections, we collapse the list, and show a trigger to display them all
+  const seeAllTrigger = document.querySelector("#see-all-sections-trigger");
+  const subsectionsList = document.querySelector(".section-list");
+
+  if (subsectionsList && subsectionsList.children.length > 6) {
+    seeAllTrigger.setAttribute("aria-hidden", false);
+
+    seeAllTrigger.addEventListener("click", function(e) {
+      subsectionsList.classList.remove("section-list--collapsed");
+      seeAllTrigger.remove();
```
luis-almeida: `remove` is not supported in `ie11` which we still support :(
Could we do `seeAllTrigger.parentNode.removeChild(seeAllTrigger);` instead just to be on the safe side?

templates/section_page.hbs
luis-almeida: I tend to agree that cutting the extra network request, and most importantly being able to use `currentColor`, justifies inlining the `svg`.
Would be nice to do the switch for `dropdown-arrow.svg` as well for consistency and at some point have a build tool that can do this for us.

script.js
aliciacatalina: Definitely! Changed in https://github.com/zendesk/copenhagen_theme/pull/70/commits/5b44b7d60be3e9f527687b04a9bec822e6589d02

PR #71, opened 2019-03-08
[PUB-1270] Do not render anchor tag on empty breadcrumb pieces

lovelovedokidoki:
The `path_steps` helper already returns '#' for the URL and '...' as name when a breadcrumb piece is a set of truncated items. We don't need these elipsis to have an anchor tag that points to nowhere.

* See JIRA: [PUB-1270: Remove link from ellipsis in search results in Copenhagen theme](https://zendesk.atlassian.net/browse/PUB-1270)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.4.17",
+  "version": "1.4.18",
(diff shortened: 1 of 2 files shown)


augustocravosilva:
Please mention @zendesk/guide-growth when doing changes to the Copenhagen Theme

PR #75, opened 2019-03-21
Adds show all string

aliciacatalina:
I need to add a string for some functionality we are including for adding support in flexible hierarchy.
There is nothing in the README for this, does this count as another version? Is this the file I need to update? Let me know @zendesk/guide-growth


augustocravosilva:
No, please add it to https://github.com/zendesk/copenhagen_theme/blob/master/translations.yml

This file is the generated after translations are done
No need to bump the version

PR #78, opened 2019-04-03
Remove unused styling for Google+ share button

VilhelmNielsen:
Google+ share button is no longer in HC as of today, this PR removes redundant styling for that button.

/cc: @zendesk/guide-growth


VilhelmNielsen:
> Could you also bump the version in the manifest file? 🙂

Sure thing. Is it semver?

PR #79, opened 2019-04-17
[GS-458] Change snippet highlight for non-latin languages

ellimist:
#### Description

As a continuation from #69 I've changed the way we highlight search results for all the supported non-latin languages. I've used Google as inspiration for this (huehue). 

They show a different background colour for Chinese, and show all the rest of the languages in bold.

@zendesk/guide-growth @zendesk/guide-search

### DIFF
--- a/styles/_search_results.scss
+++ b/styles/_search_results.scss
@@ -88,7 +88,7 @@
   }
 }
 
-/* Non-latin search results highlight*/
+/* Non-latin search results highlight */
 
 /* Add a yellow background for Chinese */
 html[lang|="zh"] {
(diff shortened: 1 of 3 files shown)


styles/_search_results.scss
```diff
@@ -88,13 +88,27 @@
   }
 }
 
-/* Don't display CJK results in italic */
-/* Add a yellow background instead */
-html[lang|="zh"],
-html[lang|="ko"],
-html[lang|="ja"] {
+/* Non-latin search results highlight*/
```
augustocravosilva: nit - space missing
```suggestion
/* Non-latin search results highlight */
```

PR #83, opened 2019-06-05
We need to return focus to elements after form submit, see HCJ-1488 for details

yara-simargil:
In some cases we want to return focus to previous active element after form submit with full page reload.

Reference: https://zendesk.atlassian.net/browse/HCJ-1488

### DIFF
--- a/script.js
+++ b/script.js
@@ -22,6 +22,18 @@ document.addEventListener('DOMContentLoaded', function() {
     });
   });
 
+  // In some cases we should preserve focus after page reload
+  function saveFocus() {
+    var activeElementId = document.activeElement.getAttribute("id");
+    sessionStorage.setItem('returnFocusTo', '#' + activeElementId);
+  }
+  var returnFocusTo = sessionStorage.getItem('returnFocusTo');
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -165,4 +175,17 @@ document.addEventListener('DOMContentLoaded', function() {
       seeAllTrigger.parentNode.removeChild(seeAllTrigger);
     });
   }
+
+  // In some cases we should preserve focus after page reload, see #HCJ-1488
```
luis-almeida: Lets remove this mention to an internal ticket as this is an open source repo.

script.js
```diff
@@ -165,4 +175,17 @@ document.addEventListener('DOMContentLoaded', function() {
...
+  function saveFocus() {
+    var activeElementId = document.activeElement.getAttribute("id");
+    sessionStorage.setItem('returnFocusTo', '#' + activeElementId);
+  }
```
luis-almeida: Nit: define  the function before usage to avoid hoisting and for consistency.

yara-simargil: Thanks for the comments.
Moved it to the top.

PR #84, opened 2019-06-05
Fix color contrast for status labels, unify color with support and garden

yara-simargil:
During a11y audit we figured out that some status labels had poor color contrast.
I tried unifying color with what we use in Support (`pending` status uses blue, `new` status uses yellow + red) and what closest accessible versions are available in Graden.
Also I've set `font-weight` to `600` as we do in Support and in `tag` component in Garden.

Reference: https://zendesk.atlassian.net/browse/HCJ-1946

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.5.2",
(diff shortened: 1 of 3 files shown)


VilhelmNielsen:
Could you bump the version in manifest, please? 😃

PR #87, opened 2019-06-11
Make collapsible menus keyboard accessible

anpa:
This PR adds keyboard navigation to `collapsible-nav` and `collapsible-sidebar`.

#### Before
a11y_before

#### After
a11y_after

### DIFF
--- a/templates/article_page.hbs
+++ b/templates/article_page.hbs
@@ -9,7 +9,7 @@
     <section class="article-sidebar">
       {{#if settings.show_articles_in_section}}
         <section class="section-articles collapsible-sidebar">
-          <button class="collapsible-sidebar-toggle" aria-labelledby="section-articles-title" aria-expanded="false"></button>
+          <button type="button" class="collapsible-sidebar-toggle" aria-labelledby="section-articles-title" aria-expanded="false"></button>
           <h3 id="section-articles-title" class="collapsible-sidebar-title sidenav-title">{{t 'articles_in_section'}}</h3>
           <div class="collapsible-sidebar-body">
             <ul>
--- a/templates/article_page.hbs
+++ b/templates/article_page.hbs
@@ -9,17 +9,20 @@
     <section class="article-sidebar">
       {{#if settings.show_articles_in_section}}
         <section class="section-articles collapsible-sidebar">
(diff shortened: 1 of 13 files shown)


templates/article_page.hbs
```diff
@@ -9,17 +9,20 @@
...
-            <li>
-              <a href="{{url}}" class="sidenav-item {{#is id ../article.id}}current-article{{/is}}">{{title}}</a>
-            </li>
-          {{/each}}
-        </ul>
-        {{#if section.more_articles}}
-          <a href="{{section.url}}" class="article-sidebar-item">{{t 'see_more'}}</a>
-        {{/if}}
+          <button class="collapsible-sidebar-toggle" aria-labelledby="section-articles-title" aria-expanded="false"></button>
+          <h3 id="section-articles-title" class="collapsible-sidebar-title sidenav-title">{{t 'articles_in_section'}}</h3>
```
yara-simargil: I wonder why we have an empty button with following `h3` explaining it?
Also I'd prefer to always set `type` for buttons.

VilhelmNielsen:
It doesn't seem very accessible to me, that the current page in the menu isn't reachable by keyboard. I think it should be an a-tag with `aria-current="page"` ([WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/#aria-current))

templates/article_page.hbs
anpa: @yara-simargil The reason to have an empty button is to show the chevron only on mobile. It's explained by the `h3` because the purpose of the `collapsible-sidebar` depends on the context (as opposed to `collapsible-nav` that is only meant for navigation).

_i.e.: (toggle) Ticket details, (toggle) Articles in this section, ..._

Good idea with the button `type`. Will update it 👍

anpa:
@VilhelmNielsen that's correct. Unfortunately it seems to be an existing pattern in the theme to not show the current page (or item) as a link. I scope that change out of this PR as that require some further changes. We will pick it out afterwards if that's ok.

anpa:
@zendesk/guide-growth I can't seem to merge without your approval. Do you mind another look pls?

PR #88, opened 2019-06-12
Fix color contrast for default status label

yara-simargil:
Following up on https://github.com/zendesk/copenhagen_theme/pull/84

Default background color for status icon also has insufficient color contrast. I'm suggesting to use `#058541` instead.

Screenshot 2019-06-12 at 15 36 27

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.5.5",
+  "version": "1.5.6",
   "api_version": 1,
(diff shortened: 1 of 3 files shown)


DovileM:
If it's possible let's use https://garden.zendesk.com/css-components/utilities/color/
colors with 600 values: RED-600, GREEN-600, YELLOW-600 
As Garden did a lot of work to test all the colors for accessibility. 

Let me know if that helps.

yara-simargil:
Updated to use GREEN-600. After talking to Dovile we've decided all the rest are fine.

PR #89, opened 2019-06-18
Improve page headings

anpa:
This PR refactors some pages to ensure:

- each page has an H1 level heading;
- each page does not skip heading levels;

This is part of an effort to improve accessibility in the theme.

### DIFF
--- a/style.css
+++ b/style.css
@@ -4078,6 +4078,11 @@ ul {
   }
 }
 
+.search-results-subheading {
+  font-size: 18px;
+  font-weight: 600;
+}
+
 .search-results-list {
   margin-bottom: 25px;
 }
--- a/style.css
+++ b/style.css
@@ -2000,8 +2000,9 @@ ul {
   margin-bottom: 30px;
 }
 
-.community-header h4 {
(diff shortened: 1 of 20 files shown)


style.css
```diff
@@ -2514,17 +2524,11 @@ ul {
   }
 }
 
-.post-sidebar h5 {
+.post-sidebar-title {
+  font-size: 18px;
   font-weight: 600;
 }
 
-@media (min-width: 1024px) {
```
yara-simargil: Why don't we need these border anymore?

styles/_post.scss
```diff
@@ -139,12 +139,8 @@
     padding: 30px 0;
     text-align: center;
 
-    h5 {
-      @include desktop {
```
yara-simargil: And this one as well, why can we remove it?

style.css
anpa: It wasn't being used as there was no `h5` inside `.post-sidebar`

yara-simargil: Ah, I see. 👍

PR #90, opened 2019-06-25
[COMM-145] Add classes to differentiate the meta data in user profiles page

vecerek:
@TetianaG, what would be the most appropriate way to distinguish between different meta data in the user profile page with regards to browser testing?

JIRA: [COMM-145]

[COMM-145]: https://zendesk.atlassian.net/browse/COMM-145


vecerek:
No longer relevant.

PR #95, opened 2019-08-21
Collapse multibrand filters when there are more than 5 items

Jiyu-Zhang-Zendesk:
Collapse multibrand search filters when there are more than 5 help centers or categories:

Screen Shot 2019-08-21 at 20 08 45

And expand it when user clicks the button:

Screen Shot 2019-08-21 at 20 10 19

This is a refined version of @ellimist's original branch.

Without toggle button our SERP looks like this in mobile web:
Screen Shot 2019-08-22 at 14 39 38
Note that there is no way to expand filters.

With toggle button it looks like this in mobile web:
Screen Shot 2019-08-26 at 19 51 24
Once user clicks it, he/she will see all help centers or categories or topics.

### DIFF
--- a/script.js
+++ b/script.js
@@ -200,4 +200,21 @@ document.addEventListener('DOMContentLoaded', function() {
       seeAllTrigger.parentNode.removeChild(seeAllTrigger);
     });
   }
+
+  // If multibrand search has more than 5 help centers or categories collapse the list
+  const multibrandFilterLists = document.querySelectorAll(".multibrand-filter-list");
+  Array.prototype.forEach.call(multibrandFilterLists, function(filter) {
+    if (filter.children.length > 5) {
+      // Display the show more button
+      var trigger = filter.querySelector(".see-all-filters-trigger");
+      trigger.setAttribute("aria-hidden", false);
+
+      // Add event handler for click
+      trigger.addEventListener("click", function(e) {
+        e.stopPropagation();
+        trigger.parentNode.removeChild(trigger);
+        filter.classList.remove("multibrand-filter-list--collapsed")
+      })
+    }
+  });
 });
--- a/script.js
+++ b/script.js
@@ -206,7 +206,7 @@ document.addEventListener('DOMContentLoaded', function() {
   Array.prototype.forEach.call(multibrandFilterLists, function(filter) {
     if (filter.children.length > 5) {
       // Display the show more button
-      var trigger = filter.querySelector(".see-all-filters-trigger");
+      var trigger = filter.querySelector(".see-all-filters");
       trigger.setAttribute("aria-hidden", false);
 
       // Add event handler for click
--- a/style.css
+++ b/style.css
@@ -4096,28 +4096,33 @@ ul {
   }
 }
 
-.search-results-sidebar .sidenav-item[aria-selected="true"] {
-  background-color: $brand_color;
-  color: $brand_text_color;
+.search-results-sidebar .sidenav-item:hover, .search-results-sidebar .sidenav-item[aria-selected="true"] {
+  background-color: #e9ebed;
+  color: inherit;
   text-decoration: none;
 }
 
 .search-results-sidebar .collapsible-sidebar {
-  padding-bottom: 15px;
+  margin-bottom: 30px;
 }
 
 .search-results-sidebar .multibrand-filter-list--collapsed li:nth-child(1n + 6) {
   display: none;
 }
 
-.search-results-sidebar .see-all-filters-trigger {
+.search-results-sidebar .see-all-filters {
   cursor: pointer;
   display: block;
   padding: 10px;
 }
 
-.search-results-sidebar .see-all-filters-trigger[aria-hidden="true"] {
+.search-results-sidebar .see-all-filters[aria-hidden="true"] {
   display: none;
+}
+
+.search-results-sidebar .see-all-filters:after {
+  content: ' \2304';
+  font-weight: bold;
 }
 
 .search-results-subheading {
@@ -4206,22 +4211,14 @@ ul {
   margin: 0 5px;
 }
 
-/* Non-latin search results highlight */
-/* Add a yellow background for Chinese */
-html[lang|="zh"] .search-result-description em {
+/* By default use bold instead of italic to highlight */
+.search-results-description em {
   font-style: normal;
-  background: yellow;
+  font-weight: bold;
 }
 
-/* Use bold to highlight for the rest of supported non-latin languages */
-html[lang|="ar"] .search-result-description em,
-html[lang|="bg"] .search-result-description em,
-html[lang|="el"] .search-result-description em,
-html[lang|="he"] .search-result-description em,
-html[lang|="hi"] .search-result-description em,
-html[lang|="ko"] .search-result-description em,
-html[lang|="ja"] .search-result-description em,
-html[lang|="ru"] .search-result-description em,
-html[lang|="th"] .search-result-description em {
-  font-style: bold;
+/* Add a yellow background for Chinese */
+html[lang|="zh"] .search-results-description em {
+  font-style: normal;
+  background: yellow;
 }
--- a/style.css
+++ b/style.css
@@ -4102,6 +4102,24 @@ ul {
   text-decoration: none;
 }
 
+.search-results-sidebar .collapsible-sidebar {
+  padding-bottom: 15px;
+}
+
+.search-results-sidebar .multibrand-filter-list--collapsed li:nth-child(1n + 6) {
+  display: none;
+}
+
+.search-results-sidebar .see-all-filters-trigger {
+  cursor: pointer;
+  display: block;
+  padding: 10px;
+}
+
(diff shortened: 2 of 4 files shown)


templates/search_results.hbs
```diff
@@ -10,12 +10,13 @@
...
         <section class="filters-in-section collapsible-sidebar">
           <h3 class="collapsible-sidebar-title sidenav-title">{{t 'filter_by_help_center'}}</h3>
-          <ul>
+          <ul class="multibrand-filter-list multibrand-filter-list--collapsed">
             {{#each help_center_filters}}
               <li>
                 <a href="{{url}}" class="sidenav-item" aria-selected="{{selected}}">{{name}} ({{count}})</a>
               </li>
             {{/each}}
+            <a tabindex="0" class="see-all-filters-trigger" aria-hidden="true" title="{{t 'see_more'}}">{{t 'see_more'}}</a>
```
VilhelmNielsen: Please use `button` instead for `a`.

And `tabindex="0"` shouldn't be necessary.

templates/search_results.hbs
```diff
@@ -39,12 +40,13 @@
...
             <h3 class="collapsible-sidebar-title sidenav-title">{{t 'filter_by_topic'}}</h3>
           {{/is}}
-          <ul>
+          <ul class="multibrand-filter-list multibrand-filter-list--collapsed">
             {{#each subfilters}}
               <li>
                 <a href="{{url}}" class="sidenav-item" aria-selected="{{selected}}">{{name}} ({{count}})</a>
               </li>
             {{/each}}
+            <a tabindex="0" class="see-all-filters-trigger" aria-hidden="true" title="{{t 'see_more'}}">{{t 'see_more'}}</a>
```
VilhelmNielsen: Please use `button` instead for `a`.

And `tabindex="0"` shouldn't be necessary.

templates/search_results.hbs
Jiyu-Zhang-Zendesk: @VilhelmNielsen thanks for reviewing this. according to the design it's a link instead of a button:
https://zendesk.invisionapp.com/share/QZS1UPISHY5#/screens/357598308
that's why we use `a`

Jiyu-Zhang-Zendesk: `tabindex="0"` was inspired by this but not be necessary:
https://github.com/zendesk/copenhagen_theme/blob/master/templates/section_page.hbs#L32
I'll double check

templates/search_results.hbs
Jiyu-Zhang-Zendesk: same here

script.js
```diff
@@ -167,16 +167,18 @@ document.addEventListener('DOMContentLoaded', function() {
...
+      el.addEventListener('click', function (e) {
+        toggleNavigation(toggle, this);
+      });
+
+      el.addEventListener('keyup', function (e) {
+        if (e.keyCode === 27) { // Escape key
+          closeNavigation(toggle, this);
+        }
+      });
+    }
```
cmacaulay: 👍

templates/search_results.hbs
cmacaulay: Should you change the language to match what's in the design mockup? So, instead of `see more`, it would say `Show more Help Centers`, or for below, `Show more Categories` and `Show more Sections`?

VilhelmNielsen: @Jiyu-Zhang-Zendesk Since it doesn't cause navigation, it should be `button` semantically (for a11y). Then apply styling so it looks like a link.

Jiyu-Zhang-Zendesk: @VilhelmNielsen `tabindex="0"` is still needed here because our `a` doesn't have `href`. so without `tabindex` it won't be focusable

Jiyu-Zhang-Zendesk: @cmacaulay that's a good idea and I also mentioned it in the standup. not critical but might be nice to have.

Jiyu-Zhang-Zendesk: > @Jiyu-Zhang-Zendesk Since it doesn't cause navigation, it should be `button` semantically (for a11y). Then apply styling so it looks like a link.

I asked Augusto about this. This is a good point but we also have consideration over consistency since we already use hyperlinks for such cases in the section page:
https://github.com/zendesk/copenhagen_theme/pull/70
let's talk about this tomorrow. thanks

style.css
```diff
@@ -4188,22 +4211,14 @@ ul {
...
-html[lang|="ar"] .search-result-description em,
-html[lang|="bg"] .search-result-description em,
-html[lang|="el"] .search-result-description em,
-html[lang|="he"] .search-result-description em,
-html[lang|="hi"] .search-result-description em,
-html[lang|="ko"] .search-result-description em,
-html[lang|="ja"] .search-result-description em,
-html[lang|="ru"] .search-result-description em,
-html[lang|="th"] .search-result-description em {
-  font-style: bold;
```
luis-almeida: Are these no longer needed?

styles/_search_results.scss
```diff
@@ -124,27 +147,16 @@
...
-/* Non-latin search results highlight */
+/* By default use bold instead of italic to highlight */
+.search-results-description em {
+  font-style: normal;
+  font-weight: bold;
+}
 
 /* Add a yellow background for Chinese */
 html[lang|="zh"] {
-  .search-result-description em {
```
luis-almeida: I can't find references to `search-result-description` so I assume it was wrong and we're just correcting it to `search-results-description` (?)

templates/search_results.hbs
```diff
@@ -9,18 +9,21 @@
...
+          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false"></button>
           <h3 class="collapsible-sidebar-title sidenav-title">{{t 'filter_by_help_center'}}</h3>
-          <ul>
+          <ul class="multibrand-filter-list multibrand-filter-list--collapsed">
             {{#each help_center_filters}}
               <li>
                 <a href="{{url}}" class="sidenav-item" aria-selected="{{selected}}">{{name}} ({{count}})</a>
               </li>
             {{/each}}
+            <a tabindex="0" class="see-all-filters-trigger" aria-hidden="true" title="{{t 'show_more_help_centers'}}">{{t 'show_more_help_centers'}}</a>
```
luis-almeida: Nit: No need for the `-trigger` suffix, `see-all-filters` should be enough.

templates/search_results.hbs
```diff
@@ -9,18 +9,21 @@
     <section class="search-results-sidebar">
       {{#if help_center_filters}}
         <section class="filters-in-section collapsible-sidebar">
+          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false"></button>
```
luis-almeida: Is this empty button needed?

templates/search_results.hbs
```diff
@@ -9,18 +9,21 @@
...
                 <a href="{{url}}" class="sidenav-item" aria-selected="{{selected}}">{{name}} ({{count}})</a>
               </li>
             {{/each}}
+            <a tabindex="0" class="see-all-filters-trigger" aria-hidden="true" title="{{t 'show_more_help_centers'}}">{{t 'show_more_help_centers'}}</a>
           </ul>
         </section>
       {{/if}}
       {{#if help_center.community_enabled}}
         <section class="filters-in-section collapsible-sidebar">
+          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false"></button>
```
luis-almeida: Same here.

templates/search_results.hbs
```diff
@@ -33,18 +36,25 @@
       {{/if}}
       {{#if subfilters}}
         <section class="filters-in-section collapsible-sidebar">
+          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false"></button>
```
luis-almeida: Same here.

style.css
Jiyu-Zhang-Zendesk: we now set `font-weight` to `bold` for all locales (except Chinese). so we don't need these any more

styles/_search_results.scss
Jiyu-Zhang-Zendesk: yes. this should have been a typo. now correct it.

templates/search_results.hbs
Jiyu-Zhang-Zendesk: yes. without the toggle button we can't expand filters on mobile web. I have uploaded two screenshots to compare in the description of this PR.

templates/search_results.hbs
Jiyu-Zhang-Zendesk: good point. I'll update it soon

templates/search_results.hbs
Jiyu-Zhang-Zendesk: this one is also needed to expand filters on mobile web

templates/search_results.hbs
Jiyu-Zhang-Zendesk: this one is also needed to expand subfilters on mobile web

templates/search_results.hbs
Jiyu-Zhang-Zendesk: renamed to `see-all-filters` now

PR #96, opened 2019-08-22
Moderator actions: flagged, hidden

zdrve:
DO NOT MERGE YET (because the help center master branch does not yet expose the provided data)

Support the upcoming "flagged" and "hidden" properties of community posts, community comments, and article comments. The presenters will expose "flagged" and "hidden"; these booleans are independent. "flagged" will be false if you do not have permission to see flagged-ness; likewise for "hidden".

Currently the translations for the labels are separate for posts vs comments; maybe it would be better if that weren't the case?

### DIFF
--- a/style.css
+++ b/style.css
@@ -2291,6 +2291,16 @@ ul {
   text-align: center;
 }
 
+.status-label-flagged {
+  background-color: #dea639;
+  text-align: center;
+}
+
+.status-label-hidden {
+  background-color: #c07160;
+  text-align: center;
+}
+
 .status-label-open {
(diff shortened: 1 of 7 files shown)


luis-almeida:
The code seems ok to me. This PR will only be mergeable after the api changes and accompanying documentation.

Please squash the commits and add a descriptive commit message.

augustocravosilva:
> Currently the translations for the labels are separate for posts vs comments; maybe it would be better if that weren't the case?

I think it would make sense to only have 1 translation for each property

PR #97, opened 2019-08-28
Align the styles of multibrand SERP with the design

Jiyu-Zhang-Zendesk:
Change multibrand theme to be aligned with the design. Split this to a separate PR from the old one with `show more` button.

Changes:
- Align author name and updated time to the right
- Align # of votes to the right and in the same line as the title
- Change the color of breadcrumbs in search results
- Use different color for doc counts in multibrand filters

### DIFF
--- a/style.css
+++ b/style.css
@@ -4110,6 +4110,18 @@ ul {
   display: none;
 }
 
+.search-results-sidebar .multibrand-filter-list .doc-count {
+  color: #999;
+}
+
+.search-results-sidebar .multibrand-filter-list .doc-count::before {
+  content: "(";
+}
+
+.search-results-sidebar .multibrand-filter-list .doc-count::after {
+  content: ")";
+}
+
 .search-results-sidebar .see-all-filters {
   cursor: pointer;
   display: block;
@@ -4120,7 +4132,7 @@ ul {
   display: none;
 }
 
-.search-results-sidebar .see-all-filters:after {
+.search-results-sidebar .see-all-filters::after {
   content: ' \2304';
   font-weight: bold;
 }
@@ -4146,14 +4158,37 @@ ul {
   margin-bottom: 0;
 }
 
+.search-results .meta-group {
+  display: flex;
+  align-items: center;
+  clear: right;
+  color: #999;
+}
+
+.search-results .meta-group li:first-child {
+  margin-right: auto;
+}
+
+.search-results .meta-group .meta-data {
+  color: inherit;
+}
+
+.search-results .meta-group .meta-data:not(:last-child) {
+  margin-right: 20px;
+}
+
+.search-results .meta-group .meta-data::after {
+  content: none;
+}
+
+.search-results-description {
+  margin-top: 10px;
+  word-break: break-word;
+}
+
 .search-result-title {
   font-size: 16px;
   display: inline-block;
-}
-
-.search-result-description {
-  margin-top: 15px;
-  word-break: break-word;
 }
 
 .search-result-icons {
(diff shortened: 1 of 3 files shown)


style.css
```diff
@@ -4146,16 +4158,39 @@ ul {
   margin-bottom: 0;
 }
 
-.search-result-title {
-  font-size: 16px;
-  display: inline-block;
+.search-results .meta-group {
+  display: flex;
+  align-items: center;
+  clear: right;
```
luis-almeida: Is this `clear` rule needed?
As far as I know it should only work with floating elements.

style.css
```diff
@@ -4202,13 +4237,16 @@ ul {
 }
 
 .search-result-breadcrumbs {
+  display: table-row;
```
luis-almeida: Could we use something other than `table-row`, e.g. `inline-block`, since we're not within a table?

style.css
```diff
@@ -4202,13 +4237,16 @@ ul {
...
   margin: 0;
 }
 
-.search-result-breadcrumbs li:last-child::after {
-  content: "·";
-  display: inline-block;
-  margin: 0 5px;
+.search-result-breadcrumbs li {
+  display: table-cell;
```
luis-almeida: Same here.

styles/_search_results.scss
```diff
@@ -80,6 +92,34 @@
       }
     }
   }
+
+  .meta-group {
+    display: flex;
+    align-items: center;
+    clear: right;
```
luis-almeida: Same here. Is this `clear` rule needed?

styles/_search_results.scss
```diff
@@ -140,12 +175,15 @@
...
   &-breadcrumbs {
+    display: table-row;
     margin: 0;
 
-    li:last-child::after {
-       content: "·";
-       display: inline-block;
-       margin: 0 5px;
+    li {
+      display: table-cell;
```
luis-almeida: Same here. Lets not use table layouts as much as possible.

styles/_search_results.scss
```diff
@@ -80,6 +92,34 @@
...
+
+    li:first-child {
+      margin-right: auto;
+    }
+
+    .meta-data {
+      color: inherit;
+
+      &:not(:last-child) {
+        margin-right: 20px;
```
luis-almeida: There's several `margin-right` applied in this PR. Usually needs the opposite for `rtl`.
Are we adding support for `rtl` in a following PR?

Jiyu-Zhang-Zendesk: good catch. I actually haven't thought about `rtl` yet. do we have an example to handle `rtl` in the css?

styles/_search_results.scss
Jiyu-Zhang-Zendesk: I used `float: right` to align # of votes to the right of article titles. If I don't clear that the breadcrumbs in search results will not be displayed correctly.

styles/_search_results.scss
luis-almeida: Here's an example of us swapping the `right` specific styles with `left` for `rtl` support:
https://github.com/zendesk/copenhagen_theme/blob/master/style.css#L1460-L1468

styles/_search_results.scss
luis-almeida: Got it! I missed that float rule 👍

style.css
Jiyu-Zhang-Zendesk: there might be a better way here. I used `table-row` and `table-cell` to make sure `.search-result-breadcrumbs li` has the same height as `.search-result-breadcrumbs` such that they are aligned vertically with `meta-data`

luis-almeida: I think we used those at the time to keep support for older browsers that we no longer support. But you have a point, let's keep it consistent 👍

PR #99, opened 2019-09-05
Move parentheses inside result count and fix the layout in rtl

Jiyu-Zhang-Zendesk:
- Move parentheses inside result count elements as suggested by @augustocravosilva 
- Fix the broken layout of parentheses in rtl
- Fix the issue that `see more` button is displayed when there are more than 4 filters (should be 5)

Layout in rtl before fix:
Screen Shot 2019-09-05 at 08 57 35

Layout in rtl after fix:
Screen Shot 2019-09-05 at 08 57 49

### DIFF
--- a/script.js
+++ b/script.js
@@ -204,7 +204,7 @@ document.addEventListener('DOMContentLoaded', function() {
   // If multibrand search has more than 5 help centers or categories collapse the list
   const multibrandFilterLists = document.querySelectorAll(".multibrand-filter-list");
   Array.prototype.forEach.call(multibrandFilterLists, function(filter) {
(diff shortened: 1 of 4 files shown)


Jiyu-Zhang-Zendesk:
> Is this to be merged into master? If so, please bump the version:
> 
> https://github.com/zendesk/copenhagen_theme/blob/bc646972db322ba09bd877602ef36465d51eaed1/manifest.json#L4

@VilhelmNielsen sorry for the late response. this is to be merged to `multibrand-search` not master so I guess it's fine to keep the version as it is in that branch

PR #107, opened 2019-11-18
Update manifest to v2

VilhelmNielsen:
I've created a branch for API v2 changes, and this commit bumps the api_version in the manifest.

Should the theme version be bumped to `2.0.0`?

cc @zendesk/guide-growth

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.8.2",
+  "version": "2.0.0-alpha.1",
   "api_version": 2,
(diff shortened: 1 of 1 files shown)


VilhelmNielsen:
Can we do `2.0.0-alpha.x` or similar? Or should we just disregard versioning on this branch entirely?

augustocravosilva:
`2.0.0.alpha.1` should work 👍

VilhelmNielsen:
> `2.0.0.alpha.1` should work 👍

That gave this error

```
file: manifest.json

message: The property '#/version' was not of a maximum string length of 10
```

PR #115, opened 2019-12-05
[COMM-90] & [COMM-628] Remove arrows from vote buttons

fredefox:
This changes should follow the following change in Help Center: https://github.com/zendesk/help_center/pull/19781

### [COMM-628]

Rather than adding

* Unicode Character 'UPWARDS BLACK ARROW' (U+2B06)
* Unicode Character 'DOWNWARDS BLACK ARROW' (U+2B07)

to the vote buttons using `::before` and a font to go with those
symbols we want to use SVG to display those arrows.  To style those
icons I'm also removing the `::before` selectors for the remaining
rules that deal with styling (hover highlight etc) for these elements.

The modification of the `font-size`, `display` and `flex-direction`
was to fix a centering/alignment issue with this new approach.

I've copied over / inlined this file:

https://github.com/zendeskgarden/svg-icons/blob/master/src/16/chevron-down-fill.svg

From zendeskgarden.  For the upwards pointing caret I've added an SVG
transformation to the root `<svg>` element to mirror it along the
horizontal axis.  Otherwise the two elements are the same.

### [COMM-90]
I've removed the `outline: none` rule. The visual clue is consistent with the other keyboard-navigatable elements and the chevron is aligned within the outline. This is what it looks like:

image

[COMM-628]: https://zendesk.atlassian.net/browse/COMM-628

[COMM-90]: https://zendesk.atlassian.net/browse/COMM-90

### DIFF
--- a/templates/community_post_page.hbs
+++ b/templates/community_post_page.hbs
@@ -80,9 +80,17 @@
           <div class="post-actions-wrapper">
             <div class="post-vote vote" role='radiogroup'>
               {{#with post}}
-                {{vote 'up' role='radio' class='vote-up' selected_class='vote-voted'}}
+                {{#vote 'up' role='radio' class='vote-up' selected_class='vote-voted'}}
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16" transform="scale(1,-1)">
+                  <path fill="none" stroke="currentColor" stroke-linecap="round" stroke-width="2" d="M4 6.5l3.6 3.6c.2.2.5.2.7 0L12 6.5"/>
+                </svg>
+                {{/vote}}
                 {{vote 'sum' class='vote-sum'}}
-                {{vote 'down' role='radio' class='vote-down' selected_class='vote-voted'}}
+                {{#vote 'down' role='radio' class='vote-down' selected_class='vote-voted'}}
+                <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16">
+                  <path fill="none" stroke="currentColor" stroke-linecap="round" stroke-width="2" d="M4 6.5l3.6 3.6c.2.2.5.2.7 0L12 6.5"/>
+                </svg>
+                {{/vote}}
               {{/with}}
             </div>
 
@@ -174,9 +182,17 @@
               <div class="comment-actions-container">
                 {{#unless official}}
                   <div class="comment-vote vote" role='radiogroup'>
-                    {{vote 'up' role='radio' class='vote-up' selected_class='vote-voted'}}
+                    {{#vote 'up' role='radio' class='vote-up' selected_class='vote-voted'}}
+                    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16" transform="scale(1,-1)">
+                      <path fill="none" stroke="currentColor" stroke-linecap="round" stroke-width="2" d="M4 6.5l3.6 3.6c.2.2.5.2.7 0L12 6.5"/>
+                    </svg>
+                    {{/vote}}
                     {{vote 'sum' class='vote-sum'}}
-                    {{vote 'down' role='radio' class='vote-down' selected_class='vote-voted'}}
+                    {{#vote 'down' role='radio' class='vote-down' selected_class='vote-voted'}}
+                    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16">
+                      <path fill="none" stroke="currentColor" stroke-linecap="round" stroke-width="2" d="M4 6.5l3.6 3.6c.2.2.5.2.7 0L12 6.5"/>
+                    </svg>
+                    {{/vote}}
                   </div>
                 {{/unless}}
                 <div class="comment-actions actions">
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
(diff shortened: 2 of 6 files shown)


VilhelmNielsen:
This breaks the current icons, yeah?

fredefox:
> This breaks the current icons, yeah? How are the changes reflected in Help Center?

Spot on! That PR is on its way. I've updated the description. Cheers.

fredefox:
Here's what it looks like with the current changes:

image

If you, like me, feel that the carets are not perfectly aligned here is a screenshot where I added a border to the container with the number.

image

fredefox:
I feel that the 0 in the first screenshot is a bit of-center w.r.t the vertical axis. Perhaps this has to do with some kind out rounding issues in the render? Couldn't really find an issue with the CSS.

augustocravosilva:
cc @DovileM

fredefox:
There was a similar file in the Help Center code base that I initially thought I had to modify, but that didn't do anything. I don't know if I need to change that as well. The file I'm thinking of is https://github.com/zendesk/help_center/blob/master/app/views/communities/posts/show.html.hbs

DovileM:
@fredefox  - how big is the arrow up/down clickable area now? We need to make sue that users are able to click it with ease, recommendation is to use 44px clickable area: 
https://share.goabstract.com/cd18018a-ffbf-4132-ac11-fb358cdb9760?collectionLayerId=019853c7-bae3-4f04-bcfa-029c8c8bf504&mode=design

fredefox:
> 44px clickable area

I think we're gonna need to change the design to accommodate this. The existing vote-buttons 24x25. There are multiple other clickable areas present on the page that have a clickable area that is smaller than 44px in either direction. Without fiddling too much with alignment etc. here's a preview of what it would look like if we padded the area around the vote buttons with space so that the clickable area is 44x44.
image

I can of course implement this (sans borders, and with the carets properly aligned) if this is what we want.

fredefox:
I suppose it's actually pretty much in line with the abstract link @DovileM provided. So I guess I'll just go ahead with it.

DovileM:
> I suppose it's actually pretty much in line with the abstract link @DovileM provided. So I guess I'll just go ahead with it.

I guess we just need to center align the icon in the clickable area. 

I know that we might have other areas that need to have bigger clickable areas but let's take it one at the time. We are going to be doing redesign of the theme as well in 2020 H1 I think.

fredefox:
I want to note that the linked design document differs from my implementation. For instance, in my implementation the vote-up container (clickable area) is flush with the avatar. This is not the case in the design. One effect of this is that panel containing the up vote, down vote, cogwheel and vote count will take up more vertical space. Incurring more overall vertical space for short posts. When I tested at scaling=100% this happens for posts with less than 5 lines in a single paragraph or less than 3 single-line paragraphs.

fredefox:
Here's what that looks like with a bit more context:

image

fredefox:
Note that this means for such small comments, the vertical space used increases from 131.125 logical pixels to 169 logical pixels. That's quite an increase.

DovileM:
> Note that this means for such small comments, the vertical space used increases from 131.125 logical pixels to 169 logical pixels. That's quite an increase.

The cog wheel will only be shown to Admins and moderators right? So I think the increase is smaller for major population. Otherwise I think better affordance is better vs the increase in height. People are scrolling anyways. But I will take this into consideration for the new redesign.

fredefox:
> I think better affordance is better vs the increase in height.

I definitely agree with this. I don't personally think it's a problem that it takes up more space. I'm just flagging it. It anyways only applies to small comments. The cited numbers are for one-liner comments.

VilhelmNielsen:
@fredefox Can you please address my comment here: https://github.com/zendesk/help_center/pull/19781#issuecomment-563116043

If this is a breaking change (for the Theming API), then it needs to be scoped to Theming API v2, and this PR would have to go to the `api_v2` branch

VilhelmNielsen:
Alright, thanks for clarifying. Please bump the version number in `manifest.json`

fredefox:
Bumped from 1.9.2 to 1.9.3.

templates/community_post_page.hbs
```diff
@@ -174,9 +182,17 @@
               <div class="comment-actions-container">
                 {{#unless official}}
                   <div class="comment-vote vote" role='radiogroup'>
-                    {{vote 'up' role='radio' class='vote-up' selected_class='vote-voted'}}
+                    {{#vote 'up' role='radio' class='vote-up' selected_class='vote-voted'}}
+                    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16" transform="scale(1,-1)">
+                      <path fill="none" stroke="currentColor" stroke-linecap="round" stroke-width="2" d="M4 6.5l3.6 3.6c.2.2.5.2.7 0L12 6.5"/>
+                    </svg>
+                    {{/vote}}
```
luis-almeida: Can the `vote` helper accept a block already?

fredefox:
I had forgotten the article page. The latest push includes analogous changes as to the post page.

templates/community_post_page.hbs
fredefox: > This changes should follow the following change in Help Center: zendesk/help_center#19781

fredefox:
Note that I forgot to fix the *article* votes. I've done that in https://github.com/zendesk/copenhagen_theme/pull/115/commits/4bbc15643fe6d1f035a9ae4df24c532dd0e81815

fredefox:
> I'm not a big fan of *-selectors, could you add a class instead?

Sure dawg! https://github.com/zendesk/copenhagen_theme/pull/115/commits/a656a365ee0dfd0a416fe5bc2c072354e0002b7a

PR #120, opened 2019-12-10
[PUB-827] Fix a11y issue in Requests page navigation

yara-simargil:
A followup to https://github.com/zendesk/copenhagen_theme/pull/105 fixing same issue in Requests page navigation.


VilhelmNielsen:
Oh, please bump the version in manifest :)

PR #121, opened 2019-12-10
[COMM-350] Change topic and post list items to be sections

VilhelmNielsen:
The ticket only mentions the Topics List page, but since Post List is the same markup, I've made the change there as-well.

Recent Activities should be aligned with this change.

JIRA: https://zendesk.atlassian.net/browse/COMM-350


fredefox:
This is nice, it was always puzzling me why the posts were in an unordered list.

PR #122, opened 2019-12-10
[COMM-352] Fix the heading structure on Community list page

VilhelmNielsen:
I think the heading within a popup button is strange, and that part should probably change, but this is the quickest solution without redesigning the interactions so much, while still addressing the issue.

Don't merge before the strings are translated (same as #117)

JIRA: https://zendesk.atlassian.net/browse/COMM-352

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.12.2",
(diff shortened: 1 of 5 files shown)


fredefox:
Can I haz screenshot?

VilhelmNielsen:
> Can I haz screenshot?

No visual changes, this only affects screen readers.

PR #123, opened 2019-12-10
[COMM-629] Add SVG cog wheel to moderation actions button

fredefox:
[COMM-629]
====
The SVG is copied verbatim from `gear-stroke.svg`[1] from the Zendesk Garden SVG
Icons repository. Another possibility is `gear-fill.svg`[2] from the
same repo.

[1]: https://github.com/zendeskgarden/svg-icons/blob/master/src/16/gear-stroke.svg
[2]: https://github.com/zendeskgarden/svg-icons/blob/master/src/16/gear-fill.svg

This PR contains some of the same changes as can be found in https://github.com/zendesk/copenhagen_theme/pull/115 the PR should be merged after this is merged: https://github.com/zendesk/help_center/pull/19813
 
Note that this change (which also includes https://github.com/zendesk/copenhagen_theme/pull/115) means that the height of short comments increase from 131.125 logical pixels to 160 logical pixels. See also the discussion in the aforementioned PR.

TODO:
===
Replicating the TODO from https://github.com/zendesk/help_center/pull/19813 since some changes might need to go in here:

* [X] The dropdown toggle should ~have `[role=button]`~ be a button.  ~Simply adding
   it will break styling though because we have styling targeting that
   exact property.~
* [ ] There's styling for hovering the cogwheel, but none for the vote-arrows. This should be consistent.
* [x] Similarly, the padding should be consistent for the cog wheel and the vote buttons.

Risk
===
Low; may affect rendering moderation actions button for admins.

[COMM-629]: https://zendesk.atlassian.net/browse/COMM-629


fredefox:
Here's what the post page looks like with my changes:
image

styles/_actions.scss
```diff
@@ -8,22 +8,24 @@
...
-      background-color: transparent;
-      border-radius: 50%;
-      color: $secondary-text-color;
-      content: "\2699";
-      display: block;
-      font-size: $font-size-small;
-      margin: auto;
-      padding: 5px;
-    }
+    @extend .icon;
```
VilhelmNielsen: I think we'll be killing `.icon` when all the icons are SVG. Could you replace it with the necessary styles? (probably just `vertical-align-middle`)

styles/_vote.scss
```diff
@@ -26,29 +24,25 @@
...
-.vote-up:hover::before,
-.vote-down:hover::before {
+.vote-up:hover,
+.vote-down:hover {
   color: $brand_color;
 }
 
-.vote-up::before, .vote-down::before {
+.vote-up, .vote-down {
   @extend .icon;
```
VilhelmNielsen: Same here.

styles/_actions.scss
fredefox: Do you just mean to inline the rules from `.icon`?

fredefox: Sorry, I think I understand. I think the correct thing to do is get rid of `@extend .icon` altogether. I've anyways added:
```
  display: flex;
  align-items: center;
  justify-content: center;
```
To get centering.

fredefox: But perhaps we should make a new class with those three rules? I've added this both to `.vote-up`, `.vote-down`, and `.actions .dropdown-toggle`.

fredefox: I've removed the `@extend .icon` - also for the vote buttons.

styles/_vote.scss
fredefox: Done.

styles/_actions.scss
VilhelmNielsen: Thanks, that's exactly what I meant!

PR #141, opened 2020-01-07
Fix syntax error for rules excluding in lighthouse

TetianaG:
Follow up to #139


TetianaG:
Exclusion of rules lowers the accessibility score.

PR #146, opened 2020-01-28
[COMM-722] Center actions icon

fredefox:


fredefox:
Before: 
image

After: 
image

fredefox:
I don't know why there's the additional change in `style.css` I just ran the compile script.

PR #147, opened 2020-01-28
[COMM-90] Add outline to focused elements

fredefox:


fredefox:
Before: 
image

After: 
image

styles/_base.scss
```diff
@@ -148,3 +144,8 @@ ul {
   width: 1px;
   white-space: nowrap;
 }
+
+:focus {
```
VilhelmNielsen: I don't think this will work.

fredefox: I think you're right. I dun goofed.

augustocravosilva:
Closing this one for now. We might need to revisit focus styles fixes in  the v2 version.

PR #148, opened 2020-01-30
[GG-915] Add notifications styles

VilhelmNielsen:
Adds notification styles into theme.

JIRA: https://zendesk.atlassian.net/browse/GG-915

### DIFF
--- a/styles/_notifications.scss
+++ b/styles/_notifications.scss
@@ -1,3 +1,4 @@
+/***** Notifications *****/
 .notification {
   border: 1px solid;
   display: table;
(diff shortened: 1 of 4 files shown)


styles/_notifications.scss
```diff
@@ -0,0 +1,151 @@
...
+.notification-notice {
+  background: #dbf3ff;
+  border-color: #b5e0f5;
+}
+
+.notification-notice .notification-icon::before {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='12' height='12' focusable='false' viewBox='0 0 12 12'%3E%3Cg fill='none' stroke='%23555555'%3E%3Cpath stroke-linecap='round' stroke-linejoin='round' d='M3.5 6l2 2L9 4.5'/%3E%3Ccircle cx='6' cy='6' r='5.5'/%3E%3C/g%3E%3C/svg%3E");
+}
+
+/* Alert / Lock */
```
luis-almeida: I don't think we need need `notification-lock` as these will stay in HC as callouts.

styles/_notifications.scss
```diff
@@ -0,0 +1,151 @@
+.notification {
```
luis-almeida: Can we add a `/***** Notifications *****/` comment at the top of this partial so it segments this section in the compiled file?

VilhelmNielsen: Amended

styles/_notifications.scss
VilhelmNielsen: I've removed it in a new commit 👍

PR #151, opened 2020-02-05
[COMM-721] Fix flipped vote buttons

fredefox:
# [COMM-721]

Use a CSS declaration rather than `svg[transform]` for rotating the
up-vote.  Fixes an issue where the transformation was not applied in
non-compliant browsers (e.g. IE and Safari).

# Risk
None - simple style change

[COMM-721]: https://zendesk.atlassian.net/browse/COMM-721

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "Copenhagen",
   "author": "Zendesk",
-  "version": "1.14.1",
+  "version": "1.14.2",
   "api_version": 1,
   "default_locale": "en-us",
   "settings": [{
--- a/style.css
+++ b/style.css
@@ -1972,6 +1972,10 @@ ul {
   unicode-bidi: bidi-override;
 }
 
+.vote-up svg {
+  transform: scale(1, -1);
+}
+
 .vote-up:hover,
 .vote-down:hover {
   color: $brand_color;
--- a/styles/_vote.scss
+++ b/styles/_vote.scss
@@ -24,6 +24,10 @@
   }
 }
 
+.vote-up svg {
+  transform: scale(1, -1);
+}
+
 .vote-up:hover,
 .vote-down:hover {
   color: $brand_color;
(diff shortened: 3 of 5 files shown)


fredefox:
Worth considering: Should we be referencing the SVG file in stead of inlining it c.f.: https://www.w3.org/Graphics/SVG/IG/resources/svgprimer.html#SVG_in_HTML

luis-almeida:
> Worth considering: Should we be referencing the SVG file in stead of inlining it c.f.:

Doesn't "look" as good but by inlining them we save quite a few extra http calls.

fredefox:
> Doesn't "look" as good but by inlining them we save quite a few extra http calls.
I see. I'll quote this and close
* https://github.com/zendesk/copenhagen_theme/pull/152 and
* https://github.com/zendesk/copenhagen_theme/pull/153

fredefox:
quite a few = 1, though

luis-almeida:
> quite a few = 1, though

`1` in this PR. We've got several others already among different templates:
https://github.com/zendesk/copenhagen_theme/search?q=svg

Many more others to come:
https://github.com/zendesk/copenhagen_theme/pull/101/files

VilhelmNielsen:
@fredefox:
> Worth considering: Should we be referencing the SVG file in stead of inlining it c.f.: [w3.org/Graphics/SVG/IG/resources/svgprimer.html#SVG_in_HTML](https://www.w3.org/Graphics/SVG/IG/resources/svgprimer.html#SVG_in_HTML)

We've considered [`<use>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use) in the past (and I'd think that would be the way to go), but IE11 doesn't support external resources for `<use>`, and then we'd have to include a sprite sheet with all icons for every page or polyfill it.

It might be worth looking at in the future, but as @luis-almeida said, inlining is the way to go for now.

PR #159, opened 2020-02-17
[GG-1008] Add visually hidden label to quick search field

VilhelmNielsen:


templates/requests_page.hbs
```diff
@@ -35,6 +35,7 @@
 
   {{#form 'requests_filter' id="main-content" class='requests-table-toolbar'}}
     <div class="search">
+      {{label 'query' for='quick-search' class='visibility-hidden'}}
```
luis-almeida: Does this snippet still work after the `id`/`for` swap?:
https://github.com/zendesk/copenhagen_theme/blob/master/script.js#L118

VilhelmNielsen: Yep, still works, since the `id` is unchanged

PR #160, opened 2020-02-17
Add `button` class name to `button-large` elements

VilhelmNielsen:
These button relied on `[role="button"]` styling before, which are no
longer in v2.

### DIFF
--- a/templates/community_post_list_page.hbs
+++ b/templates/community_post_list_page.hbs
@@ -36,7 +36,7 @@
         </span>
       </span>
     <span class="post-to-community">
-      {{link 'new_post' role='button' class='button-large'}}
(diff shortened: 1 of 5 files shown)


templates/community_post_list_page.hbs
```diff
@@ -36,7 +36,7 @@
         </span>
       </span>
     <span class="post-to-community">
-      {{link 'new_post' role='button' class='button-large'}}
+      {{link 'new_post' role='button' class='button button-large'}}
```
augustocravosilva: should these links actually have a `role='button'` . They are links :)

VilhelmNielsen: Good point, they're links!

PR #164, opened 2020-02-18
[GG-1010] Use proper ARIA attributes in search results page

VilhelmNielsen:


templates/search_results.hbs
```diff
@@ -26,12 +26,21 @@
           <h3 class="collapsible-sidebar-title sidenav-title">{{t 'filter_by_help_center'}}</h3>
           <ul class="multibrand-filter-list multibrand-filter-list--collapsed">
             {{#each help_center_filters}}
+              {{#if selected}}
               <li>
-                <a href="{{url}}" class="sidenav-item" aria-selected="{{selected}}">
+                <a href="{{url}}" class="sidenav-item current" aria-current="page">
```
augustocravosilva: why not only using the `if` in the line that changes or inlining the `if`?

VilhelmNielsen: Good point.

VilhelmNielsen: @augustocravosilva 
```
<a href="{{url}}" class="sidenav-item {{#if selected}}current{{/if}}" {{#if selected}}aria-current="page"{{/if}}>
```

Are we moving into "unreadable" territory?

augustocravosilva: We're doing a similar thing in https://github.com/zendesk/copenhagen_theme/blob/cdc996ebf3ef87393bf8eb11dfafa4be8e21a064/templates/community_post_page.hbs#L138
But maybe here we can do
```
{{#if selected}}
<a href="{{url}}" class="sidenav-item current" aria-current="page">
{{else}}
<a href="{{url}}" class="sidenav-item">
{{/if}}
```

VilhelmNielsen: Sure, I've amended the commit.

PR #167, opened 2020-02-18
Aligning with subscribe and actions v2

luis-almeida:
- [x] Adding native support for dropdowns
- [x] Styling actions dropdowns
- [x] Styling subscribe dropdowns


VilhelmNielsen:
Oofa doofa, we need a version bump! 😱

PR #172, opened 2020-02-19
Fix follower count

luis-almeida:


styles/_community.scss
```diff
@@ -90,11 +90,7 @@
   margin-bottom: 10px;
   width: 100%;
 
-  .dropdown {
-    width: 100%;
-  }
-
-  [role="button"] {
+  .button {
```
VilhelmNielsen: Does the button have this class already, or do we need to add it in the template?

styles/_community.scss
```diff
@@ -109,7 +105,7 @@
       color: $brand_text_color;
     }
 
-    &[aria-selected="true"] {
+    &.button-primary {
```
VilhelmNielsen: Same as above

styles/_community.scss
luis-almeida: Yup:
https://github.com/zendesk/copenhagen_theme/blob/api_v2/templates/community_topic_page.hbs#L32
https://github.com/zendesk/copenhagen_theme/blob/api_v2/templates/community_post_page.hbs#L43
https://github.com/zendesk/copenhagen_theme/blob/api_v2/templates/user_profile_page.hbs#L33

PR #182, opened 2020-02-20
Use `aria-pressed` for article vote button style

VilhelmNielsen:
A bit preemptive.

### DIFF
--- a/styles/_buttons.scss
+++ b/styles/_buttons.scss
@@ -26,13 +26,15 @@
   &:hover,
   &:active,
   &:focus,
-  &.button-primary {
+  &.button-primary,
+  &[aria-pressed="true"] {
     background-color: $brand_color;
     color: $brand_text_color;
     text-decoration: none;
   }
 
-  &.button-primary {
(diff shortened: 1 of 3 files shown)


styles/_buttons.scss
```diff
@@ -26,13 +26,15 @@
   &:hover,
   &:active,
   &:focus,
-  &.button-primary {
+  &.button-primary,
+  &[aria-pressed="true"] {
```
luis-almeida: Could we move this to where it is needed only?
I'm thinking that the vote buttons which only show the svg carets might not need all this other styles (?)

luis-almeida:
Actually just remembered we might be base this off from reverting this [revert](https://github.com/zendesk/copenhagen_theme/pull/170) to bring https://github.com/zendesk/copenhagen_theme/pull/169/ back.

luis-almeida:
Hope it's ok I took over in this one: https://github.com/zendesk/copenhagen_theme/pull/184 🙏

PR #183, opened 2020-02-20
Styling follow buttons without using class and selected_class

luis-almeida:


styles/_my-activities.scss
```diff
@@ -109,21 +109,21 @@
...
+        @extend .button;
+        line-height: 40px;
       max-height: 40px;
       padding: 0 20px;
-    }
-
-    .organization-unsubscribe {
-      background-color: $button-color;
-      color: $brand_text_color;
+        &[data-selected="true"] {
```
VilhelmNielsen: Are we changing these to `aria-pressed` later?

luis-almeida: I thought we were discussing `aria-pressed` for votes only.
These changes are related to the follow button/menu-button. When the state changes the label changes as well so not sure `aria-pressed` is appropriate (?) also due to the menu behaviour.

VilhelmNielsen: Ah yes, sorry, I'm mixing things up here. Excuse my fried egg brain 🍳

luis-almeida: No worries! Same here 🍳

PR #195, opened 2020-02-27
Use SVG for logo image

VilhelmNielsen:
We need to specify a width and height, because the SVG doesn't have a
set size, and this will help browsers to prevent content jumping, so
it's a nice change.


settings/logo.svg
```diff
@@ -0,0 +1,15 @@
...
+  </defs>
+  <rect x="16" y="16" width="168" height="168" rx="50" transform="rotate(-45 100 99.998)"/>
+  <rect class="cls-1" x="38.04" y="38.04" width="123.91" height="123.91" rx="30" transform="rotate(-45 100 99.998)"/>
+  <circle class="cls-1" cx="100" cy="100" r="59.4"/>
+  <circle class="cls-1" cx="100" cy="100" r="32.09"/>
+  <rect class="cls-2" x="58.71" y="47.47" width="16" height="38.48" rx="3" transform="rotate(-45 66.711 66.707)"/>
+  <rect class="cls-2" x="123.71" y="47.47" width="16" height="38.48" rx="3" transform="rotate(-135 131.708 66.708)"/>
+  <rect class="cls-2" x="58.71" y="113.47" width="16" height="38.48" rx="3" transform="rotate(45 66.709 132.712)"/>
+  <rect class="cls-2" x="123.71" y="113.47" width="16" height="38.48" rx="3" transform="rotate(135 131.708 132.71)"/>
+</svg>
```
VilhelmNielsen: Looking at this odd markup again, perhaps we should make sure this works in all browsers cc @TetianaG

VilhelmNielsen: I've checked IE11, and the latest Firefox, Safari and Chrome, and it looks fine 👍

PR #201, opened 2020-03-09
Do not lower case 'answered' post status

zdrve:
https://zendesk.atlassian.net/browse/COMM-346

 * Do not lower case the "answered" status of Community Posts
 * *Do* lower case the "pending" status of Classic Requests (like we do for all other statuses of such requests); *do not* lower case the "pending" for Community Content moderation

### DIFF
--- a/styles/_status-label.scss
+++ b/styles/_status-label.scss
@@ -51,7 +51,8 @@
     color: $secondary-text-color;
   }
 
-  &-pending {
+  &-pending,
+  &-pending-moderation {
     background-color: #1f73b7;
     text-align: center;
   }
@@ -77,7 +78,8 @@
   &-closed,
   &-solved,
   &-new,
-  &-hold {
+  &-hold,
+  &-pending {
     text-transform: lowercase; // To be consistent with Lotus
   }
 }
--- a/styles/_status-label.scss
+++ b/styles/_status-label.scss
@@ -77,8 +77,7 @@
   &-closed,
   &-solved,
   &-new,
-  &-hold,
-  &-answered {
+  &-hold {
     text-transform: lowercase; // To be consistent with Lotus
   }
 }
--- a/style.css
+++ b/style.css
@@ -2414,7 +2414,7 @@ ul {
   color: lighten($text_color, 20%);
 }
 
-.status-label-pending {
+.status-label-pending, .status-label-pending-moderation {
   background-color: #1f73b7;
   text-align: center;
 }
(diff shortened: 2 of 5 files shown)


zdrve:
Disclaimer: not yet tested.

styles/_status-label.scss
```diff
@@ -51,7 +51,8 @@
     color: $secondary-text-color;
   }
 
-  &-pending {
+  &-pending,
```
anpa: Giving that we are changing from `status-label-pending` to `status-label-pending-moderation`, do we still need to target `status-label-pending`?

styles/_status-label.scss
```diff
@@ -78,7 +79,7 @@
   &-solved,
   &-new,
   &-hold,
-  &-answered {
+  &-pending {
```
anpa: Should this be `&-pending-moderation`?

styles/_status-label.scss
```diff
@@ -78,7 +79,7 @@
...
     text-transform: lowercase; // To be consistent with Lotus
```
anpa: Not related with this PR but I think this comment is obsolete 😄

styles/_status-label.scss
zdrve: Yes, because the whole point of the rename is that status-label-pending and status-label-pending-moderation are different things (that happen to have identical rendering).

The former relates to lotus requests; the latter relates to guide community posts.

styles/_status-label.scss
zdrve: No; see above.

styles/_status-label.scss
zdrve: No, I don't think it is. Well at least it might be, but I don't see how that can be concluded from here.

styles/_status-label.scss
anpa: Gotcha! Couldn't find a reference for `.status-label-pending`, but we have it indeed for requests: https://github.com/zendesk/copenhagen_theme/blob/master/templates/requests_page.hbs#L125

zdrve: `status-label-pending` is referenced purely from the templates within this theme.

styles/_status-label.scss
anpa: `Support` doesn't lowercase labels anymore. But this should be addressed separately as part of a different issue: https://zendesk.atlassian.net/browse/HCJ-2459

zdrve: 👍

anpa:
Updated commit messages to follow the [conventional commit standard](https://www.conventionalcommits.org/en/v1.0.0/)

PR #202, opened 2020-03-16
fix(header): remove size attributes from logo img

VilhelmNielsen:
## Description

Explicit size properties on `img` tag for the logo forced bigger logos to be squished. Since logos are a setting, it should be modifiable without code changes. Sizing properties are moved to the SVG, as the logo image should decide its aspect ratio.

## Checklist

- [ ] ~:nail_care: SASS files are compiled~

### DIFF
--- a/settings/logo.svg
+++ b/settings/logo.svg
@@ -1,4 +1,4 @@
-<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 200">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 200" width="37" height="37">
   <defs>
     <style>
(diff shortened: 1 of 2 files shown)


augustocravosilva:
Won't this cause large images to not be scaled down?
Might affect existing themes in a bad way

luis-almeida:
> Won't this cause large images to not be scaled down?

Wasn't that the main issue since the dimensions were hardcoded in the html?
There is already a set `max-height`: https://github.com/zendesk/copenhagen_theme/blob/master/style.css#L616-L618

> Might affect existing themes in a bad way

The asset lives in the theme. How would it impact existing themes?

augustocravosilva:
Wasn't aware we kept the max-heigh after hardcoding
In that case, all good!

VilhelmNielsen:
@augustocravosilva This is what we had before, minus the attributes on the SVG. As @luis-almeida points out, we have a `max-height` rule for images taller than that.

PR #211, opened 2020-04-02
[GG-1121] fix(header): Make sign in link visible on mobile

VilhelmNielsen:
## Description

Currently, the sign in link is not visible on mobile, since we went away from `{{user_info}}`. This re-aligns the old styling rules.

## Checklist

- [ ] :memo: changes are tested in Chrome, Firefox, Safari, Edge, and IE11

### DIFF
--- a/style.css
+++ b/style.css
@@ -673,6 +673,22 @@ ul {
   display: inline-block;
 }
 
+@media (max-width: 768px) {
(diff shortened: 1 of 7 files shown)


VilhelmNielsen:
@luis-almeida @anpa I didn't account for hiding the user name on mobile, so I've added commits for that; can you have another look?

PR #217, opened 2020-05-29
Add split button support and apply the new user profile {{actions}} helper

Anifacted:
## Description

In order to support multiple actions on the user profile page, as new features are arriving, we need to display a split button of multiple actions rather than the current single edit profile button.

## Screenshots

image

## Checklist

### DIFF
--- a/styles/_user-profiles.scss
+++ b/styles/_user-profiles.scss
@@ -89,14 +89,9 @@
   }
 }
 
-.edit-profile-button a {
-  @extend .button;
-  background-color: $brand_color;
-  border: 0;
-  color: $brand_text_color;
-  line-height: normal;
-  padding: 8px 20px;
-  outline-color: $brand_color;
+.user-profile-actions {
+  width: 100%;
+  margin-bottom: 15px;
 }
 
 .profile-header .description {
@@ -539,7 +534,7 @@
   text-align: center;
   color: #ccc;
 
- [dir="rtl"] & {
+  [dir="rtl"] & {
     right: 0;
   }
 
--- a/templates/user_profile_page.hbs
+++ b/templates/user_profile_page.hbs
@@ -28,9 +28,7 @@
             </svg>
           </span>
         {{/if}}
-        <span class="edit-profile-button">
-          {{edit}}
-        </span>
+        {{actions class='user-profile-actions split-button'}}
         <div class="community-follow">
           {{subscribe}}
         </div>
--- a/style.css
+++ b/style.css
@@ -345,7 +345,7 @@ ul {
 }
 
 /***** Buttons *****/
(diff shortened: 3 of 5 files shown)


templates/user_profile_page.hbs
```diff
@@ -28,9 +28,9 @@
             </svg>
           </span>
         {{/if}}
-        <span class="edit-profile-button">
-          {{edit}}
-        </span>
+        <div class="user-profile-actions">
+          {{actions}}
```
VilhelmNielsen, line 32: In this case, `actions` could take `className` as a parameter, instead of wrapping the button, which we did for legacy concerns.

Anifacted: 👍 
Would you still wrap it in a `div` then, or?

VilhelmNielsen: No, since the helper already outputs a div

Anifacted: I pushed an update. And on the corresponding HC branch.

templates/user_profile_page.hbs
```diff
@@ -28,9 +28,7 @@
             </svg>
           </span>
         {{/if}}
-        <span class="edit-profile-button">
-          {{edit}}
-        </span>
+          {{actions class='user-profile-actions split-button'}}
```
zdrve, line 31:
```suggestion
        {{actions class='user-profile-actions split-button'}}
```

styles/_user-profiles.scss
```diff
@@ -99,6 +99,11 @@
   outline-color: $brand_color;
 }
```
augustocravosilva, line 100: This rule can be deleted now

Anifacted: It's not really part of my changes, though?

Anifacted: oh, you mean the entire CSS class

Anifacted: But we still support `{{edit}}` for the time being, so shouldn't we keep it around?

augustocravosilva: yes, class sorry, as the `.edit-profile-button` will no longer be used

augustocravosilva: > But we still support {{edit}} for the time being, so shouldn't we keep it around?

If it is not used in the CPH theme we don't need to have it here. Custom themes keep their own CSS.

Anifacted: Ah, makes sense 👍

Anifacted: Pushed the removal of it

TetianaG:
@Anifacted It works for admin when admin opens its own profile. But when the admin opens the end user profile it sees the following: 
Screenshot 2020-06-22 at 17 06 40

Also I am not sure about the responsiveness on mobile. Does it look correct?
Screenshot 2020-06-22 at 17 11 08

Anifacted:
@TetianaG The second screenshot looks correct to me. For the first screenshot, are you saying that the menu is not opening at all when you click the chevron button?

Anifacted:
(Otherwise, nothing seems wrong to me)

PR #221, opened 2020-06-11
Update if statement to check for type filters instead of community

cmacaulay:
## Description

- Fixes bug where `type_filters` are nested in a check for community, meaning that those without communities but with external record types wouldn't see the results properly. 

## Checklist

### DIFF
--- a/templates/search_results.hbs
+++ b/templates/search_results.hbs
@@ -41,7 +41,7 @@
           </ul>
         </section>
       {{/if}}
-      {{#if help_center.community_enabled}}
+      {{#if type_filters}}
         <section class="filters-in-section collapsible-sidebar" aria-expanded="false">
           <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false">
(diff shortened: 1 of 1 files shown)


templates/search_results.hbs
```diff
@@ -41,7 +41,7 @@
           </ul>
         </section>
       {{/if}}
-      {{#if help_center.community_enabled}}
+      {{#if type_filters}}
```
ronan-mch, line 44: Do you know how truthiness works in CurlyBars? Will this return true if `type_filters` is an empty array?

ronan-mch: Answered my own question, an empty collection [resolves to false](https://github.com/zendesk/curlybars/blob/master/docs/templates.md#how-conditions-are-evaluated).

cmacaulay: We do the same for `source_filters` up above 😸

ronan-mch:
There's one thing I'm a bit confused about. The behaviour for `source_filters` is to display only if there is more than one source. Am I understanding correctly that this behaviour is inconsistent, i.e. type filters will display even if there is only one type available?

PR #228, opened 2020-08-14
!fixup [COMM-1046] Achievement widget

fredefox:
[COMM-1046]
===

TODO
===

* [ ] Only consider badges with `category == "achievements"`
* [ ] Do not show on user profile page
* [ ] Show in the correct places (which?)
* [ ] Dynamically generate the "+n" icon.
* [ ] Use compatible JS-syntax.
* [ ] Don't actually output the slug as part of the generated HTML.

[COMM-1046]: https://zendesk.atlassian.net/browse/COMM-1046

### DIFF
--- a/script.js
+++ b/script.js
@@ -397,4 +397,43 @@ document.addEventListener('DOMContentLoaded', function() {
       }
     });
   });
+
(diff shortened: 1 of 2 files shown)


Anifacted:
Closing in favor of an implementation using curlybars expressions

PR #242, opened 2020-11-16
feat: add subsections pagination to section template

TetianaG:
## Description

Removing "See all" subsections and adding subsections pagination to Section template.

## Checklist

### DIFF
--- a/script.js
+++ b/script.js
@@ -190,19 +190,6 @@ document.addEventListener('DOMContentLoaded', function() {
     });
   }
 
-  // If a section has more than 6 subsections, we collapse the list, and show a trigger to display them all
-  var seeAllTrigger = document.querySelector("#see-all-sections-trigger");
-  var subsectionsList = document.querySelector(".section-list");
-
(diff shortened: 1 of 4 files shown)


augustocravosilva:
I believe there's some related JS and CSS that can be removed as well

augustocravosilva:
Please re-write the existing pagination as `{{pagination "section.articles"}}` to be consistent

augustocravosilva (approved):
Be aware that this will cause up to _max subsections per page_ (30?) to show right away instead of 6

TetianaG:
The current limit is 20 subsections.

luis-almeida:
> The current limit is 20 subsections.

Yes but from what I understand the JavaScript snippet was collapsing the list of subsections to only show up to 6.

TetianaG:
> > The current limit is 20 subsections.
> 
> Yes but from what I understand the JavaScript snippet was collapsing the list of subsections to only show up to 6.

Yes, but the maximum number of possible subsections is 20 right now.

PR #245, opened 2020-12-01
fix(script): prevent TypeError with null element

syntaxval:
## Description

When there are no matching textarea element that the query selector is trying to get, the script would throw TypeError when accessing `dataset` property of a (null)`requestCommentTextarea`

## Checklist


VilhelmNielsen:
Hi @syntaxval , thank you for bringing this to our attention. We've applied a fix in our code style in #246, so I'll close this one 🙂

PR #248, opened 2020-12-14
Visual update

kaspersorensen:
## Description

Updates to the visual appearance of the Copenhagen Theme as per ROAD-8816. This change includes the following changes:

* New logo and favicon
* New images
* Updated default colors
* A new setting for showing the Help Center name next to the logo
* Removed the "visited link" text color adjustments
* Added margin above the "Join the conversation" link.
* New theme thumbnail

## Screenshots

At the moment the theme can be viewed on this test HC: https://z3n-kaspersor.zendesk.com/hc/en-us

Main front page:

image

Community front page:

image

## Checklist


translations/en-us.json
```diff
@@ -4,6 +4,8 @@
   "brand_color_label": "Brand color",
   "brand_text_color_description": "Brand color for hover and active states",
   "brand_text_color_label": "Brand text color",
+  "brand_name_label": "Show Help Center name",
+  "brand_name_description": "Display the Help Center name next to the logo",
```
anpa, line 8: The `translations/*.json` files are generated automatically, we should not update them manually. We can instead run `node bin/update-translations.js` in the project root. That will download all translations and add them to the json files :)

translations.yml
```diff
@@ -349,9 +349,9 @@ parts:
       key: "txt.help_center_copenhagen_theme.show_brand_name_label"
       title: "Label for the show brand name setting"
       screenshot: "https://drive.google.com/file/d/16zQao3zk3uRScIUpkyc6bXAWwqHh3Zh-/view?usp=sharing"
-      value: "Show Help Center name"
+      value: "Show brand name"
```
anpa, line 352: I don't think we can update these strings directly since they were already submitted for translation (see [here](https://zendesk.atlassian.net/wiki/spaces/globalization/pages/742069700/Edit+Translations)). Can you please double check it with @zendesk/localization ? We usually mark strings as obsolete and add new ones instead.

templates/header.hbs
```diff
@@ -3,7 +3,10 @@
 <header class="header">
   <div class="logo">
     {{#link 'help_center'}}
-      <img src="{{settings.logo}}" alt="{{t 'home_page' name=help_center.name}}">
+      <img src="{{settings.logo}}" alt="{{t 'home_page' name=help_center.name}}" />
+      {{#if settings.show_brand_name}}
+        <span>{{help_center.name}}</span>
```
anpa, line 8: For the home page, we have an [hidden element](https://github.com/zendesk/copenhagen_theme/blob/master/templates/home_page.hbs#L1) with the help center name as `h1` for screen readers to read it. This change would make the screen reader read it twice - not sure if it's an issue though. @VilhelmNielsen what do you think?

styles/_variables.scss
```diff
@@ -1,13 +1,13 @@
 // Variables
 
 // Colors
-$brand_color: #0072ef !default;
+$brand_color: rgba(23, 73, 77, 1) !default;
```
anpa, line 4: nit: for consistency, let's use the hex values here
```suggestion
$brand_color: #17494d !default;
```

styles/_variables.scss
```diff
@@ -1,13 +1,13 @@
...
 $brand_text_color: #fff !default; //Color for text with an accent color background
 $text_color: #333 !default;
-$link_color: #0072ef !default;
+$link_color: rgb(23, 73, 77) !default;
```
anpa, line 7: Same here:

```suggestion
$link_color: #17494d !default;
```

styles/_variables.scss
```diff
@@ -1,13 +1,13 @@
...
 $background_color: #fff !default;
 
-$visited-link-color: zass-darken($link_color, 20%);
+$visited-link-color: $link_color;
```
anpa, line 10: Does this mean we won't have a different color for visited links?

kaspersorensen: Yes. This was asked for by design. CC @niksavic

templates/header.hbs
VilhelmNielsen: Good catch, plus the alt text for the logo would make assistive technology announce the link as, eg: “Mondocam Help Center home page Mondocam.“

It's probably worth it to slap an `aria-hidden=true` on the `span` tag.

manifest.json
```diff
@@ -365,8 +372,8 @@
         {
           "identifier": "show_follow_topic",
           "type": "checkbox",
-          "description": "follow_topic_description",
-          "label": "follow_topic_label",
+          "description": "show_follow_topic_description",
```
anpa, line 375: Not sure I understand this change. Would you like to update these strings as well? If so, both `description` and `label` values should match keys in [translations.yml](https://github.com/zendesk/copenhagen_theme/blob/master/translations.yml#L344).

translations/en-us.json
kaspersorensen: Will fix this. I don't know why it did this. Think it was a merge conflict that I fixed wrong then.

kaspersorensen:
@anpa I updated the branch now based on the feedback. I ran the "update-translations.js" script and it generated updates to the translation files. I chose not to squash that commit, so that you can see them separately as the translation part seems like a chore-commit thing.

kaspersorensen:
Hold on, the setting is not showing correctly now:
image

VilhelmNielsen:
As I noted in https://github.com/zendesk/copenhagen_theme/pull/252, the strings do not seem to be translated.

styles/_variables.scss
anpa: Can we consider keeping a different color? This functionality helps users understanding if they've already navigated to a link https://www.nngroup.com/articles/change-the-color-of-visited-links/

styles/_header.scss
```diff
@@ -12,6 +12,16 @@ $header-height: 71px;
 
 .logo img {
   max-height: 37px;
+  vertical-align: middle;
+}
+
+.logo span {
+  margin: 0 10px 0 10px;
```
anpa, line 19: nit: we can use the shorthand here

```suggestion
  margin: 0 10px;
```

style.css
```diff
@@ -661,6 +661,16 @@ ul {
...
+  vertical-align: middle;
+}
+
+.logo span {
+  margin: 0 10px 0 10px;
+  color: $brand_color;
+}
+
+.logo a:hover {
+  text-decoration: none;
```
anpa, line 673: We should also target `:focus` and `:active` states as they will also show an underline.

i.e.: on focus (user navigates with the keyboard - keeps pressing TAB):

Screenshot 2021-01-11 at 13 37 13

I also suggest updating the `.logo a` to use `display: inline-block;` so the focus outline becomes a rectangle instead:

Screenshot 2021-01-11 at 13 40 03

styles/_header.scss
```diff
@@ -12,6 +12,16 @@ $header-height: 71px;
...
+  color: $brand_color;
+}
+
+.logo a:hover {
+  text-decoration: none;
```
anpa, line 24: We should also target `:focus` and `:active` states as they will also show an underline.

i.e.: on focus (user navigates with the keyboard - keeps pressing TAB):

Screenshot 2021-01-11 at 13 37 13

I also suggest updating the `.logo a` to use `display: inline-block;` so the focus outline becomes a rectangle instead:

Screenshot 2021-01-11 at 13 40 03

translations.yml
magalimarquetzen: I don't see a yml here anymore? but indeed please obsolete and create a new key for those types of changes :)

augustocravosilva:
@kaspersorensen the new images could use some optimization to improve loading time 🙏

kaspersorensen:
@augustocravosilva good point. How much do we normally compress? Is this up to design or do we have a convention? I can do it in Gimp, just don't know how much to turn the dial :-)

augustocravosilva:
2000 px width should be enough for the banners
1500 px for the other image

thanks

kaspersorensen:
OK, all the comments have been addressed and resolved. I have also reapplied the theme on https://z3n-kaspersor.zendesk.com/ in case you want a quick preview.

styles/_variables.scss
```diff
@@ -1,13 +1,13 @@
...
-$brand_color: #0072ef !default;
+$brand_color: #17494d !default;
 $brand_text_color: #fff !default; //Color for text with an accent color background
 $text_color: #333 !default;
-$link_color: #0072ef !default;
+$link_color: #17494d !default;
 $background_color: #fff !default;
 
-$visited-link-color: zass-darken($link_color, 20%);
+$visited-link-color: $visited_link_color;
```
anpa, line 10: nit: it seems we no longer use this variable, we should be able to remove it 🎉

anpa (approved):
Looks great! Thanks for this nice update! 🎨 

👍 once the translations are in

anpa (commented):
Sorry I forgot 🙈 Can you please squash the commits and [update the commit message to use `feat(settings):`](https://github.com/zendesk/copenhagen_theme#commit-types)? This way the theme will be released as a minor version once it's merged 🥳

kaspersorensen:
Good trick @anpa. I updated the branch and added the feat prefix to the commit msg.

kaspersorensen:
Oh but wait, we're still missing the new translation strings for the key that is used in the setting for visible link colors.

manifest.json
```diff
@@ -205,6 +212,13 @@
           "description": "logo_description",
           "label": "logo_label"
         },
+        {
+          "identifier": "show_brand_name",
+          "type": "checkbox",
+          "description": "show_brand_name_description",
+          "label": "show_brand_name_label",
+          "value": true
```
augustocravosilva, line 220: for new settings the recommendation is to first add it with a default value of what exists in current themes (in this case false) and then swap the value after an update has been delivered to all accounts. Otherwise existing themes will start showing unexpected new content 😬

kaspersorensen: OK I will change this to false as a temporary value for the rollout.

rdiaszen:
I don't see the yml file. Please let me know if should check anything else related to strings

kaspersorensen:
Hi @rdiaszen 
The yml file was merged separately in #253

manifest.json
```diff
@@ -36,6 +36,13 @@
           "label": "link_color_label",
           "value": "#1F73B7"
         },
+        {
+          "identifier": "visited_link_color",
+          "type": "color",
+          "description": "visited_link_color_description",
+          "label": "visited_link_color_label",
+          "value": "#4B61C3"
```
anpa, line 44: @kaspersorensen @augustocravosilva should we also update this value to be the current default, since this is also a new setting?

The current default depends on the [link color of each theme](https://github.com/zendesk/copenhagen_theme/pull/248/files#diff-bc9dff5ad00c2491d2b9619883dea5d5e508fec2aebbf3a77d70087d2cdafd4bL10) though 🤔

kaspersorensen: I think this one is much less intrusive. I would be fine defending this change to customers who complain. The fix for them is really easy and it's "just" the color of visited links, not as visible a change as the brand name showing up next to the logo.

augustocravosilva: Yeah, i was wondering if this setting is even necessary. I don't think many, if any, existing customers using the cph theme would notice.
Alternatively, I would suggest this setting could be a range [0-50] as the % of darkness compared to the normal link color. this would allow existing customers to keep it at 20% darkness and the relation between the link color and visited color would be kept every time they change the link color.

kaspersorensen: We don't want it to be tied to the normal link color. It's much more flexible to have it independent.

kaspersorensen:
Once #255 is merged, I claim this PR is ready to merge as well.

anpa (approved):
Thanks for all the updates! 👏 ❤️

tyteen4a03:
Apologies for bumping this, but the introduction of `visited_link_color` variable was missing its counterpart `$visited_link_color` in `styles/_variables.scss`, causing my IDE to report "variable not found" errors.

I have raised a bug report with Zendesk.

PR #283, opened 2021-11-24
Navbar changes

JakubKopys:
## Description

1. Merges two navigations for mobile. Currently, we have a hamburger menu and a dropdown menu that's displayed when the user clicks on the avatar.
2. Change the breakpoint for the mobile navbar. Currently, the mobile navbar is displayed only for mobile (width <= 768px). After these changes, mobile navbar will be displayed on both mobile and tablet viewports (width <= 1024px)
3. Additionally, fixes the a11y issue with desktop version of nav - all items are now displayed inside a list.

## Screenshots

New navbar - admin:
https://user-images.githubusercontent.com/12835055/144049443-9870aaf6-cd5d-4cb2-8a03-e7048b215773.mov
New navbar - end user:
https://user-images.githubusercontent.com/12835055/144049466-8b5d1496-7883-43de-b205-0286ac5dc4b3.mov

Screenshot 2021-11-30 at 13 40 49
Screenshot 2021-11-30 at 13 40 32
Screenshot 2021-11-30 at 13 40 19
Screenshot 2021-11-30 at 13 40 10

## Checklist

### DIFF
--- a/styles/_header.scss
+++ b/styles/_header.scss
@@ -50,6 +50,15 @@ $header-height: 71px;
   }
 }
 
+.user-nav-list {
+  display: block;
+  list-style: none;
+
+  > li {
+    display: inline-block;
+  }
+}
+
 .nav-wrapper {
   a {
     @include tablet {
--- a/styles/_header.scss
+++ b/styles/_header.scss
@@ -47,6 +47,17 @@ $header-height: 71px;
       display: block;
       margin: 20px;
     }
+
+    > .user-nav-list {
+      li {
+        display: block;
+      }
+
+      a {
+        display: block;
+        margin: 20px;
+      }
+    }
   }
 }
 
@@ -59,7 +70,11 @@ $header-height: 71px;
   }
 }
 
-.nav-wrapper {
+.nav-wrapper-desktop {
+  @include mobile { display: none; }
+  @include tablet { display: none; }
+  @include desktop { display: inline-block; }
+
   a {
     @include tablet {
       display: inline-block;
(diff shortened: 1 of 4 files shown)


sergioazevedo (approved):

styles/_header.scss
```diff
@@ -72,42 +96,123 @@ $header-height: 71px;
...
+      padding: 8px 24px;
+
+      width: 100%;
+      height: 100%;
+      color: $text_color;
+      cursor: pointer;
+      text-align: start;
+
+      &:active, &:focus, &:hover {
+        background-color: rgba(31, 115, 183, 0.1);
```
anpa, line 177: We've been using the `#f3f3f3` color when hovering [dropdown items](https://github.com/zendesk/copenhagen_theme/blob/master/styles/_dropdowns.scss#L77). Example:

Screenshot 2021-12-08 at 14 48 11

Using a different color here will make them inconsistent.

Another issue is that, by changing from a neutral color (gray) to a shade of blue, it may not play well with themes that are branded with different colors.

To unblock this PR, I recommend using `#f3f3f3` here and add this change in a separate PR. Meanwhile, let's check with design if:
- if we should use a neutral color here or if this color should be part of the theme settings;
- this change should be applied to all dropdowns in the theme;

styles/_header.scss
```diff
@@ -72,42 +96,123 @@ $header-height: 71px;
...
+      width: 30px;
+      height: 30px;
     }
 
-    [dir="rtl"] & {
-      margin-left: 10px;
-      margin-right: 0;
+    &[aria-expanded="true"] {
+      .icon-menu {
+        background: rgba(31, 115, 183, 0.2);
```
anpa, line 121: Same question with this icon background color.

styles/_header.scss
```diff
@@ -72,42 +96,123 @@ $header-height: 71px;
...
+    min-width: 71px;
+    cursor: pointer;
 
     .icon-menu {
+      padding: 7px;
       vertical-align: middle;
-      width: 13px;
-      height: 13px;
+      width: 30px;
+      height: 30px;
```
anpa, line 116: Bigger icon, ftw! ❤️

styles/_header.scss
```diff
@@ -72,42 +96,123 @@ $header-height: 71px;
...
+  .menu-list-mobile {
+    position: absolute;
+    background-color: #fff;
+    box-shadow: 0 10px 10px 0 rgba(0, 0, 0, .15);
+    border-top: solid 0.1px #ddd;
+    border-bottom: solid 1px #ddd;
+    right: 0;
+    left: 0;
+    top: $header-height;
+    z-index: 2;
```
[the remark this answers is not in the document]
JakubKopys: We can't :/ This results in search icon to overlap with the nav https://github.com/zendesk/copenhagen_theme/blob/f5de0066b9e000f72da06a0af34e088606b33afb/style.css#L1099

This didn't occur before because the mobile navbar had less item, and therefore, was smaller

styles/_header.scss
```diff
@@ -72,42 +96,123 @@ $header-height: 71px;
...
-    @include tablet { display: none; }
+  .menu-button-mobile {
     background: none;
     border: 0;
-    color: $link_color;
-    display: inline-block;
-    margin-right: 10px;
-    padding: 0;
     width: auto;
+    min-width: 71px;
```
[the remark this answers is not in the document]
JakubKopys: image

I wanted to avoid that on smaller screens.

templates/header.hbs
```diff
@@ -9,15 +9,20 @@
...
+        <li>{{link 'new_request' class='submit-a-request'}}</li>
+        {{#unless signed_in}}
+          <li>
+            {{#link "sign_in" class="sign-in"}}
+              {{t 'sign_in'}}
+            {{/link}}
+          </li>
+        {{/unless}}
+      </ul>
     </nav>
```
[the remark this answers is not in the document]
JakubKopys: Well, ideally probably this should be in a nav item, but I've decided that's out of the scope of this PR (I've talked about that with Luis a couple of weeks ago and he agreed). This PR got pretty big anyway 😅 

As a reminder - initially, this was just a small task to put [those links](https://github.com/zendesk/copenhagen_theme/blob/master/templates/header.hbs#L19-L20) in a list because that got flagged during A11y review.

To sum up - we have only 1 week left to ship it, I don't think adding even more improvements is a good idea in this situation. I suggest we put this improvement in our backlog if we think that's important though.

styles/_header.scss
JakubKopys: Good catch, I've forwarded this comment to design

PR #289, opened 2022-01-19
Add separator in user dropdown

mancristiana:
## Description

Added a missing separator between Sign out and Change password options in the user dropdown menu.
Issue: https://zendesk.atlassian.net/browse/COMM-1599

## Screenshots
Before
Screenshot 2022-01-20 at 14 47 40

After
Screenshot 2022-01-20 at 14 47 24

## Checklist

### DIFF
--- a/styles/_dropdowns.scss
+++ b/styles/_dropdowns.scss
@@ -47,13 +47,12 @@
   }
 
   [role="separator"] {
-    border-bottom: 1px solid rgb(216, 216, 216);
+    border-bottom: 1px solid $border-color;
     color: rgb(150, 150, 150);
     display: block;
     font-weight: normal;
     font-size: 11px;
-    padding: 5px 0;
-    margin: 5px 20px 10px 20px;
+    margin: 4px 0;
   }
 
   [role="menuitem"] {
--- a/styles/_dropdowns.scss
+++ b/styles/_dropdowns.scss
@@ -47,11 +47,7 @@
   }
 
   [role="separator"] {
-    border-bottom: 1px solid $border-color;
-    color: rgb(150, 150, 150);
-    display: block;
-    font-weight: normal;
-    font-size: 11px;
+    border-bottom: 1px solid #e9ebed;
     margin: 4px 0;
   }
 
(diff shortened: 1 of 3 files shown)


JakubKopys (commented):
The task shows a divider with a full-width divider and this divider is narrower. Is that fine? Has that been consulted with the design (or, should it be?)? 🤔

mancristiana:
@JakubKopys Since this narrower separator style was already defined in the theme as part of the dropdown style, I expect it was already approved by design. I opted for the existing style for consistency.
I will consult with design, however, if the separator design should change in this case and be full width, shouldn't all separators in dropdowns be full width?
@folajimia who created the issue said it was fine with him to go with the narrower separator.

JakubKopys:
> @JakubKopys Since this narrower separator style was already defined in the theme as part of the dropdown style, I expect it was already approved by design. I opted for the existing style for consistency. I will consult with design, however, if the separator design should change in this case and be full width, shouldn't all separators in dropdowns be full width? @folajimia who created the issue said it was fine with him to go with the narrower separator.
> 
> UPDATE: According to designers, we should use the full width divider because it should match the style in Zendesk Garden. Thus, I will update the style of the dropdown separator. This will affect all separators in dropdown menus.

It might be worth checking if those changes don't break the mobile version of the navigation 🙇

mancristiana:
> It might be worth checking if those changes don't break the mobile version of the navigation 🙇
For the user dropdown menu there are two separate markups for mobile and desktop. Thus, the changes on desktop don't affect the mobile. The separator role in dropdown is not used anywhere else, so no accidental style changes there.

styles/_dropdowns.scss
```diff
@@ -47,13 +47,12 @@
   }
 
   [role="separator"] {
-    border-bottom: 1px solid rgb(216, 216, 216);
```
JakubKopys, line 50: Nit: why do we change this colour? Looks like we might it slightlyyyy lighter, right? Is that intentional?

JakubKopys (approved):

styles/_dropdowns.scss
mancristiana: Good question. The color should match the design in Figma https://www.figma.com/file/6g87L4FdKZTA3knt3Rsfdx/Garden?node-id=254%3A1241 
which is #e9ebed. If anything we should change it to that color.

JakubKopys (approved):

PR #296, opened 2022-02-07
fix: add aria labels to search result menus

JakubKopys:
## Description

Add aria-label attributes to search result menu toggle buttons

## Checklist

### DIFF
--- a/templates/search_results.hbs
+++ b/templates/search_results.hbs
@@ -15,7 +15,7 @@
     <aside class="search-results-sidebar">
       {{#if source_filters}}
         <section class="filters-in-section collapsible-sidebar" aria-expanded="false">
-          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false">
+          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false" aria-label="{{t 'search_result_source_menu'}}">
             <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" focusable="false" viewBox="0 0 12 12" aria-hidden="true" class="collapsible-sidebar-toggle-icon chevron-icon">
               <path fill="none" stroke="currentColor" stroke-linecap="round" d="M3 4.5l2.6 2.6c.2.2.5.2.7 0L9 4.5"/>
             </svg>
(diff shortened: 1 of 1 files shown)


luis-almeida (commented):
Are these new strings?

templates/search_results.hbs
```diff
@@ -41,9 +41,9 @@
           </ul>
         </section>
       {{/if}}
-      {{#if type_filters}}
+      {{#if true}}
```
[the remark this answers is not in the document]
JakubKopys: Oh, good catch

luis-almeida:
Sorry should have added n inline comment instead.
Are these new strings? If so we'll need to add them to the api beforehand.

templates/search_results.hbs
```diff
@@ -15,7 +15,7 @@
     <aside class="search-results-sidebar">
       {{#if source_filters}}
         <section class="filters-in-section collapsible-sidebar" aria-expanded="false">
-          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false">
+          <button aria-label="{{t 'search_result_source_menu'}}" type="button" class="collapsible-sidebar-toggle" aria-expanded="false">
```
mancristiana, line 18: Nitpick: I know we don't have a consistent guideline for this, but it might be easier to read the code in the future if we order attributes https://codeguide.co/#html-attribute-order
<button class="collapsible-sidebar-toggle" type="button"  aria-expanded="false" aria-label="{{t 'search_result_source_menu'}}">

templates/search_results.hbs
```diff
@@ -43,7 +43,7 @@
       {{/if}}
       {{#if type_filters}}
         <section class="filters-in-section collapsible-sidebar" aria-expanded="false">
-          <button type="button" class="collapsible-sidebar-toggle" aria-expanded="false">
+          <button aria-label="{{t 'search_result_type_menu'}}" type="button" class="collapsible-sidebar-toggle" aria-expanded="false">
```
mancristiana, line 46: Nitpick: The translation key might be misleading since this is not a menu, but a button. Consider naming this `search_result_type_filters_expand_button`

JakubKopys: The aria-label will be "Search Result Type Menu" as suggested by the reporter in the task https://zendesk.atlassian.net/jira/software/c/projects/PRODA11Y/boards/3358?modal=detail&selectedIssue=PRODA11Y-219, so I think this should be fine?

luis-almeida: What is the value of this string? To make them more user friendly we often try to name them closely to how they read like in english, e.g. `add_to_conversation` -> `Add to conversation`

JakubKopys: `Search Result Type Menu`, `Search Result Source Menu` and `Search Result Subfilters Menu`, so that's in line with what you said

luis-almeida (approved):
When the string become available in the API 👍🏼

PR #299, opened 2022-02-10
fix: add aria-label to the '+ N more badges' links

JakubKopys:
## Description

Adds aria-label to the "+ n more" badges link.

## References
- https://zendesk.atlassian.net/browse/PRODA11Y-247

## Checklist

### DIFF
--- a/templates/article_page.hbs
+++ b/templates/article_page.hbs
@@ -250,7 +250,9 @@
 
                             {{#if (compare (calc author.badges.length "-" 4) ">" 0)}}
                               <div class="community-badge community-badge-achievements">
-                                {{#link "user_profile" id=author.id filter_by="badges" class="community-badge-achievements-rest"}}{{t 'plus_more' count=(calc author.badges.length "-" 4)}}{{/link}}
+                                <a href="{{page_path "user_profile" id=author.id filter_by="badges"}}" class="community-badge-achievements-rest"  aria-label="{{t 'more_awards_to' count=(calc author.badges.length "-" 4) name=author.name}}">
+                                  {{t 'plus_more' count=(calc author.badges.length "-" 4)}}
+                                </a>
                               </div>
                             {{/if}}
                           </div>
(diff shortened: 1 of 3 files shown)


luis-almeida (approved):
Once the `more_awards_to` string is available 👍🏼

templates/article_page.hbs
```diff
@@ -250,7 +250,9 @@
 
                             {{#if (compare (calc author.badges.length "-" 4) ">" 0)}}
                               <div class="community-badge community-badge-achievements">
-                                {{#link "user_profile" id=author.id filter_by="badges" class="community-badge-achievements-rest"}}{{t 'plus_more' count=(calc author.badges.length "-" 4)}}{{/link}}
+                                <a href="{{page_path "user_profile" id=author.id filter_by="badges"}}" class="community-badge-achievements-rest"  aria-label="{{t 'more_awards_to' name=author.name}}">
+                                  {{t 'plus_more' count=(calc author.badges.length "-" 4)}}
```
mancristiana, line 254: Shouldn't the `aria-label` contain the count as well? 
I'm asking since `aria-label` replaces the content of the link in the accessibility tree. So assistive technology would read out something along the lines of "more awards to Jakub Kopys" instead of "3 more awards to Jakub Kopys". So since the number that is visible on the page would not be read, users with low visibility might be confused by the difference between what is read and displayed visually.

mancristiana (changes requested):

templates/article_page.hbs
JakubKopys: I've followed the recommendation from the task (A11Y team?) - looks like they considered it good enough? But tbh I think that's a great suggestion, I will look into that

mancristiana (approved):

templates/community_post_page.hbs
```diff
@@ -99,7 +99,10 @@
 
                   {{#if (compare (calc post.author.badges.length "-" 4) ">" 0)}}
                     <div class="community-badge community-badge-achievements">
-                      {{#link "user_profile" id=post.author.id filter_by="badges" class="community-badge-achievements-rest"}}{{t 'plus_more' count=(calc post.author.badges.length "-" 4)}}{{/link}}
+                      test
```
mancristiana, line 102: Is this a leftover?

mancristiana (approved):

PR #300, opened 2022-02-10
feat: Create accessible Search field "clear" button

jgorfine-zendesk:
## Description

Due to [current technical limitations](https://adrianroselli.com/2019/07/ignore-typesearch.html#Delete), the "clear" button that comes with the `<input type="search">` is not keyboard- or assistive tech-accessible out of the box —at least, not right now. 

For this reason, we've created a new clear button (using a `<button>` element) that does more or less the same thing as the native clear button, but _is_ accessible to keyboard and assistive technology users.

This PR, opened as part of the Guide team's first Accessibility Smash of 2022, resolves [PRODA11Y-274](https://zendesk.atlassian.net/browse/PRODA11Y-274).

### Related PRs

* [#24503](https://github.com/zendesk/help_center/pull/24503), `help_center`: Adds button label for translation
* [#24603](https://github.com/zendesk/help_center/pull/24603), `help_center`: Creates key for button label translation
* [#24505](https://github.com/zendesk/help_center/pull/24505), `help_center`: Updates Autocomplete keyboard behavior to be more accessible

### Homepage, LTR

Screenshot of new clear button on the Guide homepage, in a left-to-right layout

### Homepage, RTL

Screenshot of new clear button on the Guide homepage, in a right-to-left layout

### Inner page, default

Screenshot of new clear button on an inner Guide page, default state

### Inner page, hover

Screenshot of new clear button on an inner Guide page, hover state

### Inner page, focus

Screenshot of new clear button on an inner Guide page, focus state

## Checklist


script.js
```diff
@@ -1,3 +1,83 @@
+// Vanilla JS debounce function, by Josh W. Comeau:
+// https://www.joshwcomeau.com/snippets/javascript/debounce/
+function debounce(callback, wait) {
+  let timeoutId = null;
```
luis-almeida, line 4: There's no usage of es6 syntax anywhere in the file so far so far.
Worth noticing that we should be ok with browser support as we no longer support IE11, but in case there are still users relying on that browser (or older versions of safari) this will throw invalid syntax errors.

script.js
```diff
@@ -1,3 +1,83 @@
...
+  if (searchInputDeleteKeys.includes(event.key)) {
+    clearSearchInput();
+  }
+}
+
+// Ensure there's always a label for the search clear button,
+// just in case the localized label isn't available.
+// This label is a big part of what makes the custom button accessible.
+function getClearSearchButtonLabel() {
+  return window.searchClearButtonLabelLocalized === undefined || window.searchClearButtonLabelLocalized === "" ? "Clear search" : window.searchClearButtonLabelLocalized;
```
luis-almeida, line 38: We should be able to trust the api so no need for the fallback string.

script.js
```diff
@@ -22,6 +102,12 @@ document.addEventListener('DOMContentLoaded', function() {
     return null;
   }
 
+  // Set up clear functionality for the search field
+  searchForm = document.querySelector("form[role='search']");
+  searchInput = searchForm.querySelector("input[type='search']");
```
luis-almeida, line 107: This will not work if there are multiple `{{search}}` helpers on the same page,  e.g. in the header and footer.
Could we adapt this code to run on every search form/input on the page?

script.js
jgorfine-zendesk: @luis-almeida I see what you're saying. I leaned into ES6 (here, and in a few other places) because we're very clear that we no longer support IE11. By saying
> in case there are still users relying on that browser (or older versions of safari) this will throw invalid syntax errors.
that seems to imply that we _do_ actually have a need to support older browsers, though. Is there a team stance on this?

script.js
jgorfine-zendesk: @luis-almeida I thought this might be worthwhile to have, in case something happens to the API, down the line. Does that make sense?

luis-almeida: I understand what you mean but this should never be the case.
If ourselves can't trust the API is stable than we can't ask our customers to trust us either.

script.js
luis-almeida: I agree with you. Sorry if my comment was confusing.
What I mean is that we no longer have to support older browsers so we should definitely be ok with this change. But since we have only been using es5 syntax so far, it means that there were no syntax errors only runtime errors.
This change will lead to syntax errors in IE11 which might remove support for all JS based features in the theme. We should be aware of that in case some customers complain.

script.js
luis-almeida: Can we remove this fallback logic?
This encourages distrust in the api and it's not something we should allow.

luis-almeida (commented):
Apart from the unlocalized fallback string it looks good and ready to merge.
Once that's resolved please consider squashing the commits for a cleaner changelog.

luis-almeida (approved):
I pushed the pending change as we're running out of time to get this PR closed. Hope that's ok.

jgorfine-zendesk:
Hi @luis-almeida — yes, that's great. Thanks!

luis-almeida (approved):
Also had to rebase. Should be ready to merge now.

PR #307, opened 2022-03-31
feat: trim the statuses using css, when they're longer than the allowed space

nwiggershansen:
## Description

Custom statuses are coming, with a limit of 45 characters, this doesn't really work well with the UI, this will help truncate statuses that would otherwise lose information due to being too large for the label. 

The reason for doing it in-theme, instead of truncating it on the backend, is that we don't want to limit our customers from the full information of a status, if they have a UI geared towards showing more characters. 

## Screenshots

**BEFORE**
Screenshot 2022-03-31 at 14 01 04
Screenshot 2022-03-31 at 14 00 52

**AFTER**

_Pixel 5:_ 
Screenshot 2022-04-12 at 08 06 48
_Ipad Air:_
Screenshot 2022-04-12 at 08 07 07
_Desktop:_ 
Screenshot 2022-04-12 at 08 06 30

_Pixel 5:_ 
Screenshot 2022-04-12 at 08 06 10
_Ipad Air:_
Screenshot 2022-04-12 at 08 05 58
_Desktop:_ 
Screenshot 2022-04-12 at 08 06 20
## Checklist

### DIFF
--- a/style.css
+++ b/style.css
@@ -2803,6 +2803,19 @@ ul {
   background-color: #000;
 }
 
+.status-label-request {
+  max-width: 100%;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+@media (max-width: 768px) {
+  .status-label-request {
+    max-width: 150px;
+  }
+}
+
 /***** Post *****/
 /*
 * The post grid is defined this way:
--- a/styles/_status-label.scss
+++ b/styles/_status-label.scss
@@ -73,4 +73,14 @@
   &-hold {
     background-color: #000;
   }
+
+  &-request {
+    @include mobile {
+      max-width: 150px;
+    }
+    max-width: 100%;
+    white-space: nowrap;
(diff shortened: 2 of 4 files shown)


anpa:
Could we use CSS instead to do this truncation?
https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

nwiggershansen:
> Could we use CSS instead to do this truncation? https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

As far as I can tell, it doesn't work, since we don't have a specified width on the field.

script.js
```diff
@@ -312,9 +312,9 @@ document.addEventListener('DOMContentLoaded', function() {
   ) {
     notificationElm.previousElementSibling.focus();
   }
-
+
```
anpa, line 315: nitty nit: extra space

anpa (approved):
Looks great! Thanks for following up ❤️

anpa (approved):

PR #325, opened 2022-08-31
feat: introduce content tags

vecerek:
## Description

This PR introduces content tags on the following three pages:

1. `New post page`: a new content tags field is added.
2. `Article page`: a new section listing the article content tags is added.
3. `Post page`: a new section listing the post content tags is added.

I couldn't use the following markup to render the content tag links:

```hbs
{{#link "search_result" content_tag_id=(id) target="_blank"}}
  {{name}}
{{/link}}
```

for the following two reasons:

1. There seems to be a bug in Curlybars because `content_tag_id=(id)` does not validate. I'll need to look into that.
2. `target="_blank"` is ignored by the link helper, we would need to add the attribute to the allowlist.

## Screenshots

New post page:

Screenshot 2022-08-31 at 09 19 05

Content tags section:

Screenshot 2022-08-31 at 09 21 40

## Checklist

### DIFF
--- a/style.css
+++ b/style.css
@@ -5082,3 +5082,39 @@ html[dir="rtl"] .notification-left-aligned {
 .dropdown-chevron-icon {
   vertical-align: middle;
 }
+
+.content-tags > p {
+  color: #68737D;
+  margin-top: 32px;
+  margin-bottom: 4px;
+}
+
+.content-tags-add-hint {
+  color: #68737D;
+  font-size: 14px;
+}
+
+.content-tag-list {
+  display: flex;
+  flex-wrap: wrap;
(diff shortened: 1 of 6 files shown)


luis-almeida:
Code looks good but I wanted to address your points before approving:

> There seems to be a bug in Curlybars because content_tag_id=(id) does not validate. I'll need to look into that.

You don't need to wrap the id in parenthesis. The following should work:
`{{#link "search_result" content_tag_id=id}}`

> target="_blank" is ignored by the link helper, we would need to add the attribute to the allowlist.

Good point! Lets add support for it :)

luis-almeida (approved):

PR #333, opened 2022-11-21
[GG-2903] Improve contrast in theme elements

jbdietrich:
## Description

According to A11y best practices input elements should have a contrast ratio of at least 3:1 with surrounding elements.

After this change, the border color of many input elements changes from `#dddddd` to `#87929D`. The latter has a contrast ratio of > 3 with the theme's default background color: `#ffffff`.

We still have a need for some elements to use the lower contrast border color -- `#ddd` -- so this PR splits the `$border-color` SASS variable into two separate values.

| Page name    |  Page element | Fixed?           | Screenshot   |
| ------------- | ------------- | ------------- | ------------- |
| Search results page  | Search bar  | ✅                    |   Screenshot 2022-11-24 at 11 17 05 |
| My requests page  | Search bar  |              ✅           |        Screenshot 2022-11-24 at 11 21 45   |
| My requests page  | Status dropdown  |              ✅           |        Screenshot 2022-11-24 at 11 21 45   |
| Article page | Search bar | ✅ | Screenshot 2022-11-28 at 12 53 38 |
| Article page | Comment form | 🟨 Partial fix -- editor must be updated in HC | Screenshot 2022-11-28 at 15 35 01 |
| Submit a request page | Form fields | 🟨 Partial fix -- some elements need to be updated in HC |  Screenshot 2022-11-28 at 12 43 59 | 
| Topic show page | Search bar | ✅ | Screenshot 2022-11-28 at 12 55 19 |
| New post page | Form fields | 🟨 Partial fix -- some elements need to be updated in HC | Screenshot 2022-11-28 at 13 00 43 |
| Edit post modal | Form fields | ❌ must be updated in HC | N/A |
| Create a ticket modal | Form fields | ❌ must be updated in HC | N/A |
| Award badges modal | Form fields | ❌ must be updated in HC | N/A |
| Post page page | Comment form | 🟨 Partial fix -- editor must be updated in HC | Screenshot 2022-11-28 at 15 33 54 |

## Checklist

- [x] :green_book: all commit messages follow the [conventional commits](https://conventionalcommits.org/) standard
- [x] :nail_care: SASS files are compiled
- [x] :arrow_left: changes are compatible with RTL direction
- [x] :wheelchair: changes are accessible
- [x] :memo: changes are tested in Chrome, Firefox, Safari and Edge
- [x] :iphone: changes are responsive and tested in mobile
- [ ] :+1: PR is approved by @zendesk/vikings

PR #340, opened 2022-12-12
[GG-2903] Increase contrast ratio for more theme elements

jbdietrich:
Initially I added these changes to Help Center directly. It turns out that doing so would amount to a breaking change for customers who assume that these elements will remain the same color in their custom themes. To err on the side of caution, I am adding them to the Copenhagen theme instead.

## Description

Follow up to https://github.com/zendesk/copenhagen_theme/pull/333

Initially I added these changes to Help Center directly. It turns out that doing so would amount to a breaking change for customers who assume that these elements will remain the same color in their custom themes. To err on the side of caution, I am adding them to the Copenhagen theme instead.

## Screenshots

Screenshot 2022-12-12 at 16 13 57

## Checklist

### DIFF
--- a/style.css
+++ b/style.css
@@ -590,6 +590,7 @@ ul {
 
 .form-field .nesty-input {
   border-radius: 4px;
+  border: 1px solid #87929D;
   height: 40px;
   line-height: 40px;
   outline: none;
@@ -599,6 +600,10 @@ ul {
 .form-field .nesty-input:focus {
   border: 1px solid $brand_color;
   text-decoration: none;
+}
+
+.form-field .hc-multiselect-toggle {
+  border: 1px solid #87929D;
 }
 
 .form-field .hc-multiselect-toggle:focus {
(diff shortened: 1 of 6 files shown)


luis-almeida (commented):
Looks good and this is indeed a safer approach!

For consistency, and since we already have some rules targeting a [`nesty-input`](https://github.com/zendesk/copenhagen_theme/blob/master/styles/_forms.scss#L38) and [`hc-multiselect`](https://github.com/zendesk/copenhagen_theme/blob/master/styles/_forms.scss#L51) within `forms`, we could place these new rules there as well as they all seem to apply to the request form.

Otherwise, I think it's valid as well to create the new partials but then it'd maybe be better to move the existing rules as well so they're all in one place.

jbdietrich:
> For consistency, and since we already have some rules targeting a [nesty-input](https://github.com/zendesk/copenhagen_theme/blob/master/styles/_forms.scss#L38) and [hc-multiselect](https://github.com/zendesk/copenhagen_theme/blob/master/styles/_forms.scss#L51) within forms, we could place these new rules there as well as they all seem to apply to the request form.

Good point! I think it's simpler with fewer partials. Moved them to the `_forms.scss` file.

luis-almeida (approved):

PR #341, opened 2022-12-12
Screen reader improvements

ralfpatric:
## Description

This PR introduces some improvements that will enhance the usage of screen readers. Prior to this change the meta-data elements were interpreted as list items.

## Checklist

### DIFF
--- a/style.css
+++ b/style.css
@@ -1553,6 +1553,10 @@ ul {
   padding: 20px 0;
 }
 
+.recent-activity-item h3 {
+  margin: 0;
+}
+
 .recent-activity-item-parent {
   font-size: 16px;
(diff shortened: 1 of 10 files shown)


Anifacted (changes requested):
Looks like the SCSS files are not compiled.

Anifacted (approved):

anpa (commented):
Curious, why is it better in terms of a11y to interpret meta-data elements as `span` instead of a list of items?

(thanks for giving some a11y love to the theme ❤️ )

ralfpatric:
The reason here is that some screen readers will announce that the user is traversing a list, where in reality it's not a list but just information about the post.

PR #356, opened 2023-03-02
Luis.rollup

luis-almeida:
## Checklist

- [ ] :wheelchair: changes are accessible and [tested locally](./../README.md#accessibility-testing)


Anifacted (approved):
Awesome!

anpa (commented):
Bravo Luís! This is amazing! 👏 👏 👏 

One thing: the lighthouse github action is failing as it's not able to upload the `style.css` and `script.js` files. To fix it, we need to add [a step before to generate the files](https://github.com/zendesk/copenhagen_theme/blob/master/.github/workflows/lighthouse.yml#L23).

I am also thinking that, to import a theme directly from github to theming center, we need to have both files generated. I guess this would break that flow. Is it worth removing them?

luis-almeida:
@anpa thanks for the review!

Since would start now compiling both `JS` and `CSS` it felt suboptimal to also include the compiled files in the repo. It can also create confusion as developers might edit the "compiled files" instead of the source code.

The idea was to have them built and checked into the repo at release time. Similar to what happens when a library is published to npm, e.g. source code gets compiled and added to `dist`. Themes do not have a `dist` folder so the output target is `script.js` and `style.css`.

I [added a task](https://github.com/zendesk/copenhagen_theme/pull/356/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R21-R22) to build these during `release`, so upon merging they should be checked into the repo just like before. This needs to happen for the `lighthouse` workflow as well so I could try to fix that.

Let me know if this doesn't make sense at all and I can also add just add them back.

README.md
```diff
@@ -81,32 +85,30 @@ You can add up to 10 optional templates for:
...
-- Install Ruby, we use `sassc` gem to compile our `.scss` files. You can see how to install Ruby [here](https://www.ruby-lang.org/en/documentation/installation/).
-- Install `sassc` gem. You can run:
 ```console
-  gem install sassc:1.12.1
+$ yarn install --offline
+$ yarn start
 ```
 
-Now you can compile your SASS files running:
+This will compile all the source code in `src` and `styles` and watch for changes.
```
anpa, line 99: nit: should we also add a note to use `yarn build` to compile without watching for changes?

README.md
```diff
@@ -81,32 +85,30 @@ You can add up to 10 optional templates for:
...
+
+Notes:
+
+- We intentionally do not use babel so we can get a clean bundle output. Make sure to only use widely supported ecmascript features (ES2015).
```
anpa, line 103: ❤️

README.md
```diff
@@ -81,32 +85,30 @@ You can add up to 10 optional templates for:
...
-Note: Zendesk App Tools [theme preview](#publishing-your-theme) currently does not support live SASS compilation.
+To get started:
 
-## Requirements
-
-- Install Ruby, we use `sassc` gem to compile our `.scss` files. You can see how to install Ruby [here](https://www.ruby-lang.org/en/documentation/installation/).
-- Install `sassc` gem. You can run:
 ```console
-  gem install sassc:1.12.1
+$ yarn install --offline
```
anpa, line 95: This PR does not add the yarn packages to the repo. Should we remove the `--offline` option here?

```suggestion
$ yarn install
```

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "1.0.0",
   "repository": "git@github.com:zendesk/copenhagen_theme.git",
   "scripts": {
+    "start": "rollup -c -w",
+    "build": "rollup -c",
+    "eslint": "eslint src",
```
anpa, line 8: nitty nit (feel free to ignore 😄)

```suggestion
    "lint": "eslint src",
```

styles/_notifications.scss
```diff
@@ -112,8 +112,8 @@
     display: none;
   }
 
-  &.notification-error::before {
-    @extend .notification-icon::before;
+  &.notification-error, ::before {
```
anpa, line 115: This seems to be causing the notification error to render weirdly:

Screenshot 2023-03-13 at 11 43 26

This seems to fix it:

```suggestion
  &.notification-error::before {
```

anpa:
@luis-almeida thanks for explaining! 🙏 

> You'll see that [I switched](https://github.com/zendesk/copenhagen_theme/pull/356/commits/a3c42b6e073de9187d1fbf6a51faba87c32727af) the strategy when creating the zip file and am now only including the necessary files and folders.

Good catch!

> At this point I'm only worried that [running yarn build during release](https://github.com/zendesk/copenhagen_theme/pull/356/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R21-R22) won't be enough to get the files checked into the repo. Solution would be to just do it in a separate task.

I think we are just missing telling `semantic-release` to check `style.css` and `script.js` into the repo. It should be enough to add them [here](https://github.com/zendesk/copenhagen_theme/blob/master/.releaserc#L13) together with the other changes that happen during the release. That should add the release commit so that the `master` branch will have a `style.css` and a `script.js` files.

I think what you've explained makes sense and I agree it's a cleaner approach. My only concern is that we will only be able to import the Copenhagen theme from the master branch and not from a development branch. But if we really need it, we can manually commit the `style.css` and `script.js` to the development branch, so I think it's a fair trade-off.

I left some comments, but it's looking awesome! Thanks again for this effort! Looking forward to get this one merged! 🥳

.github/workflows/release.yml
```diff
@@ -18,6 +18,8 @@ jobs:
           node-version-file: '.nvmrc'
       - name: Install dependencies
         run: yarn install --frozen-lockfile
+      - name: Build
+        run: yarn build
```
anpa, line 22: I think we are just missing telling `semantic-release` to check `style.css` and `script.js` into the repo. It should be enough to add them [here](https://github.com/zendesk/copenhagen_theme/blob/master/.releaserc#L13).

src/index.js
```diff
@@ -0,0 +1,7 @@
+import "../styles/index.scss";
```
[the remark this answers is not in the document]
luis-almeida: Unfortunately `rollup-plugin-sass` does not support using a `scss` file as the `input` (similar to entry `entry` in webpack) so we need to trigger the compilation via an `import` in the `js` file. 

The plugin I'm adding - `zass` - is basically a wrapper around what `rollup-plugin-sass` provides. It's possible to drop that dependency and implement everything from scratch so we could remove this line but I thought it was an ok tradeoff to safe that time.

luis-almeida:
> My only concern is that we will only be able to import the Copenhagen theme from the master branch and not from a development branch.

That's a very valid point and I initially misinterpreted what you meant.

> But if we really need it, we can manually commit the `style.css` and `script.js` to the development branch, so I think it's a fair trade-off.

In this repo 's case, importing from a branch is mostly done just for testing purposes.
The solution as you mentioned is to `git add -f script.js style.css` and then commit and push the wip changes.
Maybe this could be something to mention in the documentation (?)

.github/workflows/release.yml
luis-almeida: Even easier! :)
Thanks for pointing it out as I'm not too familiar with `semantic-release`.

styles/_notifications.scss
luis-almeida: This is a great catch!
I did test this one (though only in the request form) and it's a bit scary that I didn't catch the same bug.
Thanks so much for also looking into it 🙏🏼 
I'll address the issue and try to understand why I missed it in the first place.

Anifacted:
This is really awesome, @luis-almeida and thanks @anpa for thorougly reviewing 💯

luis-almeida:
@anpa I've pushed a commit addressing your comments.
Still struggling testing the css change to `.notifications` but if you could help me verify those (for extra piece-of-mind) then this could be ready to merge once you have time to give it a final review.

anpa:
@luis-almeida Thanks for addressing the comments! ❤️ 

I've tested the main functionality (including the notifications) and did a comparison of `style.css` and `script.js` with the previous versions. Everything seems to be correct. I did a small fix for the extension of the compound selectors as that seems to [not be allowed in dart-sass](https://sass-lang.com/documentation/at-rules/extend#disallowed-selectors).

So I think this is ready to be merged!!! 🥳 🥳 🥳 Thanks once again for bringing these changes! 🏆

anpa (approved):
💪 👏

timothychoi-rg:
Hi @luis-almeida , this seems to have broken support for dots in asset filenames (e.g. a.yaml.json - this is to work around Zendesk limitations). We are no longer able to use dots whereas this was fully supported.

Also, on Macs, Rollup will erroneously report `.DS_Store` with having illegal characters in its name.

anpa:
Thanks for your feedback @timothychoi-rg and our apology for the issues you've encountered. What you've described is actually a bug as the support for asset filenames should be the same before and after these changes.

We've added support for dots in [this PR](https://github.com/zendesk/copenhagen_theme/pull/364). Can you please let us know if that fixes your issue? Thanks a lot!

PR #362, opened 2023-03-22
refactor: CSS and JS changes for new WYSIWYG

Anifacted:
## Description

The new editor will support image alignment features in community posts. This PR adds extra CSS rules to support those features.

It also removes the disabling of the submit button for submitting comments on support-requests, since we cannot rely on the value of the <textarea> being populated when using the new editor framework.
Generally, disabling a submit button without further explanation is discouraged from a UX perspective, as well: https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/

## Screenshots

image

## Checklist

- [ ] :wheelchair: changes are accessible and [tested locally](./../README.md#accessibility-testing)


luis-almeida:
@Anifacted you'll need to rebase this PR on master.
`script.js` has been split into modules and will be rebuilt during release. Your changes will be undone if they're not within `src/`. Same thing for the `style.css` file.

Let us know if we should improve [documentation](https://github.com/zendesk/copenhagen_theme#stylesheet-and-javascript) around this:

> - Both `style.css` and `script.js` are ignored as they'll be regenerated during release. Do not edit these directly.
>    - If you do want to test your changes by importing the branch in Theming Center, you'll need to manually add and commit these files, e.g. `git add -f script.js style.css`

Anifacted:
@luis-almeida oh, but I did rebase and I did edit the source files, but I suppose I somehow managed to check in the compiled files.

luis-almeida:
I think that might've happened because you had this branch prior to that split where you could still easily add and commit on both bundles. They're now part of `.gitignore` so this shouldn't happen going forward.

Anifacted:
> I think that might've happened because you had this branch prior to that split where you could still easily add and commit on both bundles. They're now part of .gitignore so this shouldn't happen going forward.

Gotcha, thank you!

PR #381, opened 2023-06-23
Deprecating ruby and using zcli to import the theme for the lighthouse run

luis-almeida:
## Description

- Replacing `theme_upload.rb` with a `theme-upload.js`
- `theme-upload.js` uses `zcli` commands to import new them, publish it and cleanup old themes
- Deprecating usage of `zat` for the `lighthouse` run
  - renaming `.zat` to `.a11yrc.json`
- Updating documentation
- Removing `.ruby-version` as ruby is now obsolete in the repo

## Checklist


anpa (approved):
Yessssss!!! 😍

anpa (commented):
I believe this was the last ruby dependency actually. If that's the case, should we also remove the [ruby-version file](https://github.com/zendesk/copenhagen_theme/blob/master/.ruby-version)?

luis-almeida:
Thanks for the quick review @anpa!
Might need to bug you again as I'm still missing porting the publish and cleanup part:
https://github.com/zendesk/copenhagen_theme/pull/381/files#diff-30d2352910bc54b4f1c7c12573b846c6363b5d69686674092f426fe593ca50fbL69-L81

This draft was basically to see if `themes:import` would "hold up" in a github action.
My plan was to just be able to use `zcli` and bypass the need for any custom node script. Unfortunately there's not a `themes:delete` command yet for the cleanup part. Might not even be necessary though as we might be able to just update the existing theme. I'll spend just a bit more time on this and should be able to wrap it up soon for a re-review.

> I believe this was the last ruby dependency actually. If that's the case, should we also remove the [ruby-version file](https://github.com/zendesk/copenhagen_theme/blob/master/.ruby-version)?

I think so. But maybe best to do that as part of removing all references to `zat` which has become obsolete. If it's ok, I'll do that in a next step.

sergioazevedo (approved):

bin/theme-upload.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-env node */
+const brandId = process.env.BRAND_ID;
+const { execSync } = require("child_process");
+
+function zcli(command) {
+  const data = execSync(`yarn --silent zcli ${command} --json`);
```
anpa, line 6: Do you know if `--silent` will silence errors in case the command fails?

anpa (approved):
Super nice! 👏

bin/theme-upload.js
luis-almeida: We will still get the error output if the script fails. This is necessary to suppress the log during the run it doesn't interfere with the `--json` flag.

PR #382, opened 2023-07-04
Ticket Form Selector

Fredx87:
## Description

This PR adds a ticket form selector developed in JavaScript using React and Garden components.

It also introduces a new folder `src/modules` which contains JS/JSX files that will be bundled using ES module format. It is possible to import these modules from the Curlybars template using a `<script type="module">` tag.
The bundles created with this setup are emitted as assets, and since they are renamed when deployed, an import map is needed to retrieve the correct file in the theme.

React and Garden components are imported from NPM, and the Rollup setup has been extended to support the compilation of the modules, condensing all the dependencies in a separate bundle, and customizing the generated code to make import maps work. This setup will be optimized in the future.

## Screenshots
Screenshot 2023-07-17 at 16 45 52

## Checklist


templates/new_request_page.hbs
```diff
@@ -18,7 +18,16 @@
...
+  <div id="main-content" class="form" data-props='
+  {
+    "ticketForms": {
+     "options": [
+      {{ #each ticket_forms }}
+       { "id": {{id}}, "url": {{json_stringify url}}, "name": {{json_stringify display_name}}  } {{#unless @last}},{{/unless}}
+      {{ /each }}
+     ]
+    }
+  }
```
anpa, line 30: Another way of exposing this data is:

```hbs
<script>
  var forms = []
  {{#each ticket_forms}}
    forms.push({
      id: {{id}},
      url: "{{url}}",
      displayName: "{{display_name}}"
    })
  {{/each}}
  
  console.log(forms);
</script>


```

sergioazevedo (approved):

rollup.config.mjs
```diff
@@ -1,13 +1,84 @@
...
+    plugins: [zass()],
+    watch: {
+      clearScreen: false,
+    },
   },
-  plugins: [zass()],
-  watch: {
-    clearScreen: false,
+  {
+    input: ["src/modules/NewRequestForm.jsx", "src/modules/shared.jsx"],
```
luis-almeida, line 21: Would be nice to be consistent in terms of naming.
It doesn't seem like we're following the pattern of having the file name aligned with the default exported function name so I'm wondering if it makes sense to have some bundles in camel case and some other not.

luis-almeida (commented):
Looks really good! Just left a couple of comments/questions.

There's things we can address in follow up PRs, e.g. polish import maps or compilation of `assets` after merging.

rollup.config.mjs
Fredx87: It makes sense to have consistent file naming. I think initially I just exported the component so I called the file with the component name, but now we export a function, so probably it makes sense to change the file name as well

PR #393, opened 2023-07-20
chore: added plugin for automatic importmap generation

Fredx87:
## Description

This PR improves the management of the import map, which is now generated automatically from the chunks file emitted by rollup, instead of manually doing it.

I added a plugin that runs after Rollup generates the bundle and:
1. replaces relative imports with bare imports in the emitted files
2. creates an import map and inject it in the `document_head.hbs` template, replacing the existing one.

Note: the first step is done with a basic string replace, e.g. `'./a.js'` is replaced with `'a'`, and this should cover both static imports (`import something from './a.js'`) and dynamic ones (`import('./a.js')`). Maybe there is a better way to do it, but I didn't manage to find one.

## Checklist

### DIFF
--- a/generate-import-map.mjs
+++ b/generate-import-map.mjs
@@ -23,12 +23,14 @@ export function generateImportMap() {
   return {
     name: "rollup-plugin-generate-import-map",
     writeBundle({ dir }, bundle) {
-      const outputChunksNames = Object.keys(bundle);
+      const parsedOutputChunks = Object.keys(bundle).map((chunk) =>
+        path.parse(chunk)
+      );
       const importMap = { imports: {} };
 
-      for (const chunkName of outputChunksNames) {
-        relativeToBareImports(chunkName, outputChunksNames, dir);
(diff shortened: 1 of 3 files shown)


generate-import-map.mjs
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  const existingImportMap = content.substring(
+    importMapStart,
+    importMapEnd + `</script>`.length
+  );
+
+  const newImportMap = `<script type="importmap">
+${JSON.stringify(importMap, null, 2)}
+</script>`;
```
luis-almeida, line 81: Nit: to keep the code alignment `[].join('\n')` is also an option.

luis-almeida (commented):
Just added a few minor comments but looks pretty good!

Before approving I'd just suggest to add `*.mjs` to be linted by `eslint`.
Also not sure why `prettier` is no longer working (maybe just for me?) but that can of course be addressed in a following PR.

luis-almeida (approved):

PR #394, opened 2023-07-21
Hide generated files instead of ignoring

luis-almeida:
## Description

Compiled assets are regenerated upon every release. However, for side-running branches, or simply to be able to import a themes containing the latest changes, it's easier to not ignore compiled files.
This PR moves them to `.gitattributes` and set them as `generated` so they're always hidden by default in PRs.

## Checklist


anpa (approved):
Nice find! Let's try it out!

PR #397, opened 2023-07-27
Rendering DropDown and TextArea ticket fields

luis-almeida:
## Description

- Adding a basic `DropDown` component
  - We might want to streamline and just use the same as for the ticket form field selector but I thought it was worth it to start with this one for ticket fields.
- Adding a basic `TextArea` component
- Rendering `anonymous_requester_email`, `description` and `textarea` type fields

Handling of state and submission in a follow up PR.

## Checklist

### DIFF
--- a/src/modules/new-request-form/fields/DropDown.tsx
+++ b/src/modules/new-request-form/fields/DropDown.tsx
@@ -1,0 +1,35 @@
+import {
+  Field as GardenField,
+  Label,
+  Hint,
+  Combobox,
+  Option,
+  Message,
+} from "@zendeskgarden/react-dropdowns.next";
+import type { Field } from "../data-types";
+
+interface DropDownProps {
+  field: Field;
+}
+
+export function DropDown({ field }: DropDownProps): JSX.Element {
+  const { label, options, error, value, name, required, description } = field;
+  return (
+    <GardenField>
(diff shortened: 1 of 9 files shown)


src/modules/new-request-form/fields/DropDown.tsx
```diff
@@ -0,0 +1,35 @@
...
+  field: Field;
+}
+
+export function DropDown({ field }: DropDownProps): JSX.Element {
+  const { label, options, error, value, name, required, description } = field;
+  return (
+    <GardenField>
+      <Label>{label}</Label>
+      {description && <Hint>{description}</Hint>}
+      <Combobox inputProps={{ name, required }} isEditable={false}>
```
Fredx87, line 21: I think we should set the `validation` prop to `error` if there is an error, like the other fields

templates/new_request_page.hbs
```diff
@@ -53,9 +53,23 @@
   var ticketForms = [];
   var fields = [];
 
+  console.log(fields)
```
Fredx87, line 56: Should we remove this console.log?

luis-almeida:
@Fredx87 thanks for the review! I addressed your comments so ready for another go 🙏🏼

PR #402, opened 2023-08-01
feat: added form submission and global form errors handling

Fredx87:
## Description

This PR adds the possibility to submit the New Request Form.
- Added a Submit button rendered with Garden
- Added logic to fetch the CSRF token before submitting the form.
   - The logic should be more or less the same on the current implementation: the token is fetched just before submitting the form and appended as a hidden input. I didn't find a more "React-ish" version to do it, but I think it is fine
   - We are temporarily using the internal endpoint
- Changed `TextInput` and `TextArea` components, using `defaultValue` instead of `value`. In this way, they are considered uncontrolled components by React. Without this change, it was impossible to change the value once a value was set by the template data
- Added an `Alert` component for displaying any eventual global error on top of the form
- Minor UI improvements

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/13420283/c19aba69-7d0f-4ae1-9f0c-ee645a4211de

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,10 @@
   },
   "dependencies": {
     "@zendesk/zcli": "^1.0.0-beta.35",
+    "@zendeskgarden/react-buttons": "^8.69.1",
     "@zendeskgarden/react-dropdowns.next": "^8.69.1",
     "@zendeskgarden/react-forms": "^8.69.1",
+    "@zendeskgarden/react-notifications": "^8.69.1",
     "@zendeskgarden/react-theming": "^8.69.1",
     "node-fetch": "2.6.9",
     "react": "^17.0.2",
@@ -34,6 +36,7 @@
     "@semantic-release/git": "10.0.1",
     "@types/react": "^17.0.62",
     "@types/react-dom": "^17.0.20",
+    "@types/styled-components": "^5.1.26",
     "@typescript-eslint/eslint-plugin": "^6.1.0",
     "@typescript-eslint/parser": "^6.1.0",
     "concurrently": "8.0.1",
@@ -56,6 +59,9 @@
     "typescript": "^5.1.6",
     "url-pattern": "1.0.3",
     "wait-on": "7.0.1"
+  },
+  "resolutions": {
+    "@types/react": "^17.0.62"
   },
   "commitlint": {
     "extends": [
--- a/src/modules/new-request-form/NewRequestForm.tsx
+++ b/src/modules/new-request-form/NewRequestForm.tsx
@@ -3,10 +3,10 @@ import { TextInput } from "./fields/TextInput";
 import { TextArea } from "./fields/TextArea";
 import { DropDown } from "./fields/DropDown";
 import { TicketFormField } from "./ticket-form-field/TicketFormField";
-import type { FormEventHandler } from "react";
 import { Button } from "@zendeskgarden/react-buttons";
 import styled from "styled-components";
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -57,6 +60,9 @@
     "url-pattern": "1.0.3",
     "wait-on": "7.0.1"
   },
+  "resolutions": {
+    "@types/react": "^17.0.62"
+  },
```
Fredx87, line 65: Note: this is needed to fix [this issue](https://github.com/styled-components/styled-components/issues/3721) of `@types/styled-components`

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -18,12 +32,43 @@ export function NewRequestForm({
...
+    const { current_session } = await response.json();
+    return current_session.csrf_token as string;
+  };
+
+  const handleSubmit: FormEventHandler<HTMLFormElement> = (e) => {
+    e.preventDefault();
+
+    const form = e.target as HTMLFormElement;
+
+    fetchCsrfToken().then((token) => {
```
luis-almeida, line 52: Nit: consistent `async`/`await` for promises.

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -18,12 +32,43 @@ export function NewRequestForm({
...
+      const hiddenInput = document.createElement("input");
+      hiddenInput.type = "hidden";
+      hiddenInput.name = "authenticity_token";
+      hiddenInput.value = token;
+      form.appendChild(hiddenInput);
+      form.submit();
+      form.removeChild(hiddenInput);
```
luis-almeida, line 59: Why do we need to remove it?
It shouldn't persist when validation fails, right?

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -3,12 +3,26 @@ import { TextInput } from "./fields/TextInput";
...
   requestForm: RequestForm;
 }
 
+const StyledForm = styled.form`
+  display: flex;
+  flex-direction: column;
+  gap: ${(props) => props.theme.space.md};
+`;
+
+const Footer = styled.div`
```
luis-almeida, line 22: Nit: Should we keep consistency and call this `StyledFooter`?
Or maybe simpler to to just have `Form` and `Footer`.

luis-almeida (approved):

src/modules/new-request-form/NewRequestForm.tsx
Fredx87: In theory, we don't need it, because if there are errors when the form is submitted, the page is reloaded.

I am just not sure if it is always the case, or if it is possible that the `submit` fails and we end up adding multiple times the same hidden input. So I thought it may be helpful to have the field removed to be sure that we don't end up in an ambiguous state.

Actually, I found out that with this implementation the user can click multiple times on the `Submit` button and we end up calling the CSRF token API multiple times. Probably we need to disable the submit button while the token is being fetched or make sure to call the API only once as we are doing in HC

Fredx87:
I ended up creating a `useSubmitHandler` for managing the fetching and injection of the CSRF token since we'll need it for sure on other forms as well in the future, and I added a logic to prevent multiple submissions

luis-almeida (approved):
🚀 !

PR #404, opened 2023-08-03
feat: organization field

luis-almeida:
## Description

- [x] Fixing for submission when ticket forms are not enabled
- [x] Removing obsolete html markup
- [x] Rendering the organization field 

## Checklist

### DIFF
--- a/src/modules/new-request-form/NewRequestForm.tsx
+++ b/src/modules/new-request-form/NewRequestForm.tsx
@@ -47,11 +47,13 @@ export function NewRequestForm({
       onSubmit={handleSubmit}
     >
       {errors && <Alert type="error">{errors}</Alert>}
-      <TicketFormField
-        label={ticket_forms_instructions}
-        ticketFormField={ticket_form_field}
-        ticketForms={ticketForms}
-      />
+      {ticketForms.length > 0 && (
+        <TicketFormField
+          label={ticket_forms_instructions}
+          ticketFormField={ticket_form_field}
+          ticketForms={ticketForms}
+        />
+      )}
       {fields.map((field) => {
         switch (field.type) {
           case "anonymous_requester_email":
@@ -67,7 +69,7 @@ export function NewRequestForm({
         }
       })}
       <Footer>
-        {ticket_form_field.value && (
+        {(ticketForms.length === 0 || ticket_form_field.value) && (
           <Button isPrimary type="submit">
             Submit
           </Button>
--- a/src/modules/new-request-form/NewRequestForm.tsx
+++ b/src/modules/new-request-form/NewRequestForm.tsx
@@ -63,6 +63,7 @@ export function NewRequestForm({
           case "textarea":
             return <TextArea field={field} />;
           case "priority":
+          case "organization_id":
             return <DropDown field={field} />;
           default:
             return <></>;
--- a/src/modules/new-request-form/fields/DropDown.tsx
+++ b/src/modules/new-request-form/fields/DropDown.tsx
@@ -22,12 +22,17 @@ export function DropDown({ field }: DropDownProps): JSX.Element {
         inputProps={{ name, required }}
         isEditable={false}
         validation={error ? "error" : undefined}
+        renderValue={({ selection }) =>
+          selection && "value" in selection
+            ? options.find((option) => option.value === selection.value)?.name
+            : "-"
+        }
       >
         {options.map((option) => (
           <Option
             key={option.value}
             value={option.value}
-            isSelected={option.value === value}
+            isSelected={option.value?.toString() === value?.toString()}
           >
             {option.name}
           </Option>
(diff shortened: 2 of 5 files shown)


src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -47,11 +47,13 @@ export function NewRequestForm({
       onSubmit={handleSubmit}
     >
       {errors && <Alert type="error">{errors}</Alert>}
-      <TicketFormField
-        label={ticket_forms_instructions}
-        ticketFormField={ticket_form_field}
-        ticketForms={ticketForms}
-      />
+      {ticketForms.length > 0 && (
```
Fredx87, line 50: This is already handled in the `TicketFormField` component: https://github.com/zendesk/copenhagen_theme/blob/b80ed7e9b4490d85f7ac1438268a29e4e8f7ebd4/src/modules/new-request-form/ticket-form-field/TicketFormField.tsx#L44
But maybe it makes sense to move it here for clarity (but we need to remove the other check)

Side note: is it possible to have 0 ticket forms? I added the check `ticketForms.length > 1` because if I don't configure the ticket forms I always have at least one (the default one)

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -61,13 +63,14 @@ export function NewRequestForm({
...
           case "priority":
+          case "organization_id":
             return <DropDown field={field} />;
           default:
             return <></>;
         }
       })}
       <Footer>
-        {ticket_form_field.value && (
+        {(ticketForms.length === 0 || ticket_form_field.value) && (
```
Fredx87, line 73: Same as above: is it possible to have 0 ticket forms?

src/modules/new-request-form/NewRequestForm.tsx
luis-almeida: > is it possible to have 0 ticket forms?

Yup! When the plan does not have the feature, ticket forms is empty.
So basically we have this state:

- Customer **does not have**  the `ticket_forms` feature:
  - `ticketForms` is empty
  - We cannot post `request[ticket_form_id]` or validation will fail
- Customer **has**  the `ticket_forms` feature:
  - There is at least one `ticketForm`
  - We must post `request[ticket_form_id]` or validation will fail
    - If there's only one ticket form we don't want to show the dropdown
    - If there's several ticket forms we must show the dropdown

> This is already handled in the TicketFormField component:

It is for the dropdown but not for the hidden input field. So when there are no ticket forms, we're posting `request[ticket_form_id]` with the value of `null` and that fails validation:
https://github.com/zendesk/copenhagen_theme/blob/b80ed7e9b4490d85f7ac1438268a29e4e8f7ebd4/src/modules/new-request-form/ticket-form-field/TicketFormField.tsx#L42

> But maybe it makes sense to move it here for clarity (but we need to remove the other check)

I think we still need this check because we don't want to show the dropdown with only one option.

src/modules/new-request-form/fields/DropDown.tsx
```diff
@@ -22,12 +22,17 @@ export function DropDown({ field }: DropDownProps): JSX.Element {
...
+            ? options.find((option) => option.value === selection.value)?.name
+            : "-"
+        }
       >
         {options.map((option) => (
           <Option
             key={option.value}
             value={option.value}
-            isSelected={option.value === value}
+            isSelected={option.value.toString() === value.toString()}
```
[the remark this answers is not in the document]
luis-almeida: > I was testing the CC field, and in that case, we need to pass different values when we submit the form (e.g. request[collaborators][]: test@example.com, request[collaborators][]: test2@example.com) but then we receive back a string value: "test@example.com,  test2@example.com".

I understand what you mean and that's an interesting problem as well!
But is the type still valid? Meaning `Field["value"]` is always a `string`, before and after submission?

Because in the case of a `field` with `options` that is not the case:
- Before submitting the form we have `type Field { value: number }`
- When validation fails, we have `type Field { value: string }`

So if we don't fix this on the API side, we should update the type to:
```ts
type Field { 
  value: number | string
}
```

src/modules/new-request-form/NewRequestForm.tsx
Fredx87: Ah! I didn't know about it! It makes sense then.
 
Maybe I would add a comment stating that if the user has no ticket forms we shouldn't populate the `request[ticket_form_id]`, to make it clear why we have 2 different `if`s, one checking for 0 and another checking for 1

Fredx87 (approved):

src/modules/new-request-form/fields/DropDown.tsx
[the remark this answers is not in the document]
Fredx87: So, now we are getting a `number` at the beginning and then a `string` when validation fails. That's not ideal and as you said it would be better to have the same type.

I was thinking of returning always a string for consistency and simplicity. In the future, we'll need to handle also the number custom field, and if I remember correctly we'll need to use string as well o we will risk losing precision for decimal values

PR #405, opened 2023-08-07
feat: add follow-up string to request form header

annterina:
## Description

Adding a follow-up string to the header of the new request page.

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/25595674/001bbefc-1f37-4eb4-8d7f-237c03207e62

## Checklist

### DIFF
--- a/templates/new_request_page.hbs
+++ b/templates/new_request_page.hbs
@@ -12,6 +12,11 @@
 
   <h1>
     {{t 'submit_a_request'}}
+    {{#if parent_id}}
+      <span class="follow-up-hint">
+        Follow-up to <a href="{{page_path 'request' id=parent_id}}">request #{{parent_id}}</a>
+      </span>
+    {{/if}}
   </h1>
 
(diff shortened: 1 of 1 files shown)


luis-almeida (approved):

templates/new_request_page.hbs
```diff
@@ -14,6 +14,12 @@
     {{t 'submit_a_request'}}
   </h1>
 
+  {{#if parent_id}}
+    <p class="follow-up-hint">
```
anpa, line 18: nit: is this class still needed?

anpa (approved):
Much better! Thanks for running it through design ❤️

templates/new_request_page.hbs
annterina: Good question!

I didn't find any references to it (in CSS files for example). So in that case that should be okay to remove?

anpa (approved):

PR #420, opened 2023-08-22
feat: added CC Field

Fredx87:
## Description

This PR adds a CC field to the new request form.

We decided to implement a custom component since Garden doesn't have a component that fits our use case.

In theory, it would be possible to use a Combobox, but the UX is not optimal since it doesn't make much sense to show a popup with suggested results when the user can input any value they want, and there isn't a set of predefined values.

The implementation resembles the existing component, and it is also similar to an [example](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/layout-grids/#ex2_label) shown in the ARIA Authoring Practices Guide website, where the list of recipients in rendered in a [grid](https://www.w3.org/WAI/ARIA/apg/patterns/grid/). For this reason, I used the grid container from Garden.

I tested the component with VoiceOver on Safari and NVDA on Firefox, but we'll need to wait for an A11Y audit to refine it and ensure it is properly accessible.

Keyboard interaction
- It is possible to enter a new e-mail and press Enter, Tab, Space, or Comma to add it to the CC list
- It is possible to move from the input to the CC list and back using Tab/Shift+Tab
- It is possible to move inside the CC list using arrow keys
- It is possible to delete a recipient by pressing Backspace when it is focused

Screen reader
- When an e-mail is added or removed there is an aria-live region that announces "[e-mail] has been added" or "[e-mail] has been removed"
- When the grid containing the list of CC is focused, the screen reader reads the label ("Selected CC e-mails") and describes it as a grid that can be navigated with the arrow keys
- When a recipient is focused, it says "Press Backspace to remove" and "invalid data" if the e-mail is invalid

Other notes:
- Like the previous implementation, it is possible to paste a list of e-mails separated by spaces, commas, or semicolons and they will be added to the list
- When an e-mail is invalid it is added to the list, but shown with a red background and not saved in the ticket. This is probably not accessible, and we should find a way to make it more clear that an e-mail is invalid
- The implementation is divided into a "container" hook which contains the logic related to A11Y, and a component for the UI. This is similar to the pattern used by Garden (even if it is a bit simplified), and ideally, the container can be ported to Garden and reused elsewhere.
- Since the CC field is available only for some users and only when a specific setting is enabled, it is lazy loaded.

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/13420283/308d76e3-cfb5-479b-9fea-8a5cabbd8b30

## Checklist

### DIFF
--- a/src/modules/new-request-form/NewRequestForm.tsx
+++ b/src/modules/new-request-form/NewRequestForm.tsx
@@ -29,6 +29,7 @@ const Footer = styled.div`
 `;
 
 const DatePicker = lazy(() => import("./fields/DatePicker"));
+const CcField = lazy(() => import("./fields/CcField/CcField"));
 
 export function NewRequestForm({
   ticketForms,
@@ -93,6 +94,12 @@ export function NewRequestForm({
                   setShowDueDate(value === "task");
                 }}
               />
+            );
+          case "cc_email":
+            return (
+              <Suspense fallback={<></>}>
+                <CcField field={field} />
+              </Suspense>
             );
           case "checkbox":
             return <Checkbox field={field} />;
--- a/src/modules/new-request-form/fields/CcField/CcField.tsx
+++ b/src/modules/new-request-form/fields/CcField/CcField.tsx
@@ -1,0 +1,147 @@
+import type { Field } from "../../data-types";
+import {
+  FauxInput,
+  Field as GardenField,
+  Hint,
+  Input,
+  Label,
+  Message,
+} from "@zendeskgarden/react-forms";
+import styled from "styled-components";
+import { Tag } from "@zendeskgarden/react-tags";
+import { focusStyles } from "@zendeskgarden/react-theming";
+import { useTagsInputContainer } from "./useTagsInputContainer";
+import { hideVisually } from "polished";
+import { useRef, useState } from "react";
+
+interface CcFieldProps {
+  field: Field;
+}
+
+const EMAIL_REGEX =
+  /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+export default function CcField({ field }: CcFieldProps): JSX.Element {
+  const { label, value, name, error, description } = field;
+  const initialValue = value
+    ? value.split(",").map((email) => email.trim())
+    : [];
+  const [tags, setTags] = useState(initialValue);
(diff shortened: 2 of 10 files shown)


src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -78,6 +79,12 @@ export function NewRequestForm({
           case "organization_id":
           case "tickettype":
             return <DropDown field={field} />;
+          case "cc_email":
+            return (
+              <Suspense fallback={<></>}>
+                <CcField field={field} />
```
luis-almeida, line 85: At some point we might need some error handling when lazy loading, e.g. providing a regular text field that asks for emails separated by `,`. In that case, it might be better to not lazy load here. No need to address now though.

src/modules/new-request-form/fields/CcField/CcField.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <input key={email} type="hidden" name={name} value={email} />
+      ))}
+      <AnnouncementMessage {...getAnnouncementProps()}>
+        {announcement}
+      </AnnouncementMessage>
+    </GardenField>
+  );
+}
+
+const Container = styled(FauxInput)`
```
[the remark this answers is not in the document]
Fredx87: Yeah, I agree. I personally think that the file is easier to read if the main exported component is on top, and the rest on the bottom. But since we used the opposite convention in other files, I will follow that. Maybe we can revisit it later

src/modules/new-request-form/fields/CcField/CcField.tsx
```diff
@@ -0,0 +1,147 @@
...
+                </span>
+              );
+            })}
+          </span>
+        </span>
+        <InputWrapper>
+          {/* Used to automatically resize the input based on the content */}
+          <InputMirror isBare aria-hidden="true" tabIndex={-1}>
+            {/* When the input is empty we show a space, otherwise the height of the mirror is less than the input */}
+            {inputValue || " "}
```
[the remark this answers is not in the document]
Fredx87: I'll give another try to see if it is fixable with CSS only

src/modules/new-request-form/NewRequestForm.tsx
Fredx87: I am not sure if it makes sense to have a functional fallback component because we would need to support 2 different implementations. For example, with the regular text field, we would still need to parse the value and create some hidden inputs to submit the values.

luis-almeida: I agree it's maybe it's overkill for error handling 👍🏼 
Let's see how large the js payloads will become. I'm thinking that in some extreme use case, e.g. Disney we might need to do some optimization and that may mean offering "smaller" components. Just something to keep in mind for later.

luis-almeida (approved):
👍🏼 🚀

PR #426, opened 2023-08-31
chore: linting fixes and improvements

luis-almeida:
## Description

I have just realised that we were only running the `checks` workflow on PRs against `master`.
This means that our all the PRs on `v4-alpha` were unchecked. In this PR I editing the action so checks run on PRs against all branches.

Keeping consistency around file/folder naming, app structure or export/import patterns is hard.
In this PR I'm also adding `ls-lint` to check for file and folder name patterns. There may be better tools so feel free to add feedback.

## Checklist


.ls-lint.yml
```diff
@@ -0,0 +1,12 @@
+ls:
+  templates:
+    .hbs: snake_case
+  translations:
+    .json: kebab-case
+  src:
+    .dir: kebab-case
+    .js: camelCase | PascalCase
+    .spec.js: camelCase | PascalCase
```
Fredx87, line 9: Shouldn't we check also the `.ts` files?

Fredx87 (commented):
> In this PR I'm also adding ls-lint to check for file and folder name patterns. There may be better tools so feel free to add feedback.

It seems that there are also some `eslint` plugins that are able to perform the same checks, like https://github.com/DukeLuo/eslint-plugin-check-file. I don't know if they are as powerful as `ls-lint`, I am ok to use `ls-lint` as well

luis-almeida:
I have opened a PR using https://github.com/DukeLuo/eslint-plugin-check-file with an alternative approach: https://github.com/zendesk/copenhagen_theme/pull/427

PR #427, opened 2023-09-05
chore: linting fixes and improvements

luis-almeida:
## Description

I have just realised that we were only running the `checks` workflow on PRs against `master`.
This means that our all the PRs on `v4-alpha` were unchecked. In this PR I editing the action so checks run on PRs against all branches.

Keeping consistency around file/folder naming, app structure or export/import patterns is hard.
In this PR I'm also adding bringing in `eslint-plugin-check-file` to check for file and folder name patterns. This is an alternative approach to using `ls-lint` tried in https://github.com/zendesk/copenhagen_theme/pull/426.

## Checklist


.eslintrc.js
```diff
@@ -0,0 +1,48 @@
...
+  rules: {
+    "@typescript-eslint/consistent-type-imports": "error",
+    "check-file/folder-naming-convention": [
+      "error",
+      { "src/**/": "KEBAB_CASE" },
+    ],
+    "check-file/filename-naming-convention": [
+      "error",
+      {
+        "src/**/*.{js,ts,tsx}": `@(${CAMEL_CASE}|${PASCAL_CASE})`,
```
luis-almeida, line 41: Unfortunately this is the only way to support multiple pattern per file name: https://github.com/dukeluo/eslint-plugin-check-file/issues/28#issuecomment-1643750137

I converted the file to `js` so we could at least name the variables but we can always inline the regex and keep it `.eslintrc.json`

sergioazevedo (approved):

Fredx87 (approved):

PR #428, opened 2023-09-06
feat: added attachment field

Fredx87:
## Description

This PR adds the management of attachments, using public APIs.

I tried to replicate the existing logic as closely as possible, I just changed a bit the error handling. In the existing implementation, there is a logic for displaying different error messages based on the response returned by the API. I tried to replicate the error cases, for example uploading a big file, but I wasn't able to trigger an error from the API, even with a fairly big attachment. I decided to implement a simple error handling for now, when we get an error either from the API or the client side upload, we just show a generic error message: `There was an error uploading {fileName}. Please try again or upload another file.`. We can eventually revisit in the future if we need more specific errors, but as I said it is not clear what are the possible failures. 
Looking at the current implementation, it seems that the eventual error is displayed inline near the attachments. I decided instead to use a Garden notification

It seems that there was also a logic for preventing the form submission while an upload was in progress, but I wasn't able to replicate this behavior, so for simplicity, the form is always submitted even if there are some uploads in progress. In this case, the pending uploads are simply not saved on the ticket.

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/13420283/e17e9fe4-cc7e-49b4-aa7f-81e9a65ffd49

## Checklist


src/modules/new-request-form/fields/attachments/Attachments.tsx
```diff
@@ -0,0 +1,192 @@
...
+import type { Field } from "../../data-types";
+import type { AttachedFile } from "./AttachedFile";
+import { FileListItem } from "./FileListItem";
+import { useAttachedFiles } from "./useAttachedFiles";
+
+interface AttachmentProps {
+  field: Field;
+}
+
+async function fetchCsrfToken() {
```
luis-almeida, line 26: Maybe we could turned this into a `useCsrfToken ` reusable hook and memoize it so we only get it once independently of where it's used.

src/modules/new-request-form/fields/attachments/Attachments.tsx
```diff
@@ -0,0 +1,192 @@
...
+  } = useAttachedFiles(
+    attachments?.map((value) => ({
+      status: "uploaded",
+      value,
+    })) ?? []
+  );
+
+  const { addToast } = useToast();
+
+  const notifyError = useCallback(
```
luis-almeida, line 62: Do we need the `useCallback` in this case?

src/modules/new-request-form/fields/attachments/FileListItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <FileList.Item>
+      <File type="generic" title={fileName} onKeyDown={handleFileKeyDown}>
+        {file.status === "pending" ? (
+          fileName
+        ) : (
+          <UploadLink isExternal href={file.value.url} target="_blank">
+            {fileName}
+          </UploadLink>
+        )}
+        {file.status === "pending" ? (
```
luis-almeida, line 54: Nit: maybe easier to follow if we combine the two `pending` checks.

src/modules/new-request-form/fields/attachments/Attachments.tsx
Fredx87: I am not sure how the CSRF token works. Does it expire? For example, if the user opens a page and fetches the CSRF token, then it stays idle for 1 hour, and then he starts an upload, is the token still valid?

If the answer is yes, we can use a hook and memoize it. Otherwise, it is better to always fetch it before using it.

luis-almeida: It is based on the session so in theory in your example it should still be valid as long as the session is alive.

> Otherwise, it is better to always fetch it before using it.

My only concern is if we need several times, e.g. upload and form submission, it may lead to unnecessary calls.
Totally fine to address it later on as well 👍🏼

src/modules/new-request-form/fields/attachments/Attachments.tsx
Fredx87: It seems that the `onDrop` from `react-dropzone` needs the `useCallback`: https://react-dropzone.js.org/. Our `onDrop` is calling `notifyError` so it needs to be passed in the dependency array of it and memorized as well.

src/modules/new-request-form/fields/attachments/useAttachedFiles.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const setUploaded = useCallback((pendingId: string, value: Attachment) => {
+    setFiles((current) =>
+      current.map((file) =>
+        file.status === "pending" && file.id === pendingId
+          ? { status: "uploaded", value }
+          : file
+      )
+    );
+  }, []);
```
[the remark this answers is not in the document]
Fredx87: As above, all these functions are used by the `onDrop` callback and they needs to be memorized as well

src/modules/new-request-form/fields/attachments/Attachments.tsx
```diff
@@ -0,0 +1,192 @@
...
+          // There is a bit of delay between the upload ending and the
+          // load event firing, so we don't want to set the progress to 100
+          // otherwise it is not clear that the upload is still in progress.
+          if (progress <= 90) {
+            setPendingFileProgress(pendingId, progress);
+          }
+        });
+
+        xhr.addEventListener("load", () => {
+          if (xhr.status < 400) {
```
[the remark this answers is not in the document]
Fredx87: Make sense. Probably we need to exclude also `3xx` statuses.

src/modules/new-request-form/fields/attachments/Attachments.tsx
Fredx87: If we can safely fetch it once it is better to memoize it to avoid unnecessary calls, but let's revisit it later when we'll need to use it in another places

PR #429, opened 2023-09-14
[A11y] Add aria-label to user navigation, wrap breadcrumbs in div

EmmaEwert:
## Description

> When ARIA regions, landmarks or HTML5 section elements are provided, users must be able to distinguish them from other regions, landmarks or sections in the page, particularly when two or more instances of the same type are used. When such an element does not provide a descriptive label to clearly identify itself, users of screen readers may have trouble locating the correct section or understanding its purpose.

This PR adds a localized aria-label for the header User Navigation. It also changes the breadcrumbs wrapping `nav` to a `div`, since the breadcrumbs helper itself provides an aria-labelled `nav`.

## Screenshots

Screenshot 2023-08-08 at 10 14 30

## Checklist

### DIFF
--- a/templates/article_page.hbs
+++ b/templates/article_page.hbs
@@ -1,6 +1,6 @@
 <div class="container-divider"></div>
 <div class="container">
-  <nav class="sub-nav">
+  <div class="sub-nav">
     {{breadcrumbs}}
     <div class="search-container">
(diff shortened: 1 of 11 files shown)


Fredx87 (approved):
👍

PR #430, opened 2023-09-20
feat: adding support for conditional fields

luis-almeida:
## Description

- [x] Introduces `end_user_conditions`
- [x] Aligns the types for the field `value` between ticket form fields and enduser condition child fields.
  - This means the value of checkboxes is now a `boolean` instead of `string` (`"on" | "off"`)
- [x] We must now keep `ticketFields` in local state so we can filter out conditional fields
- [x] Filtering logic for the visibility of ticket fields is encapsulated within `useEndUserConditions` hook

Still missing displaying `optional` conditional fields as we're waiting for a decision on the UI.

## Checklist

### DIFF
--- a/src/modules/new-request-form/data-types/Field.ts
+++ b/src/modules/new-request-form/data-types/Field.ts
@@ -1,12 +1,12 @@
 export interface Field {
+  id: number;
   name: string;
-  value: string;
+  value?: string | boolean;
   error: string;
   label: string;
   required: boolean;
   description: string;
   type: string;
-  id: string;
   options: FieldOption[];
   attachments?: Attachment[];
 }
--- a/src/modules/new-request-form/useEndUserConditions.tsx
+++ b/src/modules/new-request-form/useEndUserConditions.tsx
@@ -1,0 +1,21 @@
+import type { EndUserCondition, Field } from "./data-types";
+
+export function useEndUserConditions(
+  fields: Field[],
+  endUserConditions: EndUserCondition[]
+): Field[] {
+  return fields.filter((field) => {
+    const conditions = endUserConditions.filter((condition) =>
+      condition.child_fields.some((childField) => childField.id === field.id)
+    );
+
+    const hasNoConditions = conditions.length === 0;
+    const meetsAnyCondition = conditions.some(
+      (condition) =>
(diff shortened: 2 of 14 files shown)


sergioazevedo (approved):

src/modules/new-request-form/useEndUserConditions.tsx
```diff
@@ -0,0 +1,21 @@
...
+    );
+
+    const hasNoConditions = conditions.length === 0;
+    const meetsAnyCondition = conditions.some(
+      (condition) =>
+        fields.find((field) => field.id === condition.parent_field_id)
+          ?.value === condition.value
+    );
+
+    return hasNoConditions || meetsAnyCondition;
```
Fredx87, line 19: Do we need to check for `hasNoConditions`? Shouldn't `meetsAnyCondition` be false if `conditions` is empty?

Fredx87 (approved):

src/modules/new-request-form/useEndUserConditions.tsx
luis-almeida: > Shouldn't `meetsAnyCondition` be false if `conditions` is empty?

Yes and that is the problem because we want the return value to still be `true`.
So if `conditions` is empty, `hasNoConditions` is `true` and `meetsAnyCondition` is `false`. The return value will be `true` || `false` which is `true`.
If we just use `meetsAnyCondition` than it is always `false` so it will only show `fields` with `conditions`.

src/modules/new-request-form/data-types/Field.ts
```diff
@@ -1,12 +1,12 @@
 export interface Field {
+  id: number;
   name: string;
-  value: string;
+  value?: string | boolean;
```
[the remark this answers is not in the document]
luis-almeida: Couldn't agree more and I think in the end it would be best to end up with something like `type Field = BaseField | DueDateField | CheckboxField | AttachmentsField | ...` 👍🏼 

> If we need to merge this PR ASAP to solve the issues with the new data format, we can revisit the typings in a following PR

I thought it would be premature to do this now when we still haven't decided the final shape of all `Field`s is.
If it's ok with you let's take it in a follow up PR.

PR #433, opened 2023-09-25
feat: added Tagger field

Fredx87:
## Description

This PR adds support for the Dropdown field, adding a new `Tagger` component (since `Dropdown` was used for other dropdowns and the logic for this component is different because we need to support the nested structure). It also consolidates the logic for nesting in a `useNestedOptions` hook, which is reused for both `MultiSelect` and `Tagger`.

The logic for creating the nested options has changed a bit since I found an issue with the previously implemented logic. The issue was present when multiple sub-levels shared the same name, for example having options named like `Bass::Fender::Precision`, `Bass::Fender::Jazz`, `Guitar::Fender::Stratocaster`, `Guitar::Fender::Telecaster` would result in `Precision`, `Jazz`, `Stratocaster` and `Telecaster` grouped under the same `Fender` parent. The new logic uses the full path for groups, like `Bass::Fender` and `Guitar::Fender`, to avoid this issue.

I also changed how groups are identified, surrounding them with square brackets. This makes it easy to understand if a selected option value is a group or a single option. The `back` and `next` links are now both using the group identifier as a value, making it easier to display the correct options, without the need to have a stack of previously selected groups.

When an option in a sub-level is selected and the form is submitted with errors, the correct label is shown instead of the option value. This required the workaround of passing all options to the `Combobox` in the initial render since this behavior is not supported by default by the Garden component.

I also changed how a nested option is shown in the input, from e.g. `Bass::Fender::Precision` to `Bass > Fender > Precision`, since this is how these options are displayed on the request page.

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/13420283/7bbdd810-be24-4974-adcf-beb8acd5de8e

## Checklist


src/modules/new-request-form/fields/useNestedOptions.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+type OptionGroup = RootGroup | SubGroup;
+
+/**
+ * Each group is identified by a string in the format of [level1::level2::level3].
+ * The identifier is used as a key in the nested options object, and as a value
+ * in the combobox options used to navigate between different levels.
+ * The square brackets are used to distinguish between group identifiers and option values,
+ * since they are not allowed as characters in an option value.
+ */
```
anpa, line 35: Nice comments! ❤️

anpa (approved):
It looks great! Great work @Fredx87! 👏

anpa (approved):

PR #435, opened 2023-09-28
feat: added suggested articles

Fredx87:
## Description

This PR adds the suggested articles feature, porting the logic from the current implementation in Help Center.
For now, we are using the internal endpoint, but this should be changed when a public endpoint will be available.

I needed to change the colors of the links in the Garden theme, as we are doing in the Request List App. I changed a bit how the theme is created, moving the logic in the `shared.tsx` module and just passing the needed variables from the `hbs` template.

In the current implementation, the animation for the suggested articles box happens both when it is shown (with a `slideDown` animation) and when it is hidden (with a `slideUp` animation). I decided to keep only the first one, for two reasons:
-  Animating a React component that is added or removed from the DOM dynamically, both on enter and exit, requires an external library like `react-transition-group`. This wouldn't be a problem since the library is already used by Garden, but it would have made the code more complicated
- When the box is going to be hidden because there are no results, the links are removed immediately, and then the `slideUp` animation starts. Therefore the animation is applied only to the title and it is not very evident.

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/13420283/f4265660-40f7-4f19-8c88-6ac04ab3966f

## Checklist


templates/document_head.hbs
```diff
@@ -16,14 +16,11 @@
...
 
-    setupGardenTheme(function (defaultTheme) {
-        return {
-            ...defaultTheme,
-            colors: {
-                ...defaultTheme.colors,
-                foreground: "{{settings.text_color}}",
-                primaryHue: "{{settings.brand_color}}",
-            },
-        }
```
anpa, line 27: Is there an advantage for moving this logic to `shared.tsx`?

I am thinking that, since this template is editable in the theme editor, it would make it easier / straightforward to change the styling if we build the Garden theme here. What do you think?

src/modules/new-request-form/suggested-articles/SuggestedArticles.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+function hasMinLength(value: string): boolean {
+  const firstLetter = value.charCodeAt(0);
+  /*
+   * Special case considering CJK characters. Since ideographs represent
+   * whole words, we want to start searching when just two has been typed.
+   *
+   * Unicode range reference:
+   * http://www.unicode.org/versions/Unicode5.0.0/ch12.pdf#G16616
+   */
```
anpa, line 56: ❤️

anpa (approved):
It looks awesome! 👏👏👏

templates/document_head.hbs
Fredx87: It makes sense, I moved it to the `shared.tsx` because it is possible to import things from NPM modules, we especially need the `css` function from `styled-components`, but we can eventually re-export it from `shared.tsx` if we want to.

Let's merge this PR and eventually revisit it in another PR

PR #436, opened 2023-09-28
fix: added empty option for priority and type fields

Fredx87:
## Description

The `priority` and `type` fields have a default null value, which is displayed as `-` in the input. The same value was not listed in the options menu, so when a value was selected from the dropdown, it wasn't possible to go back to the null value.

This PR fixes this issue, adding the null option to the menu, only for optional fields.

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/13420283/7224223c-1e20-4830-acfd-7a3d1476d080

## Checklist

### DIFF
--- a/src/modules/new-request-form/NewRequestForm.tsx
+++ b/src/modules/new-request-form/NewRequestForm.tsx
@@ -117,15 +117,8 @@ export function NewRequestForm({
                 onChange={(value) => handleChange(field, value)}
               />
             );
-          case "priority":
           case "organization_id":
-            return (
-              <DropDown
-                key={field.name}
-                field={field}
-                onChange={(value) => handleChange(field, value)}
-              />
-            );
+          case "priority":
           case "tickettype":
             return (
               <DropDown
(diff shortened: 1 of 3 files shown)


src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -117,15 +117,8 @@ export function NewRequestForm({
...
-              <DropDown
-                key={field.name}
-                field={field}
-                onChange={(value) => handleChange(field, value)}
-              />
-            );
+          case "priority":
           case "tickettype":
             return (
               <DropDown
```
anpa, line 124: Curious, would it be possible to use `Tagger` here instead?

anpa (approved):
Good catch!

src/modules/new-request-form/NewRequestForm.tsx
Fredx87: In theory, it should be possible to use the same component for all the dropdowns.
Normal dropdowns don't support nesting and they don't require the complex logic implemented in the `Tagger` component, both for creating the nested options and passing the initial options. It is not a big deal to use the same component for both, they will just have root options and some unuseful code for the initial options.

There is also another caveat at the moment, since for the organizations dropdown we have a number as a value instead of a string for the options, and it requires us to convert them to string beforehand, so the `Tagger` component doesn't work.

Let's keep the two different components for now and eventually revisit them later when we improve the data passed from Curlybars.

PR #440, opened 2023-10-05
feat: [A11y] Add a button to skip to web widget

smansouri:
## Description

This PR adds a button to skip to web widget using keyboard navigation. 
The last commit is to make the skip navigations links to not change color after clicking as this was requested from accessibility team.
The js code checks if the widget doesn't exist after 500 ms then it hides the skip button but I had to add a timeout since the widget is loaded after some delay. There might be some slow widgets that load after 500ms and then the skip button is not rendered or we can increase the delay to 1s to decrease this chance.

https://github.com/zendesk/copenhagen_theme/assets/12232798/76b67ccc-b9d1-46ba-92ae-aa612ed2168a

Jira: https://zendesk.atlassian.net/browse/GS-2485

 

## Checklist

### DIFF
--- a/src/navigation.js
+++ b/src/navigation.js
@@ -30,6 +30,36 @@ window.addEventListener("DOMContentLoaded", () => {
     }
   });
 
+  // Hide Skip to Web Widget button if Web Widget is not present
+  const skipToWidgetButton = document.body.querySelector(".skip-to-web-widget");
+
+  let webWidgetLoadingCounter = 20;
+  let handleSkipToWidgetButtonIntervalId;
+  const handleSkipToWidgetButtonIntervalDuration = 500;
+
+  skipToWidgetButton.style.display = "none";
+
+  const handleSkipToWidgetButton = () => {
+    if ("zE" in window && window.zE("webWidget:get", "display") !== "hidden") {
+      skipToWidgetButton.addEventListener("click", () =>
+        window.zE(
+          window.zE.widget === "classic" ? "webWidget" : "messenger",
+          "open"
+        )
+      );
+      skipToWidgetButton.style.display = "inline-block";
+      clearInterval(handleSkipToWidgetButtonIntervalId);
+    }
+    if (webWidgetLoadingCounter-- <= 0) {
+      clearInterval(handleSkipToWidgetButtonIntervalId);
+    }
+  };
+
+  handleSkipToWidgetButtonIntervalId = setInterval(
+    handleSkipToWidgetButton,
+    handleSkipToWidgetButtonIntervalDuration
+  );
+
   // Toggles expanded aria to collapsible elements
   const collapsible = document.querySelectorAll(
     ".collapsible-nav, .collapsible-sidebar"
--- a/templates/header.hbs
+++ b/templates/header.hbs
@@ -1,4 +1,5 @@
 <a class="skip-navigation" tabindex="1" href="#main-content">{{t 'skip_navigation' }}</a>
+<a class="skip-to-web-widget" tabindex="2" href="#launcher" >{{t 'skip_to_web_widget'}}</a>
 
 <header class="header">
   <div class="logo">
--- a/style.css
+++ b/style.css
@@ -845,6 +845,9 @@ ul {
   top: auto;
   z-index: -999;
 }
+.skip-navigation:visited {
+  color: white;
+}
 [dir=rtl] .skip-navigation {
   left: initial;
   right: -999px;
--- a/styles/_header.scss
+++ b/styles/_header.scss
@@ -259,3 +259,32 @@ $header-height: 71px;
 #zd-modal-container ~ .skip-navigation {
   display: none;
 }
+
+.skip-to-web-widget {
+  align-items: center;
+  background-color: black;
+  color: white;
+  display: flex;
+  font-size: 14px;
+  justify-content: center;
+  left: -919px;
+  margin: 20px;
+  padding: 20px;
+  overflow: hidden;
+  position: absolute;
+  top: auto;
+  z-index: -999;
+}
+
+.skip-to-web-widget:focus, .skip-to-web-widget:active {
+  left: auto;
+  overflow: auto;
+  text-align: center;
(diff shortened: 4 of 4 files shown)


templates/header.hbs
```diff
@@ -1,4 +1,5 @@
 <a class="skip-navigation" tabindex="1" href="#main-content">{{t 'skip_navigation' }}</a>
+<a class="skip-to-web-widget" tabindex="1" href="#launcher" >{{t 'skip_to_web_widget'}}</a>
```
jgorfine-zendesk, line 2: In the demo video, I'm noticing that the "Skip to Main Content" button is in title case, while the "skip to web widget" button is all lowercase. Would you mind updating the Web Widget text to title case, so that it's consistent?

jgorfine-zendesk: Relatedly: do our customers (and our customers' customers) think of the messenger as Web Widget? Or as Messenger? We should use whatever terminology they're familiar with and/or expecting.

jgorfine-zendesk (changes requested):
Hi @smansouri, thanks for doing this! Caught a couple of typos.

src/navigation.js
```diff
@@ -30,6 +30,17 @@ window.addEventListener("DOMContentLoaded", () => {
     }
   });
 
+  // Hide Skip to Web Widget button if Web Widget is not present
+
+  setTimeout(() => {
+    const skipButton = document.body.querySelector(".skip-to-web-widget");
+    const widtge = document.body.querySelector("iframe[id='launcher']");
```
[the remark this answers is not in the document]
smansouri: Thanks.

templates/header.hbs
smansouri: Thanks  a lot Jen for reviewing. I actually changed the string to [skip to Chat ](https://github.com/zendesk/help_center/blob/main/config/locales/translations/help_center.en.yml#L6822).  The other button has [Skip to main content](https://github.com/zendesk/help_center/blob/main/config/locales/translations/help_center.en.yml#L6814C22-L6814C22). The Chat is with capital case since it's not supposed to be translated.

templates/header.hbs
```diff
@@ -1,4 +1,5 @@
 <a class="skip-navigation" tabindex="1" href="#main-content">{{t 'skip_navigation' }}</a>
+<a class="skip-to-chat-widget" tabindex="1" href="#launcher" >{{t 'skip_to_chat'}}</a>
```
smansouri, line 3: The string is "Skip to Chat" https://github.com/zendesk/help_center/blob/main/config/locales/translations/help_center.en.yml#L6830

bence-toth (approved):
LGTM 🙌

templates/header.hbs
jgorfine-zendesk: Sounds good! Thanks for clarifying. 🙂

src/navigation.js
```diff
@@ -30,6 +30,17 @@ window.addEventListener("DOMContentLoaded", () => {
...
+  // Hide Skip to Chat button if Chat is not present
+
+  setTimeout(() => {
+    const skipButton = document.body.querySelector(".skip-to-chat-widget");
+    const chatWidget = document.body.querySelector("iframe[id='launcher']");
+
+    if (!chatWidget) {
+      skipButton.style.display = "none";
+    }
+  }, 500);
```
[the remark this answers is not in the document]
anpa: Then, a Help Center can have 2 types of widgets: classic and messenger.

- Messenger seems to support `zE('messenger', 'open')` to open the widget ([docs](https://developer.zendesk.com/api-reference/widget-messaging/web/core/#open)).
- Classic seems to support `zE('webWidget', 'open')` to open the widget ([docs](https://developer.zendesk.com/api-reference/widget/core/#open)).

So we could use a logic like this, to open the widget after clicking the button:

```js
    const skipToWidgetButton = document.querySelector("#skipToWidget");
    if (window.zE) {
     	// web widget is present, add logic to open both the classic and the messenger widget
      skipToWidgetButton.addEventListener("click", () => window.zE(
        window.zE.widget === "classic" ? "webWidget" : "messenger", "open"
      ));
    } else {
      // no web widget is present
      skipToWidget.style.display = "none";
    }
```

What do you think @smansouri? It would also be nice to have @zendesk/emu blessing on this one, to make sure we are not missing something.

anpa (commented):
Thanks for working on this Sara and sorry for the delayed review 🙈 

I added some comments as I have some concerns with this solution. Please let me know what you think or if you'd like to go through them together.

I'd also like to ask @zendesk/emu to review this PR as owners of the web widget EUX.

templates/header.hbs
smansouri: I had added a string for `skip to Web Widget` before. Is that ok or you think it should be widget and not have web.

src/navigation.js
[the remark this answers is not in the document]
smansouri: Thanks Andre, I changed it to use zE which is indeed better than looking for id. However even with this we still need to have some delay since the zE.widget which you see it in this recording.

https://github.com/zendesk/copenhagen_theme/assets/12232798/08f82789-4133-4347-9122-1070255d5599

s after some milliseconds.

templates/header.hbs
smansouri: For some reason the translation for `skip_to_web_widget` is not found, although the [translations](https://github.com/zendesk/help_center/blob/main/config/locales/translations/help_center.en.yml#L6841) were merged long time ago in HC

Fredx87: I think you also need to add the string to the [I18nKeysMapping hash](https://github.com/zendesk/help_center/blob/960acd208ba0b79e2a75158083a8948f6ef982c5/app/presenters/hbs/helpers/global_helper.rb#L31) used by the `t` helper

smansouri: Thanks I made a PR https://github.com/zendesk/help_center/pull/28668

smansouri:
Closing this, since I couldn't get it approved.

PR #448, opened 2023-10-10
fix: remove aria-controls on follow button if it exists

smansouri:
## Description

The newest version of `reach/menu-button. This attribute is [not that well supported](https://a11ysupport.io/tech/aria/aria-controls_attribute) by screen readers. Although accessibility experts were a bit surprised that this check is failing because the zendesk a11y checks don't show any error for the new markup. This PR tries to remove aria-controls to resolve this error.

The other option I can think of is switching to garden menu.

## Checklist

### DIFF
--- a/script.js
+++ b/script.js
@@ -57,6 +57,12 @@
       });
     });
 
+    // remove aria-controls if it exists to fix an accessibility issue
+    const followButton = document.querySelector("#follow-menu-button");
+    if (followButton) {
+      followButton.removeAttribute("aria-controls");
+    }
+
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -57,6 +57,12 @@
       });
     });
 
+    // remove aria-controls if it exists to fix an accessibility issue
+    const followButton = document.querySelector("#follow-menu-button");
+    if (followButton) {
+      followButton.removeAttribute("aria-controls");
```
luis-almeida, line 63: This would be better fixed on the templating api side (Help Center).

PR #450, opened 2023-10-20
Return focus to ticket form field dropdown

luis-almeida:
## Description

There is a full page reload after the user selects a new ticket form. We want to return the focus to the dropdown for a better a11y.

## Checklist

### DIFF
--- a/src/modules/new-request-form/ticket-form-field/TicketFormField.tsx
+++ b/src/modules/new-request-form/ticket-form-field/TicketFormField.tsx
@@ -1,4 +1,5 @@
 import type { IComboboxProps } from "@zendeskgarden/react-dropdowns.next";
+import { createRef, useEffect } from "react";
 import {
   Combobox,
   Field as GardenField,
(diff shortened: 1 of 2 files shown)


src/modules/new-request-form/ticket-form-field/TicketFormField.tsx
```diff
@@ -30,10 +34,21 @@ export function TicketFormField({
         newUrl.searchParams.append(key, value);
       }
 
+      sessionStorage.setItem(key, crypto.randomUUID());
```
anpa, line 37: We don't seem to be using the value. Why setting it as `crypto.randomUUID()`?

I'm wondering if it would be helpful to make it both the key and the value more descriptive i.e. `setFocusToTicketFormField: true`? What do you think?

luis-almeida: Pushed a change aligning it with your feedback. Please have another look when you have time 🙏🏼

anpa (approved):
💥

PR #456, opened 2023-12-18
Use public endpoints for AnswerBot

Fredx87:
## Description

This PR updates the Answer Bot modal, using the public endpoints instead of the internal ones.
The logic for redirecting the user to a new page and showing a notification has been ported in the theme. For this reason, there is a new `notifications` module that exposes a function to add a notification in the browser session storage, and it is also loaded on each page from the `document_head` template for showing any eventual notification present in the session storage.

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/13420283/5e25a6d8-05e9-43b7-b94f-c105b48064a3

https://github.com/zendesk/copenhagen_theme/assets/13420283/59024a95-69c9-4136-bba5-31f26203f528

## Checklist


src/modules/new-request-form/answer-bot-modal/AnswerBotModal.tsx
```diff
@@ -110,31 +97,82 @@ export function AnswerBotModal({
...
+
+  const unsolvedNotificationAndRedirect = () => {
+    addFlashNotification({
+      type: "success",
+      message: t(
+        "new-request-form.answer-bot-modal.request-submitted",
+        "Your request was successfully submitted"
+      ),
     });
+    window.location.href = getUnsolvedRedirectUrl();
```
luis-almeida, line 120: Nit: location.assign()

templates/document_head.hbs
```diff
@@ -6,8 +6,18 @@
 {
   "imports": {
     "new-request-form-bundle": "{{asset 'new-request-form-bundle.js'}}",
+    "notifications-bundle": "{{asset 'notifications-bundle.js'}}",
     "new-request-form-translations-bundle": "{{asset 'new-request-form-translations-bundle.js'}}",
-    "vendor-bundle": "{{asset 'vendor-bundle.js'}}"
+    "vendor-bundle": "{{asset 'vendor-bundle.js'}}",
+    "addFlashNotification-bundle": "{{asset 'addFlashNotification-bundle.js'}}"
```
luis-almeida, line 12: If we really need the separate notifications bundle, could it also contain this api? It seems like it will be loaded anyway in all pages.

luis-almeida: > This is what rollup emits, it creates a bundle with the code shared between the new-request-form bundle and the notifications bundle

That's probably because we're importing `addFlashNotification` with a relative path and not from the already created `notifications-bundle`. In webpack that can be supported with aliases but then [eslint-import-resolver-webpack](https://www.npmjs.com/package/eslint-import-resolver-webpack#eslint-import-resolver-webpack) is needed to "fix" eslint. Rollup might have something similar but tbh I've never set it up and it just feels like going in the wrong direction.

> I don't know if it is possible to avoid the creation of this bundle

We just did something similar for the translations. Handling it with `manualChunks` instead of adding it to `input` should work:
```js
if (id.includes("notifications")) {
  return "notifications";
}
```

I'm mostly concerned about two points:

#### Updatability

The more bundles/chunks we have the harder it will be for us to document how to update the app, and ultimately for the user to do so. Especially since the bundles start to depend more and more on each other.
We could also consider renaming `vendor` to `shared` and have notifications be part of it. That would remove the need to separate notifications and we'd have one less bundle.

#### Performance

`notifications` is being added on all pages and it depends on `vendor` which now needs to be loaded as well and it's adding quite a lot of weight. My first thought was to put it within `{{#is signed_in}} ... {{/is}}` so it's at least not loaded for anonymous users. Can anonymous users see the answer bot? If so then that won't work.
Another option is to only load it if there are indeed notifications to show in the session.

Let me know your thoughts.

luis-almeida (approved):
I'm approving the PR but my reservations are still very much present.

I do agree that we should look into other ways of providing easier updates in the future but I am worried about the one we will have now.

I also think that forcing all users to always load this much JS when the vast majority of them will never need it is quite suboptimal and very hard to justify on our end. Even if it's non-blocking. But lets move forward and try to improve in a following step 👍🏼

PR #465, opened 2024-03-11
Simplify code splitting setup

luis-almeida:
## Description

We are currently building 7 assets due to lazy loading a few different components/modules.
When looking at the gzipped file sizes, we're not gaining much from doing it (if anything). Trimming down to 2 assets will make upgrading considerably simpler for the end user (customer).

- Renaming `theming` to `garden-theme` to make it more explicit
- Creating and configuring the theme explicitly instead of via a `theme` singleton

_Before:_

```bash
$ du -sh assets/*
 12K	assets/CcField.js
4.0K	assets/DatePicker.js
4.0K	assets/ThemeProviders.js
 48K	assets/new-request-form.js
4.0K	assets/theming.js
4.0K	assets/useModalContainer.js
1.5M	assets/vendor.js
```
```bash
$ gzip assets/* && du -sh assets/*
4.0K	assets/CcField.js.gz
4.0K	assets/DatePicker.js.gz
4.0K	assets/ThemeProviders.js.gz
 12K	assets/new-request-form.js.gz
4.0K	assets/theming.js.gz
4.0K	assets/useModalContainer.js.gz
368K	assets/vendor.js.gz
```

_After:_

```bash
$ du -sh assets/*
 60K	assets/new-request-form.js
1.5M	assets/vendor.js
```

```bash
$ gzip assets/* && du -sh assets/*
 16K	assets/new-request-form.js.gz
368K	assets/vendor.js.gz
```

## Checklist


Fredx87 (commented):
I agree that if we don't get much benefit from the lazy loading, it is better to simplify things and reduce the number of generated assets. I left a couple of comments.

I am not sure about the theme setup change. The previous setup with the "singleton" was made to be able to create a single theme object once and reuse it in all the components. With the new setup, each component will need to receive the theme settings and setup the theme object itself. There was also a point by @anpa in [this PR](https://github.com/zendesk/copenhagen_theme/pull/435#pullrequestreview-1652429504) about exposing the theme customization directly in the `.hbs` templates to make it customizable in Theming Center.
On the other hand, we have just one component for now, so maybe we don't need to find a solution for sharing the theme for now.

src/modules/theming/styled.d.ts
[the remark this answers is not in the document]
luis-almeida: Probably by mistake during rebase. Re-added in 3ebda24730341a9bdd46730f5cda2cd2b20368e6

package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "start": "concurrently -k -r 'rollup -c -w' 'wait-on script.js style.css && zcli themes:preview'",
     "build": "rollup -c",
+    "prebuild": "rm assets/*",
```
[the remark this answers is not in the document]
luis-almeida: That's a good point! I based it on the assumption that all `assets` will be generated but that might not be the case.
Removing only `*.js` files make sense 👍🏼  To be extra safe we could come um with a prefix for generated files and only remove those? e.g. `rm assets/build-*.js`

luis-almeida:
> The previous setup with the "singleton" was made to be able to create a single theme object once and reuse it in all the components.

I think the current approach has the pro of only having to do it once. The cons for me outweigh that pro:
- It's confusing because we see it being assigned in the template but then have no idea of how it's being used afterwards. Especially valid if you're working in Theming Center's editor as you don't have access to the JS code.
- Eventually all apps will be coupled in terms of versioning because they are all using an external theming package that requires the same set of settings. Updating one app requires updating theming, and updating theming can potentially require all apps to be updated as well.

> With the new setup, each component will need to receive the theme settings and setup the theme object itself.

I see this as a pro: a new app may require extra settings. If so, the app can then extend the theme with the styling/config that it needs without impacting the other apps.

> On the other hand, we have just one component for now, so maybe we don't need to find a solution for sharing the theme for now.

That is very much valid! It's good to try and foresee these problems but the truth is we don't have them yet.
I think it's nice that a simpler setup could also be more flexible. Even if it's not, we can tackle it later 👍🏼

package.json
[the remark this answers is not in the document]
Fredx87: It can be a good idea to prefix them since there can be some other `js` file which is not generated by the build process

Fredx87:
I think it is ok for now to move away from the singleton theme. Let's try the new approach and eventually we can revisit it later when we'll have more components

templates/document_head.hbs
```diff
@@ -5,8 +5,8 @@
 <script type="importmap">
 {
   "imports": {
-    "new-request-form": "{{asset 'new-request-form.js'}}",
-    "vendor": "{{asset 'vendor.js'}}"
+    "new-request-form-bundle": "{{asset 'new-request-form-bundle.js'}}",
```
Fredx87, line 8: I don't like that much that we will need to always import from `xyz-bundle`, it would be nicer if the rollup plugin generates the import map as
```json
{
    "new-request-form": "{{asset 'new-request-form-bundle.js'}}"
}
```

but maybe it is going to make the plugin code more complex, so it is also ok to keep the `bundle` suffix

Fredx87 (approved):

package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "start": "concurrently -k -r 'rollup -c -w' 'wait-on script.js style.css && zcli themes:preview'",
     "build": "rollup -c",
+    "prebuild": "rm assets/*-bundle.js",
```
Fredx87, line 8: Note: we need to use the `-f` option to avoid errors when there are no files matching the pattern

templates/document_head.hbs
luis-almeida: I agree and maybe would prefer to keep the import name clean.
I'm going to leave it like this for now. Should be an easy change 👍🏼

package.json
luis-almeida: Addressed in [8144437](https://github.com/zendesk/copenhagen_theme/pull/465/commits/8144437fada1f2244e7d99eae7d2c95778acaf79). Should also fix build.

Fredx87 (approved):

PR #467, opened 2024-03-19
Setup i18n for the new request form

Fredx87:
## Description

This PR adds internationalization for the New Request Form, using the [react-i18next](https://react.i18next.com/) library. Hardcoded strings are replaced with calls to the i18n library and some tooling has been added to load translations from the Zendesk CDN, to extract translation strings from the code, and to add linting rules to avoid hardcoded strings. It also moves a couple of elements (`Follow-up to request {{parentId}}` and `Fields marked with an asterisk (*) are required.`) from the Curlybars template to the JS component, to be translated in the same way.

### Translation loading

Translations can be downloaded from the Zendesk Static Assets CDN using the `bin/update-module-translations.mjs` script. The translation files are placed in the `src/modules/new-request-form/translations/locales` and are bundled by Rollup in a single bundle file.

This is not ideal because we are creating a big single bundle, but it is the best solution at the moment to avoid the creation of a lot of assets or depending on the URLs of the translations already published on the CDN.

The size of the single bundle with the translations is ~30 Kb when gzipped:
```sh
$ gzip assets/* && du -sh assets/*
 16K    assets/new-request-form-bundle.js.gz
 28K    assets/new-request-form-translations-bundle.js.gz
388K    assets/vendor-bundle.js.gz
```
### String extraction

We need to create a YML file containing the English strings which is picked up by our internal translation system. Developers can provide the English default value in the JS source code, and then run the `bin/extract-strings.mjs` to update the YML file, adding additional info (`screenshot` and `description`) manually. 

### Linting

I've added the `@shopify/jsx-no-hardcoded-content` rule from the `@shopify/eslint-plugin` to warn developers when hardcoded content is present in the code. This should be helpful to find when we are missing some translations, but it is not perfect and for this reason, it is reported as a warning and not an error.

Screenshot 2024-04-02 at 15 48 20

The rule is configurable and at the moment it will check for hardcoded content in component children, `aria-label` and `title` attributes, and the `content` prop of the Garden Tooltip component. It also excludes the `Trans` component from `react-i18next`. We can expand the rules in the future if we find that other checks are needed.

## Screenshots

Some screenshots of the New Request Form in English, Arabic, and Danish:

Screenshot 2024-04-02 at 15 59 44

Screenshot 2024-04-02 at 16 01 15

Screenshot 2024-04-02 at 16 01 44

## Checklist


luis-almeida:
Is the list in `guide_available_locales.json` different than what we get in https://static.zdassets.com/translations/new-request-form/manifest.json?

Fredx87:
> Is the list in `guide_available_locales.json` different than what we get in https://static.zdassets.com/translations/new-request-form/manifest.json?

Yes, I performed some checks on the two files to be sure that the implementation is right.

`guide_available_locales.json` contains 359 entries, while the manifest has 86 entries.

- A locale in `guide_available_locales.json` can have `"parent_identifier": null`. This means that is a standard locale and there is an entry in the manifest for it
  - I found an exception for this rule which is the `tl-ph` locale. I don't know if it is an error, but in this case, we would fallback to the English strings
- A locale in `guide_available_locales.json` can have a `parent_identifier`.  For example, the `nn` locale has `no` as a parent, and the `en-dk` has `en-150` as a parent.
  - In general, we can read the `parent_identifier` from the `guide_available_locales.json` and find the respective entry in the manifest
  - I just found out that in some cases there are multiple levels in the `parent_identifier`, for example, `no-nn` has `nn` as a parent which has `no` as a parent 🙈. This thing is not considered in the current implementation, because it will stop at `nn` and it will not find an entry in the manifest for it.

luis-almeida:
Thanks for explaining it in detail! ❤️ 

What I was trying to understand is if we really need to read from `guide_available_locales`.
In theory Help Center needs to support all available locales as there's no other product at Zendesk that supports more languages than Guide (HC).
So if the manifest is already listing all available translations, would it be enough to just use them all? And not even check `guide_available_locales`.

Fredx87:
> So if the manifest is already listing all available translations, would it be enough to just use them all? And not even check `guide_available_locales`.

I don't understand how this should work. The manifest lists all available translations, but Help Center has more locales that are mapped to a single translation locale.

For example, if the Help Center locale is `nn` we cannot find the `nn` entry in the manifest. Using `guide_available_locales` we can check the parent id and get `no`, and at this point, we can find the entry in the manifest and fetch the right translation file. The same concept applies to 282 locates that have a `parent_identifier` in `guide_available_locales`. I cannot think of a solution that doesn't involve `guide_available_locales`.

As I said, exposing this complexity in the theme and maintaining it in the long term is not ideal. I think that the best solution at this point is to pass this info from Curlybars, for example adding a `cldr_locale` attribute to the `help_center` object, and letting the backend perform the mapping.

package.json
```diff
@@ -11,6 +11,8 @@
     "download-locales": "node ./bin/update-translations",
     "test": "jest",
     "test-a11y": "node bin/lighthouse/index.js",
+    "i18n:extract": "node bin/extract-strings.mjs new-request-form 'src/modules/new-request-form/**/*.{ts,tsx}' src/modules/new-request-form/translations",
+    "i18n:update-translations": "node bin/update-module-translations.mjs new-request-form src/modules/new-request-form/translations/locales",
```
Fredx87, line 15: At the moment these scripts are referring only to the `new-request-form`. The setup can be revisited later when we'll have more components

Fredx87:
I updated the PR removing the mapping logic, and adding the download and bundling of the translation files

README.md
```diff
@@ -107,6 +107,54 @@ Notes:
...
+
+#### String extraction
+
+The `bin/extract-strings.mjs` script can be used to extract translation strings from the source code and put them in the YAML file that is picked up by our internal translation system. The usage of the script is documented in the script itself.
+
+The script wraps the `i18next-parser` tool and converts its output to the YAML format used internally. It is possible to use a similar approach in a custom theme, either using the standard `i18next-parser` output as the source for translations or implementing a custom transformer.
+
+#### Updating translation files
+
+Use the `bin/update-module-translations.mjs` to download the latest translations for a module. All files are then bundled by the build process in a single `[MODULE]-translations-bundle.js` file.
```
luis-almeida, line 156: Would it make sense to have it work for all "apps" and not just for a given module?
When updating translations we usually always want them all.

bin/update-module-translations.mjs
```diff
@@ -0,0 +1,32 @@
+/**
+ * This script is used for downloading the latest Zendesk official translation files for a module in the `src/module` folder.
+ * 
+ * Usage: node bin/update-module-translations.mjs [PACKAGE_NAME] [OUTPUT_PATH]
+ * Example: node bin/update-module-translations.mjs new-request-form src/modules/new-request-form/translations/locales
+ */
+import { writeFile } from "node:fs/promises";
+import { resolve } from "node:path";
+
+const PACKAGE_NAME = process.argv[2];
```
luis-almeida, line 10: Same thing here as it could also apply to all modules.

bin/update-module-translations.mjs
```diff
@@ -0,0 +1,32 @@
...
+const OUTPUT_PATH = resolve(process.cwd(), process.argv[3]);
```
luis-almeida, line 11: I'm also thinking that maybe we should not let the output path be configurable.
If it's set by default than we ensure that all apps are consistent when it comes to storing their locales.
One more argument for it is that the rollup config seems to expect translations to be in `src/modules/MODULE_NAME/translations/locales/*.json` so we really should not break away from it.

src/modules/i18n/addZendeskTranslations.ts
```diff
@@ -0,0 +1,20 @@
+import i18next from "i18next";
+
+interface TranslationFile {
+  locale: string;
+  rtl: boolean;
+  translations: Record<string, string>;
+  updated_at: string;
+}
+
+export async function addZendeskTranslations(
```
luis-almeida, line 10: Maybe calling it `localize` or `loadTranslations` would be better as `addZendeskTranslations` makes it zendesk specific and maybe confusing.

src/modules/new-request-form/renderNewRequestForm.tsx
```diff
@@ -3,12 +3,21 @@ import type { Settings } from "../garden-theme/createTheme";
...
   settings: Settings,
   props: NewRequestFormProps,
   container: HTMLElement
 ) {
+  const { baseLocale } = props;
+
+  initI18next(baseLocale);
+  await addZendeskTranslations(
+    baseLocale,
+    () => import(`./translations/locales/${baseLocale}.json`)
```
luis-almeida, line 18: Why not pass in the json already imported instead of the dynamic import?

.eslintrc.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
       version: "detect",
     },
   },
-  ignorePatterns: ["assets/**"],
+  ignorePatterns: ["assets/**", "bin/**"],
```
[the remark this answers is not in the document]
Fredx87: We are linting only the `src` folder from the `lint` script: https://github.com/zendesk/copenhagen_theme/blob/ee8ef959a8e9e4a0442616bea94715ef4bc1b257/package.json#L8

I added `bin` to `ignorePatterns` because I was getting errors like `'process' is not defined` in VSCode, but this can actually be solved just by adding the comment `/* eslint-env node */` on top of the file.

 I can change that, but in any case, we are not lining the `bin` files from the CI, they will be linted only the IDE

src/modules/i18n/addZendeskTranslations.ts
Fredx87: This function imports a file with a specific format, with the translation (key, value) map under `translations`. For this reason, I see it as "Zendesk-specific". 

Alternatively, we can put in the translation files just the translation map, like we are doing with the translation of the settings since we don't need the other info.

Another possible option is to remove this function and load the translations directly in each app (see next comment)

src/modules/new-request-form/renderNewRequestForm.tsx
Fredx87: It is because the `addZendeskTranslations` function contains a try-catch around the dynamic import, to avoid throwing if the requested file is not present.

If we import the file directly here, we need to wrap it in a try-catch. At that point, the `addZendeskTranslations` is more or less useless and we can just move the logic here.

Fredx87:
I updated the PR, now the script fetches all the translation files at once. I also changed how we save the files, removing the things we don't need and storing only the mapping between the keys and the values.

src/modules/new-request-form/renderNewRequestForm.tsx
luis-almeida: Got it! One can argue that it is not a traditional `fetch` call and the file should actually exist and be loaded properly for the app to work as expected.
Where I am getting at is that this might actually be one of the errors that we do want.

Totally fine to keep this behaviour though 👍🏼

bin/update-modules-translations.mjs
```diff
@@ -0,0 +1,51 @@
...
+ */
+import { writeFile } from "node:fs/promises";
+import { resolve } from "node:path";
+
+/**
+ *  Maps each folder in the `src/modules` directory with its package name on the translation system
+ */
+const MODULES = {
+  "new-request-form": "new-request-form",
+};
```
luis-almeida, line 14: I was thinking if we could eventually find a way to have this work in a "convention over configuration" type of way where it wouldn't be needed to manually keep this map. Maybe by collocating the translation within each app and check which module contains those file.
Also totally fine to keep it this way as we can always further improve.

luis-almeida (approved):
Thanks for addressing the comments.
Looking pretty good 👍🏼

bin/update-modules-translations.mjs
Fredx87: > Got it! One can argue that it is not a traditional fetch call and the file should actually exist and be loaded properly for the app to work as expected.
Where I am getting at is that this might actually be one of the errors that we do want.
>
> Totally fine to keep this behaviour though 👍🏼

We don't have the guarantee that a file for a `base_locale` exists at runtime. For example, at the moment we have a `tl-ph` locale that doesn't have a `parent_identifier`, but we don't have translations for that locale.

It is probably an error in the JSON containing the mapping, but even if it gets fixed it can happen again in the future when new locales are added, or if for some reason we don't have a translation file for a locale available.

If we don't handle the error, the app will crash immediately. I think it is better to handle the error and let the app use the English strings as a fallback.

> I was thinking if we could eventually find a way to have this work in a "convention over configuration" type of way where it wouldn't be needed to manually keep this map. Maybe by collocating the translation within each app and check which module contains those file.
Also totally fine to keep it this way as we can always further improve.

It is a good point. Actually, we already have the package name in the translations YAML file. It is possible to change the script to read the package name from there, instead of manually keeping the map.
Let's merge this PR and then we can eventually revisit it

PR #472, opened 2024-04-17
fix: fixed organization dropdown and ticket prefilling

Fredx87:
## Description

This PR fixes the value selection and the prefilling of the organization field and the prefilling of the CC field, the anonymous requester email, and the due date field.

These issues were introduced when [we moved some fields out of the ticket_fields array](https://github.com/zendesk/copenhagen_theme/pull/457). Those features stopped working since the value tracking of the controlled components and the prefilling were still working only on the ticket_fields array.

Prefill changes:
- Passing down and returning also the fields that are outside the `ticket_fields` array
- Wrapped the function in a `useMemo` since we don't need to recalculate them on each rerender
  - Changed `usePrefilledTicketFields` to a normal function called  `prefillTicketFields` - it doesn't need to be a hook since we are not using any other hook inside it

Added a new `useState` to keep track of the organization field value. We don't need it for the other fields since they are implemented as uncontrolled components.

## Checklist

### DIFF
--- a/src/modules/new-request-form/NewRequestForm.tsx
+++ b/src/modules/new-request-form/NewRequestForm.tsx
@@ -81,8 +81,24 @@ export function NewRequestForm({
     description_mimetype_field,
   } = requestForm;
   const { answerBot } = answerBotModal;
-  const prefilledTicketFields = usePrefilledTicketFields(ticket_fields);
+  const {
+    ticketFields: prefilledTicketFields,
+    emailField,
+    ccField,
+    organizationField: prefilledOrganizationField,
+    dueDateField,
+  } = usePrefilledTicketFields({
+    ticketFields: ticket_fields,
+    emailField: email_field,
+    ccField: cc_field,
+    organizationField: organization_field,
+    dueDateField: due_date_field,
+  });
+
   const [ticketFields, setTicketFields] = useState(prefilledTicketFields);
+  const [organizationField, setOrganizationField] = useState(
+    prefilledOrganizationField
+  );
   const visibleFields = useEndUserConditions(ticketFields, end_user_conditions);
   const { formRefCallback, handleSubmit } = useFormSubmit(ticketFields);
   const { t } = useTranslation();
@@ -95,6 +111,14 @@ export function NewRequestForm({
           : ticketField
       )
     );
+  }
+
+  function handleOrganizationChange(value: string) {
+    if (organizationField === null) {
+      return;
+    }
+
+    setOrganizationField({ ...organizationField, value });
   }
 
   return (
@@ -131,19 +155,15 @@ export function NewRequestForm({
         {ticket_form_field.options.length > 0 && (
           <TicketFormField field={ticket_form_field} />
         )}
-        {email_field && (
-          <Input
-            key={email_field.name}
-            field={email_field}
-            onChange={(value) => handleChange(email_field, value)}
-          />
-        )}
-        {cc_field && <CcField field={cc_field} />}
-        {organization_field && (
+        {emailField && <Input key={emailField.name} field={emailField} />}
+        {ccField && <CcField field={ccField} />}
+        {organizationField && (
           <DropDown
-            key={organization_field.name}
-            field={organization_field}
-            onChange={(value) => handleChange(organization_field, value)}
+            key={organizationField.name}
+            field={organizationField}
+            onChange={(value) => {
+              handleOrganizationChange(value);
+            }}
           />
         )}
         {visibleFields.map((field) => {
@@ -216,7 +236,7 @@ export function NewRequestForm({
                   />
                   {field.value === "task" && (
                     <DatePicker
-                      field={due_date_field}
+                      field={dueDateField}
                       locale={baseLocale}
                       valueFormat="dateTime"
                     />
--- a/assets/new-request-form-bundle.js
+++ b/assets/new-request-form-bundle.js
@@ -18,7 +18,9 @@ function Input({ field, onChange }) {
     if (type === "decimal")
         stepProp.step = "any";
     const autocomplete = type === "anonymous_requester_email" ? "email" : undefined;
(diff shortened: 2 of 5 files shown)


src/modules/new-request-form/prefillTicketFields.tsx
```diff
@@ -31,9 +22,47 @@ const ALLOWED_HTML_TAGS = [
...
+      return prefilledTicketFields.ticketFields.find(
+        (field) => field.name === `request[${id}]`
+      );
+  }
+}
+
+export function prefillTicketFields(fields: Fields): Fields {
   const { href } = location;
   const params = new URL(href).searchParams;
+  const res: Fields = {
```
luis-almeida, line 62: Nit: can we name this `prefilledFields` instead of just `res`?

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -73,8 +73,28 @@ export function NewRequestForm({
...
     description_mimetype_field,
   } = requestForm;
-  const prefilledTicketFields = usePrefilledTicketFields(ticket_fields);
+  const {
+    ticketFields: prefilledTicketFields,
+    emailField,
+    ccField,
+    organizationField: prefilledOrganizationField,
+    dueDateField,
+  } = useMemo(
```
luis-almeida, line 82: Why do we need the `useMemo` now and could it not be moved inside the hook if really needed?
We've been following the `use` naming whenever possible so it feels pretty weird to change to an imperative call.

Fredx87: General consideration:
At the moment, we have a couple of "hooks" that are not actually hooks (`usePrefilledTicketFields` and `useEndUserConditions`). Those are just functions because they do not use any other hook inside, and they shouldn't be named with the `use` prefix, as also [stated in the React docs](https://react.dev/learn/reusing-logic-with-custom-hooks#should-all-functions-called-during-rendering-start-with-the-use-prefix):
> Functions that don’t call Hooks don’t need to be Hooks.
>
> If your function doesn’t call any Hooks, avoid the `use` prefix. Instead, write it as a regular function without the `use` prefix.
>
> Technically, this isn’t enforced by React. In principle, you could make a Hook that doesn’t call other Hooks. This is often confusing and limiting so it’s best to avoid that pattern.

I also think that in the future it would be nice to extract these pieces of logic into an SDK, to make them reusable in custom themes, and it would be nice to have these things as regular functions to make them re-usable in different frameworks. But this is a consideration for the future.

For this particular function:
> Why do we need the `useMemo` now and could it not be moved inside the hook if needed?

`useMemo` is just to avoid running the function on every re-render, since we need it only on the first render. 
Without memoizing it, it runs every time the user changes a value in some field. I don't think it was causing any issues, but it is probably better to not re-run it every time. 

I decided to put the `useMemo` in the component, but it is also fine to move it to the `usePrefilledTicketFields`, which at this point will be a hook

luis-almeida: That is valid now as it is from the new docs website. The [old react documentation](https://legacy.reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook) just mentioned:

> A custom Hook is a JavaScript function whose name starts with ”use” and that may call other Hooks.

That's why it became a naming convention that is widely used, and we see it even in projects of hook collections.
It has also been abused but I don't think that's the case here. As you mentioned, if the `useMemo` was added in the hook file, it would simplify the parent component and there would be no need to keep on renaming the file.

In a long running codebase, it is better to have one suboptimal pattern but that is consistently used than having several better patterns that differ from each other. So if we're to move away from this pattern I'd prefer that we do it across the codebase. And if we're really sticking to the docs we should probably name it `getPrefilledTicketFields` instead of `prefillTicketFields` which makes it seem like an imperative call.

When it comes to `useMemo` I think we might be over optimising and the same applies to `useCallback` at times.
If the performance gain is never going to be perceptible I'd just prioritize simplicity.
Mentioning this as General consideration as well. I'm fine fine with keeping it though 👍🏼

Fredx87: If the pattern is "we should name all the external functions used in components with the `use` prefix", I kindly disagree with it.

- A function prefixed with `use` is considered a hook in the React world, and some tools like `eslint-plugin-react-hooks` are enforcing these functions to follow the rules of hook
- A hook is a function that "hooks" into the React lifecycle, if a function is not calling any other hook in its body, it is not a custom hook and we shouldn't name it with the `use` prefix
- This is clearly stated in the new React docs, and we should not derive from it to avoid confusion
- hooks collections are collections of custom hooks that are following the rules above

> When it comes to useMemo I think we might be over optimising and the same applies to useCallback at times.
If the performance gain is never going to be perceptible I'd just prioritize simplicity.

It is not just a matter of performance optimization but also correct behavior. Re-calculating all the values of the fields from the URL at each re-render could cause some unwanted behavior. At the moment we are just using them as initial values in `useState` and `defaultValue` in some components, but if in the future we are going to change how we use these values there will be the possibility that we end up with misaligned values between the form and the ticket fields.

The same applies to `useCallback`, we are using it
- in ref callbacks. In this case, we want to run the callback only on the first render (e.g. for initializing the WYSIWYG editor. If we don't use `useCallback` the editor would be re-initialized at every re-render)
- when we pass callbacks down to some Garden components. These components are using those callbacks as dependencies for other hooks, and if we don't wrap them in `useCallback` we can have unwanted behaviors like effects re-running because the callback changed. And once we we wrap something in `useCallback` we then need to wrap any dependency in a `useCallback` as well. I agree that it complicates things but unfortunately, it is how React works

Fredx87:
I restored the `usePrefilledTicketFields` hook and moved the `useMemo` inside, it should be ready for another review

luis-almeida (approved):
Needs rebasing

anpa (approved):

PR #473, opened 2024-04-18
Use help-center-wysiwyg package

luis-almeida:
## Description

The help-center-wysiwyg is being open sourced and this PR starts making use of that package in the new request form.
There are a few caveats to be aware of:

- The `help-center-wysiwyg` package is not yet released
- The repo is still private while we wait for the license to be ready
- In order to test the implementation:
  - We're installing from a branch that contains the `dist` build
  - We're setting offline caching so installing the dependencies works on github actions as well
  - This can all be removed once the repo is made public and the `dist` is published to npm


Fredx87 (approved):
Some considerations:

- Adding the offline caching and checking in 1800+ tgz files just to test the package seems a bit overkill to me, even if we are going to remove them later
  - A simpler solution would be to run `yarn pack` in the wysiwyg repo, copying the generated tzg file in this repo, and then run ` yarn add file:zendesk-help-center-wysiwyg-v0.0.1.tgz`. In this way, we would need to check in just the wysiwyg tgz
- Based on our rollup config, the WYSIWYG will be bundled in the `vendor-bundle.js` file, increasing its size a lot (~250Kb gzipped). It would be better to create a separate bundle for it and eventually lazy load it as we were doing in Help Center
- The wysiwyg package has some dependencies that are probably needed only for building the package itself (`@ckeditor/ckeditor5-dev-utils`, `@ckeditor/ckeditor5-dev-webpack-plugin`, `babel-loader`). We should move them to `devDependencies` in the wysiwyg repo.

Since this is a temporary implementation I am also fine to merge this PR and revisit these points later

luis-almeida:
> - A simpler solution would be to run `yarn pack` in the wysiwyg repo, copying the generated tzg file in this repo, and then run ` yarn add file:zendesk-help-center-wysiwyg-v0.0.1.tgz`. In this way, we would need to check in just the wysiwyg tgz

I've never tried this so I'm trying to understand: do you mean an offline mirror but just with one `tzg`?
So yarn would download all the packages but this one?

Fredx87:
> > * A simpler solution would be to run `yarn pack` in the wysiwyg repo, copying the generated tzg file in this repo, and then run ` yarn add file:zendesk-help-center-wysiwyg-v0.0.1.tgz`. In this way, we would need to check in just the wysiwyg tgz
> 
> I've never tried this so I'm trying to understand: do you mean an offline mirror but just with one `tzg`? So yarn would download all the packages but this one?

Basically yes. It adds an entry like this in the `package.json`: 
```json
"@zendesk/help-center-wysiwyg": "file:zendesk-help-center-wysiwyg-v0.0.1.tgz",
```

And it will install that specific package from the tgz instead of looking for it on NPM. All its dependencies are still fetched from NPM

Fredx87 (approved):

PR #474, opened 2024-04-22
chore: added translation string for the New Request Form

Fredx87:
## Description

This PR adds a string to be used in the New Request Form

## Checklist

### DIFF
--- a/src/modules/new-request-form/translations/en-us.yml
+++ b/src/modules/new-request-form/translations/en-us.yml
@@ -142,3 +142,8 @@ parts:
       title: "Message shown on top of the New Request Form, informing the user that required fields are marked with an asterisk"
       screenshot: "https://drive.google.com/file/d/1EXEh7S9_GZrz4WBVcHWM8DmXetcXoHDu/view?usp=drive_link"
       value: "Fields marked with an asterisk (*) are required."
+  - translation:
+      key: "new-request-form.credit-card-digits-hint"
+      title: "Additional label shown above the credit card field, informing the user that only the last 4 digits are required."
+      screenshot: "https://drive.google.com/file/d/13Pz2p5q7FHX9FKT6MUR6Sg5vWy55X0kP/view?usp=drive_link"
+      value: "(Last 4 digits)"


src/modules/new-request-form/translations/en-us.yml
```diff
@@ -142,3 +142,8 @@ parts:
       title: "Message shown on top of the New Request Form, informing the user that required fields are marked with an asterisk"
       screenshot: "https://drive.google.com/file/d/1EXEh7S9_GZrz4WBVcHWM8DmXetcXoHDu/view?usp=drive_link"
       value: "Fields marked with an asterisk (*) are required."
+  - translation:
+      key: "new-request-form.credit-card-digits-hint"
+      title: "Additional label shown above the credit card field, informing the user that only the last 4 digits are required."
+      screenshot: "https://drive.google.com/file/d/13Pz2p5q7FHX9FKT6MUR6Sg5vWy55X0kP/view?usp=drive_link"
+      value: "(Last 4 digits)"
```
Fredx87, line 149: @kubraokcu Should we place the parenthesis in the string like this, or in the source code?

kubraokcu: > > it can be Credit Card Label (Last 4 digits) or Credit Card Label* (Last 4 digits). Would it be ok to have the string like {{defaultLabel}} (Last 4 digits) with defaultLabel replaced with the customer label with or without the asterisk?
> {{defaultLabel}} (Last 4 digits) 

That'd be perfect :)

> Actually, we would probably end up with <defaultLabel /> (Last 4 digits) because the * is inside a <span> tag and the library we are using requires this format when interpolating HTML tags. I asked the i18n team if is okay to put HTML placeholders in the string.

I saw your message, thank you for asking them!

kubraokcu (approved):

src/modules/new-request-form/translations/en-us.yml
kubraokcu: Concatenation accepted because of library constraints to use accepted placeholders

PR #479, opened 2024-04-26
fix: ui fixes for WYSIWYG and Datepicker error states

Fredx87:
## Description

This PR fixes some UI issues:

- WYSIWYG
  - Added top and bottom margins to separate the editor from the surrounding elements
  - Added a red border when the field is invalid (implemented in the WYSIWYG package - I'm bringing in a `tgz` with a new version number since the package is not published yet)
- DatePicker
  -  Added a top margin to the error message, to separate it from the input. It should be added automatically by Garden but it seems that there is a bug
  - Added the `validation` prop to the input for displaying a red border when there is an error

## Screenshots

Before:
Screenshot 2024-04-25 at 12 48 36

Screenshot 2024-04-25 at 12 48 46

After:
Screenshot 2024-04-26 at 15 13 15

## Checklist


src/modules/new-request-form/fields/textarea/TextArea.tsx
```diff
@@ -19,6 +20,18 @@ interface TextAreaProps {
   onChange: (value: string) => void;
 }
 
+const StyledField = styled(GardenField)`
+  .ck.ck-editor {
```
luis-almeida, line 24: Could this be added from above as `margin-bottom` so we wouldn't need to target `ck-editor` classes directly?

luis-almeida (approved):

src/modules/new-request-form/fields/textarea/TextArea.tsx
Fredx87: This implementation is coherent with the one used for the Garden `Textarea`, where the margin-top is added to textarea

luis-almeida (approved):

PR #492, opened 2024-06-20
fix: clearing date field when it is hidden

gosiexon-zen:
## Description

When a Date field was conditional field it cleared its data when being hidden. Now it use onChange like the other fields so the date is preserved
[Jira issue](https://zendesk.atlassian.net/browse/GG-3487)

## Screenshots

https://github.com/zendesk/copenhagen_theme/assets/83698095/416de6c7-489c-4e8e-89eb-8b00b03806b2

## Checklist


Fredx87 (commented):
There is an issue with the Due Date field, which is shown when the ticket type is "Task". When the date is changed, the ticket type is reset to the null value. I left a comment and it should be enough to fix it.

We should also commit the minified JS bundles instead of the ones used in development, running `yarn build` before committing. We wouldn't need this step once we merge the v4 branch on master, since the release task will run the build command. For now, we need to do it manually to ensure that we have the minified files.

Fredx87:
There is another issue with the Due Date field that we can fix in this PR as well. If the user inserts a value in the Due Date field and then changes a field that triggers some end-user conditions, the value in the due date field is cleared

PR #505, opened 2024-08-01
Lookup field implementation

gosiexon-zen:
## Description
This code add support for Lookup fields for custom objects in the Copenhagen Theme v4. 

[Jira GG-3522](https://zendesk.atlassian.net/jira/software/c/projects/GG/boards/1258?selectedIssue=GG-3522)
## Screenshots

https://github.com/user-attachments/assets/b3580e20-9c4c-437a-9200-31c7fd846020

## Checklist


src/modules/new-request-form/useFormSubmit.tsx
```diff
@@ -24,7 +24,6 @@ export function useFormSubmit(ticketFields: Field[]): UseFormSubmit {
     (ref: HTMLFormElement) => {
       if (ref && !initialized.current) {
         initialized.current = true;
-
```
gosiexon-zen, line 27: I'll revert that change

olivia-tsao (approved):

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+          } catch (error) {
+            return error;
+          } finally {
+            setIsLoadingOptions(false);
+          }
+        }
+      }
+      return;
+    },
+    [customObjectKey, fetchSelectedOption]
```
Fredx87, line 142: According to ESLint, we should add more things to the dependencies array:
> React Hook useCallback has missing dependencies: 'fieldId', 'onChange', and 'userId'. Either include them or remove the dependency array. If 'onChange' changes too often, find the parent component that defines it and wrap that definition in useCallback.

And I guess at this point we should wrap the `handleChange` in the `NewRequestForm` component with a `useCallback` as well

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+  useEffect(() => {
+    return () => debounceHandleChange.cancel();
+  }, [debounceHandleChange]);
+
+  useEffect(() => {
+    if (value && !options.find((option) => option.value === value)) {
+      fetchSelectedOption(value as string);
+      return;
+    }
+  }, [value]);
```
Fredx87, line 159: Do we need to re-run the effect every time the `value` changes?

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const handleRenderValue = useCallback(() => {
```
Fredx87, line 161: Do we need this function and a `useCallback`? Can we simply inline it in the `Combobox` props?

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+        if (inputValue === "") {
+          setOptions([]);
+          setSelectedOption(null);
+          onChange("");
+        } else {
+          const searchParams = new URLSearchParams();
+          searchParams.set("name", inputValue.toLocaleLowerCase());
+          searchParams.set("source", "zen:ticket");
+          searchParams.set("field_id", fieldId.toString());
+          searchParams.set("user_id", userId.toString());
```
Fredx87, line 112: We need to pass also the `organization_id`.

> - organization_id:
>    - if the user belongs to a single org, the id of the org
>    - if the user belongs to multiple orgs, the id of the org selected in the organization dropdown

Fredx87 (changes requested):
I left some comments in the PR.

Other things to improve:
- I created a [lookup-fields](https://github.com/zendesk/copenhagen_theme/tree/lookup-fields) branch. Let's set it as a target branch for the PR and rebase the PR on top of it, so when the PR is ready we can merge it there and eventually iterate on the implementation before merging it on master and releasing a new version
- Let's rewrite and simplify the commit history. Every commit with `feat` and `fix` will end up in the changelog. Ideally, we should have only a single commit similar to `feat: added lookup fields` for this PR
- I don't like how the lookup field works, especially the fact that the user needs to manually delete the content to be able to search for a different string. I think we should align the behavior with the Lookup field in Support, but we can ask a feedback to design as well

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const handleChange = useCallback<NonNullable<IComboboxProps["onChange"]>>(
+    async ({ inputValue, selectionValue }) => {
+      setIsFirstLoad(false);
+      if (selectionValue !== undefined) {
+        if (selectionValue == "") {
+          setSelectedOption(EMPTY_OPTION);
+          setInputValue(EMPTY_OPTION.name);
+          onChange(EMPTY_OPTION.value);
+        } else {
+          await fetchSelectedOption(selectionValue as string);
```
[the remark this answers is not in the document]
gosiexon-zen: No we don't have, options are set only on search as they are based on the inputValue.

src/modules/new-request-form/translations/en-us.yml
```diff
@@ -25,7 +26,7 @@ parts:
       title: "Title of the footer of the Answer Bot Modal"
       screenshot: "https://drive.google.com/file/d/14qsEyUiVTy9MRYBGkgVWLmlduhoscYBw/view?usp=drive_link"
       value: "Does this article answer your question?"
-      obsolete: "2024-09-22"
+      obsolete: "2024-11-07"
```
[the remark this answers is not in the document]
gosiexon-zen: Ok I changed it back.

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const [isLoadingOptions, setIsLoadingOptions] = useState<boolean>(false);
+  const [isFirstLoad, setIsFirstLoad] = useState<boolean>(true);
+  const { t } = useTranslation();
+
+  const customObjectKey = getCustomObjectKey(
+    relationship_target_type as string
+  );
+
+  const loadingOption = {
+    name: t("new-request-form.lookupfield.loading-options", "Loading items..."),
```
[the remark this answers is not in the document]
gosiexon-zen: Right, I changed it in translations PR and forget to change here also, thanks!

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+    id: "loading",
+  };
+
+  const noResultsOption = {
+    name: t(
+      "new-request-form.lookupfield.no-matches.found",
```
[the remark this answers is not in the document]
gosiexon-zen: Changed

templates/new_request_page.hbs
```diff
@@ -21,32 +29,24 @@
...
-      isSignedIn: {{json signed_in}},
-      helpCenterPath: {{json (page_path 'help_center')}},
-      requestsPath: {{json (page_path 'requests')}},
-      requestPath: {{json (page_path 'request' id=answer_bot.request_id)}}
-    },
-  };
-
-  renderNewRequestForm(settings, props, container);
-</script>
+  import { renderNewRequestForm } from "new-request-form"; const container =
```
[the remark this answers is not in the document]
gosiexon-zen: Ah yes, I have prettier installed but I see we don't use it on .hbs files

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Ok I will try this

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Probably not, it will run only when we have field prefiled or when we refresh the form on submit error.

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: no we can do this inline, in Lotus there linter didn't allow inline calls and requires useCallback 🙄

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,249 @@
...
+  const debounceHandleChange = useMemo(() => debounce(handleChange, 300), []);
+
+  useEffect(() => {
+    return () => debounceHandleChange.cancel();
+  }, [debounceHandleChange]);
+
+  useEffect(() => {
+    if (value && !options.find((option) => option.value === value)) {
+      fetchSelectedOption(value as string);
+      return;
```
Fredx87, line 187: Why do we need this return?

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }, []);
```
Fredx87, line 189: Reasoning again about this, I think we should simplify things a bit.

This component calls `onChange` when the value changes, and then it will receive again the new `value` from the props from the main component. If I understand correctly, we want to set the `selectedOption` when the user selects an option or at the first render if we have a value.

I think we have three different ways of handling it.

**Option 1**
- As mentioned in the previous comment, in `handleChange` we can search for the selected option in the `options` array,  set the `selectedOption` as the option found, and then call `onChange` and `setInputValue` directly skipping `fetchSelectedOption`
- This `useEffect` can remain with `[]` as a dependency because we want to call `fetchSelectedOption` only on the first render

**Option 2**
- We keep things as they are now, but we move `onChange(custom_object_record.id);` in `handleChange` as `onChange(selectionValue as string);`
  - With the current implementation, on the first render we get a `value` -> we call `fetchSelectedOption` -> we emit an `onChange` with the same value. We can skip this unneeded step

**Option 3**
- We change this `useEffect` to have `[value]` as the dependency array, so it will call `fetchSelectedOption` every time the `value` changes
- We change `handleChange` to emit only `onChange` when an option is selected, then when the component re-renders it will call `fetchSelectedOption` to fetch the option
- We remove the `onChange` call from `fetchSelectedOption`

Option 3 is the cleanest code-wise, we just react with an effect to changes of the `value`, but it is not efficient, because when the user selects an option the component will re-render, we will call the API, and the component will re-render again, so I would not go for that one.

Option 1 is the most efficient because we avoid an extra HTTP call. Option 2 can be ok as well, especially if we are not confident that we always have the selected option in the `options` array

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  const onFocus = () => {
```
Fredx87, line 191: Can we inline this instead of creating a new function?

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,249 @@
...
+      <input type="hidden" name={name} value={selectedOption?.value} />
+      <Combobox
+        inputProps={{ required }}
+        startIcon={<SearchIcon />}
+        validation={error ? "error" : undefined}
+        inputValue={inputValue}
+        selectionValue={selectedOption?.value}
+        onFocus={onFocus}
+        onChange={debounceHandleChange}
+        renderValue={({ selection }) =>
```
Fredx87, line 213: When the form has an error and the page is reloaded I don't see the selected record name in the field. I think we should render `selectedOption?.name` and not `selection?.label`, because in this case `selection` is null

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,216 @@
...
+          </Option>
+        )}
+        {isLoadingOptions && (
+          <Option
+            isDisabled
+            key={loadingOption.id}
+            value={loadingOption.name}
+          />
+        )}
+        {!isLoadingOptions && !isFirstLoad && options.length === 0 && (
```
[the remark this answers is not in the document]
Fredx87: Uhm, I am not sure I understand when we want to avoid showing "No matches found".

With the current implementation, the first time I type a letter I don't see the message but if I remove all the characters then I'll see "No matches found". Also if I select an item, then I search for a random string and I press enter, it will select the empty option. The next time I interact with the lookup field I'll see  "No matches found" as soon as I type the first character.

In general, I don't think looking for `isFirstLoad` makes much sense, but maybe I am missing something. I would rather check for `inputValue?.length > 0` so we don't show "No matches found" when the field is empty

src/modules/new-request-form/translations/en-us.yml
```diff
@@ -6,6 +6,7 @@
 # If you are building your own theme, you can remove this file and just load the translation JSON files, or provide
 # your translations with a different method.
 #
+
```
Fredx87, line 9: We can revert this change that just adds an empty line

templates/new_request_page.hbs
```diff
@@ -47,6 +46,5 @@
       requestPath: {{json (page_path 'request' id=answer_bot.request_id)}}
     },
   };
-
   renderNewRequestForm(settings, props, container);
-</script>
+</script>
```
Fredx87, line 50: There are still some changes in the formatting here (some removed empty lines)

Fredx87 (commented):
I think we still have some things to address, I left some comments

src/modules/new-request-form/fields/LookupField.tsx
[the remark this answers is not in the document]
gosiexon-zen: Ok, changed

src/modules/new-request-form/fields/LookupField.tsx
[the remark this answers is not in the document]
gosiexon-zen: Ok, changed

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+const EMPTY_OPTION = {
+  value: "",
+  name: "-",
+};
+
+interface LookupFieldProps {
+  field: Field;
+  userId: number;
+  organizationId: string | boolean | string[] | null | undefined;
```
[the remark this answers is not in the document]
gosiexon-zen: yes, the field value shows that it could be `Field.value?: string | boolean | string[] | undefined` but maybe we can ignore it.

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,249 @@
...
+  );
+
+  const debounceHandleChange = useMemo(() => debounce(handleChange, 300), []);
+
+  useEffect(() => {
+    return () => debounceHandleChange.cancel();
+  }, [debounceHandleChange]);
+
+  useEffect(() => {
+    if (value && !options.find((option) => option.value === value)) {
```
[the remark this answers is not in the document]
gosiexon-zen: that's right, changed

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: maybe it was linter suggestion or some leftovers from other approach 🤔 I removed it

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Yes we can, changed

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Changed

src/modules/new-request-form/translations/en-us.yml
gosiexon-zen: Changed

templates/new_request_page.hbs
gosiexon-zen: Removed these changes

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -301,6 +308,16 @@ export function NewRequestForm({
                   onChange={(value) => handleChange(field, value)}
                 />
               );
+            case "lookup":
+              return (
+                <LookupField
+                  key={field.name}
+                  field={field}
+                  userId={userId}
+                  organizationId={organizationField && organizationField?.value}
```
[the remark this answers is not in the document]
gosiexon-zen: changed, just waiting to pass here id from organizations from curlybars

Fredx87 (commented):
I think the implementation is ok, let's see if we can use the new `organizations` from Curlybars. Otherwise, we can use the current implementation for testing.

Before approving the PR it would be nice to have that change for `organizations` and also squash the changes to a single commit with the message `feat: added support for lookup fields`

src/modules/new-request-form/data-types/Organization.ts
```diff
@@ -0,0 +1,5 @@
+export interface Organization {
+  id: number;
+  name: string;
+  default: true;
```
Fredx87, line 4: We don't have a default property, right? In this case I would remove it from the type

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,245 @@
...
+        `/api/v2/custom_objects/${customObjectKey}/records/${selectionValue}`
+      );
+      const { custom_object_record } = await res.json();
+      const newSelectedOption = {
+        name: custom_object_record.name,
+        value: custom_object_record.id,
+      };
+      setSelectedOption(newSelectedOption);
+      setInputValue(custom_object_record.name);
+      onChange(custom_object_record.id);
```
Fredx87, line 90: Do we need to call `onChange` here? Since we call this function only from the initial `useEffect` to fetch the initial option, I don't think it is needed.

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+  useEffect(() => {
+    return () => debounceHandleChange.cancel();
+  }, [debounceHandleChange]);
+
+  useEffect(() => {
+    if (value) {
+      fetchSelectedOption(value as string);
+    }
+  }, []);
```
Fredx87, line 169: I would add a comment here explaining why we set the dependencies array to `[]`, to avoid the possibility of somebody changing it in the future

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,245 @@
...
+  const onFocus = () => {
+    setInputValue("");
+    selectedOption && setOptions([selectedOption]);
+  };
+
+  const handleKeyDown = useCallback((event) => {
+    if (event.key === "Enter") {
+      event.preventDefault();
+    }
+  }, []);
```
Fredx87, line 180: Why do we need this? I don't think it is a good idea to override the default key bindings

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,245 @@
...
+        placeholder={t(
+          "new-request-form.lookup-field.placeholder",
+          `Search ${label.toLowerCase()}`
+        )}
+        onFocus={onFocus}
+        onChange={debounceHandleChange}
+        renderValue={() =>
+          selectedOption ? selectedOption?.name : EMPTY_OPTION.name
+        }
+        style={{ marginTop: "8px" }}
```
Fredx87, line 208: We shouldn't override Garden styles. Especially with a static value of `8px`. The margin-top is missing just because we have the hidden input between the `Label` and the `Combobox`. Let's move the hidden input after the combobox

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,245 @@
...
+        onKeyDown={handleKeyDown}
+      >
+        {!required && selectedOption?.name !== EMPTY_OPTION.name && (
```
Fredx87, line 211: I would keep the empty option even if the field is required. We have the same behaviour in the Tagger component

src/modules/new-request-form/data-types/Organization.ts
gosiexon-zen: Right, we changed that. I will remove it

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Right, now we don't need this so I'm removing it.

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: ok, I added comment

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Because, when we hit enter in combobox we are submitting the whole form 🤷‍♀️ Not sure why it's only in this combobox, I've checked other fields also.

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Ok, I moved hidden input after the combobox and removed the style.

src/modules/new-request-form/fields/LookupField.tsx
Fredx87: I think it is the normal behavior of form inputs. If you press enter on any field (like subject, custom text input, custom date, cc field, ...) it will submit the form as well. The only fields where it is not happening are the WYSIWYG because the user needs to insert new lines and the other dropdowns because they don't have an input inside (we pass `isEditable={false}` as a prop).

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,243 @@
...
+      <Combobox
+        inputProps={{ required }}
+        startIcon={<SearchIcon />}
+        validation={error ? "error" : undefined}
+        inputValue={inputValue}
+        selectionValue={selectedOption?.value}
+        isAutocomplete
+        placeholder={t(
+          "new-request-form.lookup-field.placeholder",
+          `Search ${label.toLowerCase()}`
```
Fredx87, line 200: This is not the correct way to interpolate a string in the translation, we need pass the value as options as we are doing [here](https://github.com/zendesk/copenhagen_theme/blob/d36cd9f8dde4eea5814b61952a3c4de2426fcdaf/src/modules/new-request-form/fields/cc-field/CcField.tsx#L104). Reference: https://www.i18next.com/translation-function/interpolation

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,243 @@
...
+                name,
+                value: id,
+              })
+            )
+          );
+        } else {
+          setOptions([]);
+        }
+      } catch (error) {
+        return error;
```
Fredx87, line 123: It doesn't make much sense to return the error, nobody is using that. Maybe it is better to log it to the console

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,243 @@
...
+      } finally {
+        setIsLoadingOptions(false);
+      }
+      return;
```
Fredx87, line 127: What is this return for? I don't think we need it

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,243 @@
...
+      if (organizationId !== null)
+        searchParams.set("organization_id", organizationId);
+      setIsLoadingOptions(true);
+      try {
+        const response = await fetch(
+          `/api/v2/custom_objects/${customObjectKey}/records/autocomplete?${searchParams.toString()}`
+        );
+
+        const data = await response.json();
+        if (data !== undefined) {
```
Fredx87, line 110: Is there any case when the data can be undefined? I don't think it is possible but we would probably need to check if `response.ok` is true instead

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,243 @@
...
+      const res = await fetch(
+        `/api/v2/custom_objects/${customObjectKey}/records/${selectionValue}`
+      );
+      const { custom_object_record } = await res.json();
+      const newSelectedOption = {
+        name: custom_object_record.name,
+        value: custom_object_record.id,
+      };
+      setSelectedOption(newSelectedOption);
+      setInputValue(custom_object_record.name);
```
Fredx87, line 89: I think it is better to handle possible errors here:
- wrapping everything in a try catch
- checking if the response is ok before parsing it

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: ok, I changed that

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: There is warning that not all the statements return a value

gosiexon-zen: There is warning that not all the statements return a value

gosiexon-zen: Ok, there was before, maybe after splitting handle change it is not needed anymore

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: ok I changed to check `response.ok`

Fredx87 (approved):

PR #508, opened 2024-08-02
fix: update dompurify to 3.0.11

BrunoBFerreira:
## Description

This PR creates a fix to force the automatic creation of the release for version `4.0.5`.

The commit is empty since the last change did not generate the release, so the commit was just added as a means to trigger it. The commit message and PR title stays the same, since the change we want to release is the library update. 

## Checklist


Fredx87 (commented):
I would set the commit message to `fix: update dompurify to 3.0.11`, because it what ends up in the changelog, and just write a note in the PR description saying that the commit is empty because the release wasn't triggered by the previous merge

BrunoBFerreira:
> I would set the commit message to `fix: update dompurify to 3.0.11`, because it what ends up in the changelog, and just write a note in the PR description saying that the commit is empty because the release wasn't triggered by the previous merge

Should be updated @Fredx87.

Fredx87 (approved):

PR #516, opened 2024-08-16
fix: fixed styles with dark background and light brand color

Fredx87:
## Description

This PR improves the styles of the Garden components when the theme has a dark background color and a light brand color.

Previously in this scenario, the form was mostly unusable with unreadable text. With this fix the situation is better but not perfect:
- the Garden components now have a dark background, while other inputs and the WYSIWYG editor are still using a white background
- the description and error messages for fields are still difficult to read
- the icons in the dropdowns are barely visible

I think this is the best we can do for now given that Garden will have a proper dark mode support only in v9.

### Before

Text in the dropdowns

Screenshot 2024-08-16 at 12 52 43

Submit button text color

Screenshot 2024-08-16 at 12 53 07

Datepicker

Screenshot 2024-08-16 at 12 53 15

### After

Text in the dropdowns

Screenshot 2024-08-16 at 12 55 01

Submit button text color

Screenshot 2024-08-16 at 12 55 13

Datepicker

Screenshot 2024-08-16 at 12 55 29

## Checklist


melzreal (approved):
LGTM, I guess when the actual dark mode is introduced the override for this colors is ignored since its a separate theme entirely? thats as far as I understand it but not sure you'd like to grab an 👌 from someone with better understanding around the garden components

Fredx87:
> LGTM, I guess when the actual dark mode is introduced the override for this colors is ignored since its a separate theme entirely? thats as far as I understand it but not sure you'd like to grab an 👌 from someone with better understanding around the garden components

Honestly, I didn't have the time to check how dark mode will be implemented in v9. In general migrating to v9 will be a bigger task and we will probably need to revisit some things, but at least this is a first step to improve things a bit when people are customizing colors in this way

PR #519, opened 2024-08-20
fix: update wysiwyg package to 0.0.5

melzreal:
- Fixes translation issue with wysiwyg image upload not showing translated string

## Description

- Package version was outdated at 0.0.2, latest wysiwyg version is 0.0.5 and includes bug fix for the translated string which was missing

[JIRA] https://zendesk.atlassian.net/jira/software/c/projects/GG/boards/1258?selectedIssue=GG-3761

## Screenshots

Before (on PT instead of EN)

image

After

image

## Checklist


BrunoBFerreira (approved):

Fredx87 (commented):
I don't think that this solution that uses `resolutions` in the `package.json` is easy to work with. Every time we update the version in `dependencies` we would need to update the relative item in `resolutions`, and we will need to do it for every package from the `@zendesk` org.

I tested the following solution and it seems to work.

Create a `.npmrc` file with the following content:
```
@zendesk:registry=https://registry.npmjs.org/
```

Even if we are using one, it should respect the `.npmrc` file. We can then remove the `.yarnrc` file. and the field in `resolutions`. When I run `yarn install` locally, yarn is going to fetch the package correctly from `https://registry.yarnpkg.com/@zendesk/help-center-wysiwyg/`

package.json
```diff
@@ -105,5 +105,6 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  }
-}
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
melzreal, line 109: corepack keeps adding this line, not sure if its one to keep or remove?

Fredx87 (approved):

PR #523, opened 2024-09-16
fix: preload options to lookup field

gosiexon-zen:
## Description

[Jira issue](https://zendesk.atlassian.net/browse/GG-3824?atlOrigin=eyJpIjoiNTJlMjcxYzUxZTNlNDVjMDljYWE3ZGYxOGRiYjFlODkiLCJwIjoiaiJ9)
When the combobox is focused and the input text is empty, the user will see a list of options retrieved from the Autocomplete API. The list is limited to 100 options maximum. The selected option is shown as the second option on the list after empty option.
The search icon is removed from the field.

## Screenshots
Screenshot 2024-09-16 at 17 27 29

## Checklist


src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -114,14 +113,20 @@ export function LookupField({
...
+            ({ name, id }: { name: string; id: string }) => ({
+              name,
+              value: id,
+            })
           );
+          if (selectedOption) {
+            fetchedOptions = fetchedOptions.filter(
+              (option: FieldOption) => option.value !== selectedOption.value
+            );
+            fetchedOptions = [selectedOption, ...fetchedOptions];
```
BrunoBFerreira, line 126: Is this to guarantee that the selected option appears at the top?

gosiexon-zen: I think so 😬 just manually setting the order

luis-almeida: Is this the preferred experience for when an option is already selected?
I was going through the JIRA ticket and the slack thread but couldn't see it mentioned there.

gosiexon-zen: It is the same experience as in Lookup Field in Support and we want to make this fields almost the same.

luis-almeida: Got it! 👍🏼

BrunoBFerreira (approved):

PR #524, opened 2024-09-18
Lookup fields implementation

gosiexon-zen:
## Description
This code add support for Lookup fields for custom objects in the Copenhagen Theme v4.
[Jira: Lookup fields in the request form or templating API v4](https://zendesk.atlassian.net/browse/GG-3449)
## Screenshots

https://github.com/user-attachments/assets/8d04e0ea-a095-448c-bbd9-47ff0e848f55

## Checklist


luis-almeida (approved):

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -0,0 +1,252 @@
...
+          const { custom_object_record } = await res.json();
+          const newSelectedOption = {
+            name: custom_object_record.name,
+            value: custom_object_record.id,
+          };
+          setSelectedOption(newSelectedOption);
+          setInputValue(custom_object_record.name);
+        }
+      } catch (error) {
+        console.log(error);
```
BrunoBFerreira, line 93: Do we usually leave `console.log` calls in the code for frontend?

kubraokcu (approved):

src/modules/new-request-form/fields/LookupField.tsx
gosiexon-zen: Usually linters don't allow this, here linter tells me that every path needs to return value so I need to do something with this error and we don't have any logging tool added. I will change it to console.error, I see there is other usage of this in Copenhagen Theme repo.

BrunoBFerreira: Thank you for checking this out. If `console.log` is the best approach for it I believe it is fine to keep. Just wanted to be sure how we usually do it.

BrunoBFerreira (approved):

PR #527, opened 2024-10-02
feat: validate CC tag when leaving the field

BrunoBFerreira:
## Description

When we updated for v4, we changed the behavior of the CC field slightly. Before, if we clicked outside the input box, we would validate the current string and turn it into a tag, when valid. Currently, we are not doing that and we are just saving the input.

Since this feature is useful for our users, we are re-adding this feature in this PR. 

JIRA: [GG-3804](https://zendesk.atlassian.net/browse/GG-3804)

## Screenshots

Need to convert this to a gif

https://github.com/user-attachments/assets/e14acd12-db72-40be-9346-f3142cfa9587

## Checklist

[GG-3804]: https://zendesk.atlassian.net/browse/GG-3804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/assets/new-request-form-bundle.js
+++ b/assets/new-request-form-bundle.js
@@ -781,6 +781,16 @@ function useTagsInputContainer({ tags, onTagsChange, inputValue, onInputValueCha
         onTagsChange([...tags, ...values]);
         setAnnouncement(i18n.addedTags([...values]));
     };
+    const handleInputOnBlur = (e) => {
+        const target = e.target;
+        const tag = target.value;
+        if (tag) {
+            if (!hasTag(tag)) {
+                addTag(tag);
+            }
+            onInputValueChange("");
+        }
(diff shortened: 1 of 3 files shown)


luis-almeida:
Was the previous behaviour to also keep the focus on the field even after the the `blur`?

BrunoBFerreira:
> Was the previous behaviour to also keep the focus on the field even after the the `blur`?

Nope. Need to fix that. Thank you @luis-almeida!

BrunoBFerreira:
I believe it should be solved @luis-almeida. Could you take another look? 🙏

luis-almeida (approved):

PR #530, opened 2024-10-04
fix: paragraph gets treated as comma

melzreal:
## Description
- In French keyboards, the browser is detecting the code for a paragraph to be a comma instead.
- To fix this, for the outlier of the comma,I'm using the check on key instead of code.

The remaining checks appear consistent across keyboards so I'm going with the minimum change possible

## Checklist


src/modules/new-request-form/fields/cc-field/useTagsInputContainer.ts
```diff
@@ -90,12 +90,14 @@ export function useTagsInputContainer({
     const target = e.target as HTMLInputElement;
     const tag = target.value;
 
+    console.log(e.key);
```
luis-almeida, line 93: Nit: debug leftover?

luis-almeida (commented):
Is it still draft?

src/modules/new-request-form/fields/cc-field/useTagsInputContainer.ts
melzreal: was still in draft! I made the keys consistent now

melzreal:
> Is it still draft?

it was when you checked x) I was just asking Gianluca if we were ok with the approach or wanted consistency, ended up going with consistency for the keys by referring to the Garden constants (cuz e.key for space its just an empty space so it looked a little weird)

luis-almeida (approved):

BrunoBFerreira (approved):

PR #535, opened 2024-10-24
Service catalog item page

gosiexon-zen:
## Description
Added view for service catalog single item. For now showing only Title and description and button for submmiting a request (without any action). Data is mocked for now as backend is still in development and we are merging into feature branch not master.

[Implement rendering the Service Catalog Item Name and description](https://zendesk.atlassian.net/browse/GG-3879)
## Screenshots
Collapsible description:
https://github.com/user-attachments/assets/2e672b5e-0bb3-4243-a879-d2358666ab0d

With button:
Screenshot 2024-10-29 at 14 53 12

For small screens:
Screenshot 2024-10-29 at 14 51 49

## Checklist

### DIFF
--- a/src/modules/service-catalog/ServiceCatalogItemPage.tsx
+++ b/src/modules/service-catalog/ServiceCatalogItemPage.tsx
@@ -1,0 +1,128 @@
+import { XXXL } from "@zendeskgarden/react-typography";
+import type { ServiceCatalogItem } from "./data-types/ServiceCatalogItem";
+import styled from "styled-components";
+import { Button } from "@zendeskgarden/react-buttons";
+import { useState } from "react";
+import ChevronUp from "@zendeskgarden/svg-icons/src/16/chevron-up-fill.svg";
+import ChevronDown from "@zendeskgarden/svg-icons/src/16/chevron-down-fill.svg";
+import { getColorV8 } from "@zendeskgarden/react-theming";
+import { useTranslation } from "react-i18next";
+
+const ItemTitle = styled(XXXL)`
+  font-weight: ${(props) => props.theme.fontWeights.semibold};
+`;
+
+const CollapsibleDescription = styled.div<{ expanded: boolean }>`
+  font-size: ${(props) => props.theme.fontSizes.md};
+  text-align: left;
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: ${(props) => (props.expanded ? "none" : 3)};
+  overflow: hidden;
+  margin-top: ${(props) => props.theme.space.md};
+  padding-right: ${(props) => props.theme.space.xl};
+
+  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
+    padding-right: 0;
+  }
+`;
+const Container = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  gap: ${(props) => props.theme.space.xxl};
+
+  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
+    flex-direction: column;
+  }
+`;
+
+const LeftColumn = styled.div`
+  flex: 2;
+  border-bottom: ${(props) => props.theme.borders.sm}
+    ${(props) => getColorV8("grey", 300, props.theme)};
+  padding-bottom: ${(props) => props.theme.space.lg};
+  margin-right: ${(props) => props.theme.space.xl};
+
+  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
+    margin-right: 0;
+  }
+`;
+
+const RightColumn = styled.div`
+  flex: 1;
+  margin-left: ${(props) => props.theme.space.xl};
+
+  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
+    position: sticky;
+    bottom: 0;
+    margin-left: 0;
+  }
+`;
+
+const ToggleButton = styled(Button)`
+  margin-top: ${(props) => props.theme.space.sm};
+  font-size: ${(props) => props.theme.fontSizes.md};
+  &:hover {
+    text-decoration: none;
+  }
+`;
+
+const ButtonWrapper = styled.div`
+  padding: ${(props) => props.theme.space.lg};
+  border: ${(props) => props.theme.borders.sm}
+    ${(props) => getColorV8("grey", 300, props.theme)};
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
+    position: sticky;
+    bottom: 0;
+    background: ${(props) => props.theme.colors.background};
+    padding: ${(props) => props.theme.space.md};
+    border: none;
+    border-top: ${(props) => props.theme.borders.sm}
+      ${(props) => getColorV8("grey", 300, props.theme)};
+  }
+`;
+interface ServiceCatalogItemPageProps {
+  serviceCatalogItem: ServiceCatalogItem;
+}
+
+export function ServiceCatalogItemPage({
+  serviceCatalogItem,
+}: ServiceCatalogItemPageProps) {
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const { t } = useTranslation();
+
+  const toggleDescription = () => {
+    setIsExpanded(!isExpanded);
+  };
+
+  return (
+    <Container>
+      <LeftColumn>
+        <ItemTitle tag="h1">{serviceCatalogItem.name}</ItemTitle>
+        <CollapsibleDescription expanded={isExpanded}>
+          {serviceCatalogItem.description}
+        </CollapsibleDescription>
+        <ToggleButton aria-hidden="true" isLink onClick={toggleDescription}>
+          {isExpanded
+            ? t("service-catalog.item.read-less", "Read less")
+            : t("service-catalog.item.read-more", "Read more")}
+          <Button.EndIcon>
+            {isExpanded ? <ChevronUp /> : <ChevronDown />}
+          </Button.EndIcon>
+        </ToggleButton>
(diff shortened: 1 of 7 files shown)


src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+          {isExpanded
+            ? t("service-catalog.item.read-less", "Read less")
+            : t("service-catalog.item.read-more", "Read more")}
+          <Button.EndIcon>
+            {isExpanded ? <ChevronUp /> : <ChevronDown />}
+          </Button.EndIcon>
+        </ToggleButton>
+      </LeftColumn>
+      <RightColumn>
+        {/* Placeholder for Request a service button */}
```
pfemo, line 88: Why not just add the request button as well in this PR and have it be a dull button?

gosiexon-zen: Actually we can add it 😄 we have another Jira for that, we usually split view for for different functionalities/parts and create tasks for that.

src/modules/service-catalog/translations/en-us.yml
```diff
@@ -40,3 +40,13 @@ parts:
       title: "Empty state go to homepage link"
       screenshot: "https://drive.google.com/file/d/1D_3VXzWSuj9f34PsFPa0b4uELz8-T9ll/view?usp=drive_link"
       value: "Go to the homepage"    
+  - translation:
+      key: "service-catalog.item.read-more"
+      title: "Button to show whole description of the service catalog item"
+      screenshot: "https://drive.google.com/file/d/1pXglNRRqa23LoA7CuZOiAKQAuwkRKFqN/view?usp=drive_link"
```
kubraokcu, line 46: Seems like there's an issue with the screenshot 😄
image

src/modules/service-catalog/ServiceCatalogItemPage.tsx
pfemo: Ah okay; totally up to you. I just wondered as it seemed like the only thing missing it might have made sense to just include it to reduce overhead of reviews, opening PRs etc. but I'll let you decide what level of separation you prefer 🙌

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+const Container = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  gap: 120px;
+
+  @media (max-width: 768px) {
+    flex-direction: column;
+  }
+`;
```
Fredx87, line 35: Are we sure we want this behavior for mobile? Just flipping the flex-direction and keeping the 120px gap? From the design, it seems that on mobile the button is a sticky footer, but I am not sure.

I also don't like we are setting the gap to an hardcoded 120px value, but if that is the value we don't have alternatives

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+  width: 100%;
+  margin-top: ${(props) => props.theme.space.md};
+`;
+const Container = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  gap: 120px;
+
+  @media (max-width: 768px) {
```
Fredx87, line 32: I would not hardcode the breakpoint value but read it from the theme (` p.theme.breakpoints.md`) or use the [mediaquery](https://garden.zendesk.com/components/utilities#mediaquery) utility

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+    flex-direction: column;
+  }
+`;
+
+const LeftColumn = styled.div`
+  flex: 2;
+  border-bottom: ${(props) => props.theme.borders.sm}
+    ${(props) => getColorV8("grey", 300, props.theme)};
+  padding-bottom: ${(props) => props.theme.space.lg};
```
Fredx87, line 41: Are we sure that we always want to show a bottom border even if no options are available? I think the grey line is more like a separator between the description and the options, but we don't have a design for a page without options. 

In any case these border and padding should not be set on the whole column but only on the first part, but it is ok to do it in the PR where we will introduce the options

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const CollapsibleDescription = styled.div<{ expanded: boolean }>`
+  font-size: ${(props) => props.theme.fontSizes.md};
+  text-align: left;
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: ${(props) => (props.expanded ? "none" : 3)};
+  overflow: hidden;
+  text-overflow: ellipsis;
+  width: 100%;
```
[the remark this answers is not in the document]
gosiexon-zen: yep, removed

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const { t } = useTranslation();
+
+  const toogleDescription = () => {
+    setIsExpanded(!isExpanded);
+  };
+
+  return (
+    <Container>
+      <LeftColumn>
+        <ItemTitle>{serviceCatalogItem.name}</ItemTitle>
```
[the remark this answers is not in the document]
gosiexon-zen: Ok, Changed

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const ToggleButton = styled(Button)`
+  margin-top: ${(props) => props.theme.space.sm};
+  color: ${(props) => getColorV8("blue", 600, props.theme)};
+  font-size: ${(props) => props.theme.fontSizes.md};
+  &:hover {
+    text-decoration: none;
+    color: ${(props) => getColorV8("blue", 600, props.theme)};
+  }
+`;
```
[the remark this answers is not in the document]
gosiexon-zen: That's true, I followed the design but Guide has it's own theme so I removed the color property here.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+  serviceCatalogItem: ServiceCatalogItem;
+}
+
+export function ServiceCatalogItemPage({
+  serviceCatalogItem,
+}: ServiceCatalogItemPageProps) {
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const { t } = useTranslation();
+
+  const toogleDescription = () => {
```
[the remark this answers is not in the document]
gosiexon-zen: thanks, changed

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,92 @@
...
+  };
+
+  return (
+    <Container>
+      <LeftColumn>
+        <ItemTitle>{serviceCatalogItem.name}</ItemTitle>
+        <CollapsibleDescription expanded={isExpanded}>
+          {serviceCatalogItem.description}
+        </CollapsibleDescription>
+        <ToggleButton isLink onClick={toogleDescription}>
```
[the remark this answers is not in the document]
gosiexon-zen: Right, screen reader reads all of the description also the hidden part. So I added aria-hidden

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: Ok I will  work more on that in the PR with form fields

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: I removed the 120px value and changed it to `space.xl` and also added this value to the right and left margin property of the columns.

gosiexon-zen: I changed a bit style for mobile but I will work more on that. I see that we have some global css properties for footer and container and this sticky button doesn't look like the one on the design.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: Right, I added it and adjusted a styles a bit but it still needs some adjustments for mobile view.

src/modules/service-catalog/translations/en-us.yml
gosiexon-zen: Thanks, I fixed the screenshots.

kubraokcu (approved):
👍🏼🌐

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+  border: ${(props) => props.theme.borders.sm}
+    ${(props) => getColorV8("grey", 300, props.theme)};
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
+    position: sticky;
+    bottom: 0;
+    background: white; /* Ensure the button doesn't overlap content */
```
Fredx87, line 82: We should use `props.theme.colors.background`, the background could be different than white

Fredx87 (commented):
Added a comment. For the mobile styles, it is ok to revisit them in the PR with the ticket fields

Fredx87 (approved):

PR #537, opened 2024-10-31
feat: added form fields to service item

gosiexon-zen:
## Description
Display ticket fields associated with service catalog item for request.
Jira: [Implement Service Catalog Item ticket fields in the form](https://zendesk.atlassian.net/browse/GG-3880)

## Screenshots
Screenshot 2024-10-31 at 14 39 33

**Mobile view:** 

Screenshot 2024-10-31 at 14 44 07

## Checklist


src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -77,44 +96,78 @@ const ButtonWrapper = styled.div`
   align-items: center;
 
   @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    position: sticky;
+    position: fixed;
     bottom: 0;
+    left: 0;
+    right: 0;
+    width: 100vw;
```
Fredx87, line 103: Honestly I wouldn't use `position: fixed` here, otherwise the button is going to cover the footer when the page is scrolled down to the bottom. It is better to use `sticky` which is already set on the parent. 
It is a bit difficult to set the width to be 100% of the page without position fixed, but let's check from design if it is really needed.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -77,44 +96,78 @@ const ButtonWrapper = styled.div`
...
+    box-sizing: border-box;
```
Fredx87, line 104: Why do we need this?

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -77,44 +96,78 @@ const ButtonWrapper = styled.div`
...
-          <Button.EndIcon>
-            {isExpanded ? <ChevronUp /> : <ChevronDown />}
-          </Button.EndIcon>
-        </ToggleButton>
+        <DescriptionWrapper>
+          <ItemTitle tag="h1">{serviceCatalogItem.name}</ItemTitle>
+          <CollapsibleDescription expanded={isExpanded}>
+            {serviceCatalogItem.description}
+          </CollapsibleDescription>
+          <ToggleButton aria-hidden="true" isLink onClick={toggleDescription}>
```
Fredx87, line 151: As discussed, we need to remove this `aria-hidden` attribute since it is not compliant

src/modules/service-catalog/components/service-catalog-item/ItemRequestForm.tsx
```diff
@@ -0,0 +1,103 @@
...
+              );
+            default:
+              return <></>;
+          }
+        })}
+    </Form>
+  );
+};
+
+export default ItemRequestForm;
```
Fredx87, line 103: We usually don't use `export default`, and I would just export the component directly for consistency

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -0,0 +1,106 @@
...
+  } else {
+    return {
+      ...field,
+      name: field.title,
+      label: field.title,
+      options: [],
+      error: null,
+    };
+  }
+};
```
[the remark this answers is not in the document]
gosiexon-zen: Thanks! I applied the suggestions

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -0,0 +1,106 @@
...
+          error
+        );
+      }
+    }
+    fetchItemForm().then((ticket_field_ids) => {
+      if (ticket_field_ids) {
+        getTicketFields(ticket_field_ids);
+      }
+    });
+  }, [serviceCatalogItem, getTicketFields]);
```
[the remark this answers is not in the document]
gosiexon-zen: Ok I changed it to fetchTicketFields function to simplify the code

src/modules/ticket-fields/translations/en-us.yml
```diff
@@ -0,0 +1,32 @@
...
+      screenshot: "https://drive.google.com/file/d/1_TeqU5hihEJb03q6adVI0-nuklYask1d/view?usp=drive_link"
+      value: "Loading items..."
+  - translation:
+      key: "ticket-fields.lookup-field.no-matches-found"
+      title: "Message shown in combobox when searched option is not found"
+      screenshot: "https://drive.google.com/file/d/1zwRLhFj0I0N0JV07O7wpoEZO5tBBK8JF/view?usp=drive_link"
+      value: "No matches found"
+  - translation:
+      key: "ticket-fields.lookup-field.placeholder"
+      title: "Placeholder shown in combobox, {{label}} will be replaced by field label"
```
kubraokcu, line 30:
```suggestion
      title: "Placeholder shown in combobox, {{label}} will be replaced by field label. Alternative English for languages with grammar/declension issues: [Search label {{label}}]."
```

src/modules/service-catalog/components/service-catalog-item/ItemRequestForm.tsx
```diff
@@ -0,0 +1,103 @@
...
+  userRole,
+  userId,
+  defaultOrganizationId,
+  handleChange,
+}: ItemRequestFormProps) => {
+  return (
+    <Form>
+      {requestFields !== undefined &&
+        requestFields.map((field) => {
+          switch (field.type) {
```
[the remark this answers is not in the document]
gosiexon-zen: Added new module `ticket-fields`.

kubraokcu (approved):
👍🏼🌐

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: Changed to `sticky`

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: I removed that, it was I think to include padding in the width but it works ok now.

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -0,0 +1,108 @@
...
+          ticketField.type === "lookup" &&
+          isAssociatedLookupField(ticketField)
+        ) {
+          return null;
+        } else {
+          return formatField(ticketField);
+        }
+      })
+      .filter(Boolean);
+    return requestFields;
```
Fredx87, line 70: We need to take into account a couple of more things:
- the `ticket_field_ids` are ordered and we need to respect the same order when we show them. With this implementation, we use the order returned by the ticket fields API
- we also need to show only fields that have `editable_in_portal: true`. In theory, we should not have fields with this property set to false, but better add this check for safety
- we are returning fields that we don't show later in the form (subject, description, assignee, group, ...). It is not a big issue since we are not showing them, but better to not keep them in the state as well. Most of them will be filtered out if we keep only `editable_in_portal: true`, at that point probably we will need to filter out just subject and description

src/modules/ticket-fields/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  field: Field;
+  baseLocale: string;
+  hasAtMentions: boolean;
+  userRole: string;
+  userId: number;
+  defaultOrganizationId: string | null;
+  organizationField?: Field | null;
+  brandId: number;
+  dueDateField?: Field;
+  handleDueDateChange?: (value: string) => void;
```
Fredx87, line 22: If we are going to support the "Type" field, we will need to support the due date field as well and not mark these props as optional.
I would say that this use case has a lower priority so I'll create a Jira task for that and we can keep this caveat for now

src/modules/ticket-fields/translations/en-us.yml
```diff
@@ -0,0 +1,32 @@
...
+# This file is used for the internal Zendesk translation system, and it is generated from the extract-strings.mjs script.
+# It contains the English strings to be translated, which are used for generating the JSON files containing the translation strings
+# for each language.
+#
+# If you are building your own theme, you can remove this file and just load the translation JSON files, or provide
+# your translations with a different method.
+#
+title: "Copenhagen Theme Ticket Fields"
+packages:
+  - "ticket-fields"
```
Fredx87, line 11: I don't think it is a good idea to use `ticket-fields` as the package name and prefix for the keys, since these things are shared across Zendesk.

For `service-catalog` and `new-request-form` we did that, but we own these features and are not implemented anywhere else. Ticket fields are used everywhere, so maybe it is better to use something like `cph-theme-ticket-fields`.

Other notes:
- we are missing the `new-request-form.credit-card-digits-hint` translation. We need to move it here and use the new key in the Credit Card component
- we still need to handle translations properly. the `new-request-form` module now needs to load its own translations and the `ticket-fields` translations. Same for the `service-catalog` module. We can do it later in another PR, so we can move on with the implementation. I'll create a Jira for that
- We will need to deprecate the old strings in the `new-request-form` module. As above we can do it later

src/modules/ticket-fields/index.tsx
Fredx87: Reasoning again about ticket type and priority, there are a lot of things we need to take into account so I would not support them for now, but we can still keep them in the `ticket-fields` module.

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -14,16 +9,17 @@ import { useFormSubmit } from "./useFormSubmit";
...
 import { AnswerBotModal } from "./answer-bot-modal/AnswerBotModal";
 import { useTranslation } from "react-i18next";
 import { Paragraph } from "@zendeskgarden/react-typography";
-import { LookupField } from "./fields/LookupField";
 import type { Organization } from "./data-types/Organization";
+import type { Field } from "../ticket-fields/data-types/Field";
+import { TicketField } from "../ticket-fields";
+import { Input } from "../ticket-fields/fields/Input";
+import { DropDown } from "../ticket-fields/fields/DropDown";
+import { TextArea } from "../ticket-fields/fields/textarea/TextArea";
```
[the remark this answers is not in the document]
gosiexon-zen: Ok I changed it.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -7,6 +7,9 @@ import ChevronUp from "@zendeskgarden/svg-icons/src/16/chevron-up-fill.svg";
 import ChevronDown from "@zendeskgarden/svg-icons/src/16/chevron-down-fill.svg";
 import { getColorV8 } from "@zendeskgarden/react-theming";
 import { useTranslation } from "react-i18next";
+import type { Organization } from "../new-request-form/data-types/Organization";
```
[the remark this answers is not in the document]
gosiexon-zen: Ok, moved.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -75,26 +101,52 @@ const ButtonWrapper = styled.div`
   display: flex;
   flex-direction: column;
   align-items: center;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  width: 100%;
```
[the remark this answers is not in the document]
gosiexon-zen: Right we only need this properties for mobile view.

src/modules/service-catalog/components/service-catalog-item/ItemRequestForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+  hasAtMentions,
+  userRole,
+  userId,
+  brandId,
+  defaultOrganizationId,
+  handleChange,
+}: ItemRequestFormProps) => {
+  return (
+    <Form>
+      {requestFields !== undefined &&
```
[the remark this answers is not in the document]
gosiexon-zen: No we don't need, I removed it

src/modules/service-catalog/components/useItemFormFields.tsx
gosiexon-zen: Ok changed code to respect the order from the form and added `base_locale` to the API

src/modules/ticket-fields/index.tsx
gosiexon-zen: Ok so leaving it now as it is for now, later we can think which fields we will add.

src/modules/ticket-fields/translations/en-us.yml
gosiexon-zen: Ok so for now I've just changed package name to `cph-theme-ticket-fields`.
I see the task that you have created so I will take it later.

luis-almeida (approved):
The PR has been through thorough review and all comments have been addressed. I've gone through it again, tested it locally and I think it should be ready to go 👍🏼 

One thing I'd like to point out, is that as a team it would be really good if we could streamline and standardize our app structure patterns and naming conventions. I think there is some inconsistency within this code base but especially when compared to others that we own. Here's just a few examples:

- Some times we have an `index` file, some others we don't
- Some times we re-export/import from `index`, some others we go for the file
- Some times we have a `components` folder, some others we don't
- ...

I think this is just a consequence of us having different habits we're accustomed to since there isn't just "one right way" to do it. However, I think it's hurting us a bit especially when moving from one repo to another. It would definitely be nice if we could align.

PR #543, opened 2024-11-14
fix: add aria-label for removing attachment in new request form

gosiexon-zen:
## Description
Added aria-label with information how to remove the attachment while navigating with keyboard. 
This A11y issue was reported here: [GG-3796](https://zendesk.atlassian.net/browse/GG-3796)
Here is also explanation why we don't make the remove button accessible by tab navigation: https://zendesk.atlassian.net/browse/GG-3796?focusedCommentId=2055655

## Checklist

[GG-3796]: https://zendesk.atlassian.net/browse/GG-3796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kubraokcu (approved):
👍🏼🙌🏼🌐

Fredx87 (commented):
I added a couple of comments, and we also need to wrap the `FileListItem` components with the `FileList` component [here](https://github.com/zendesk/copenhagen_theme/blob/452cc6816baab566abdbb977f47ec11182d584fc/src/modules/new-request-form/fields/attachments/Attachments.tsx#L205) as shown in the Garden examples: https://66bcf1412f3b1af56ce3fee1--zendeskgarden.netlify.app/components/file-upload#files

jgorfine-zendesk (approved):
Hi @gosiexon-zen, thanks so much for making this change. I just tested it out in my local Guide instance, and I believe it's working as expected:

https://github.com/user-attachments/assets/ac0227d2-0524-42cd-bb7b-c9cc0c2988c6

Is there any chance we could update the "Remove file" button ([line 74 of FileListItem.tsx](https://github.com/zendesk/copenhagen_theme/blob/452cc6816baab566abdbb977f47ec11182d584fc/src/modules/new-request-form/fields/attachments/FileListItem.tsx#L74)) so it includes the file name, like the label for the group does? That way, if a screen reader user uploads multiple files, they'll be able to tell the "Remove" buttons apart, when navigating the page via form controls.

* Current behavior: "Remove file"
* Preferred behavior: "Remove file: {{fileName}}"

gosiexon-zen:
Hi @jgorfine-zendesk I applied your suggestions, but I think there should be `Stop uploading {{fileName}}` for the removing of the currently  uploading files and `Remove file: {{fileName}}` for those which are already uploaded. Let me know if it is ok.

Fredx87 (approved):

src/modules/new-request-form/fields/attachments/FileListItem.tsx
```diff
@@ -86,7 +99,11 @@ export function FileListItem({
             </FileNameWrapper>
             <Tooltip content={removeFileLabel}>
               <File.Delete
-                aria-label={removeFileLabel}
+                aria-label={t(
+                  "new-request-form.attachments.remove-file-arria-label",
+                  "Remove file: {{fileName}}",
+                  { fileName }
+                )}
```
jgorfine-zendesk, line 106: Can we please add `aria-describedby={undefined}` here? That way, the screen reader won't read out the "Remove file" text from the tooltip after it reads out the name of the button.

```suggestion
                aria-label={t(
                  "new-request-form.attachments.remove-file-arria-label",
                  "Remove file: {{fileName}}",
                  { fileName }
                )}
                aria-describedby={undefined}
```

* Current screen reader announcement: _Remove file: [file name], Remove file, button_
* Expected screen reader announcement: _Remove file: [file name], button_

(The word "button" comes from it being an HTML button element.)

src/modules/new-request-form/fields/attachments/FileListItem.tsx
```diff
@@ -55,13 +55,26 @@ export function FileListItem({
...
           <>
             <FileNameWrapper>{fileName}</FileNameWrapper>
             <Tooltip content={stopUploadLabel}>
               <File.Close
-                aria-label={stopUploadLabel}
+                aria-label={t(
+                  "new-request-form.attachments.stop-upload-aria-label",
+                  "Stop uploading {{fileName}}",
+                  { fileName }
+                )}
```
jgorfine-zendesk, line 77: Can we please add `aria-describedby={undefined}` to avoid a duplicate screen reader announcement here, too?

```suggestion
                aria-label={t(
                  "new-request-form.attachments.stop-upload-aria-label",
                  "Stop uploading {{fileName}}",
                  { fileName }
                )}
                aria-describedby={undefined}
```

* Current screen reader announcement: _Stop uploading [file name], Stop uploading, button_
* Expected screen reader announcement: _Stop uploading [file name], button_

(The word "button" comes from it being an HTML button element.)

jgorfine-zendesk (approved):
Hi @gosiexon-zen, I've requested a couple more small changes, but, after that, we should be good to go. 🙂 (I'm approving this PR anyway because the requested changes are more of a nice-to-have than a must-have.)

Here's the current screen reader experience on VoiceOver x Safari:

https://github.com/user-attachments/assets/ec81399c-d9b9-4acc-b45b-1c5aaa207573

src/modules/new-request-form/fields/attachments/FileListItem.tsx
gosiexon-zen: Ok I applied suggestion

src/modules/new-request-form/fields/attachments/FileListItem.tsx
gosiexon-zen: Ok I applied suggestion

Fredx87 (approved):

PR #544, opened 2024-11-15
Use target endpoints for Service Catalog list and item

gosiexon-zen:
## Description
Removed mock and call to Custom Objects API and instead use target endpoints created for Service Catalog:
```
/api/v2/help_center/service_catalog/items.json
/api/v2/help_center/service_catalog/items/{id}.json
```
For EAP we are reading the service item id from the URL as we are passing it by clicking the item on the list and redirecting. to the service item page. For GA we will read it from curlybars. 
Added component for description for service item to handle showing the "read more" button when the description is shorter.
[Jira issue GG-3951](https://zendesk.atlassian.net/browse/GG-3951?atlOrigin=eyJpIjoiNWE4NTUwMzljOWM5NDZiMTkzMzhkYTI3YWZkNmE2NzUiLCJwIjoiaiJ9) 

## Checklist

### DIFF
--- a/src/modules/service-catalog/ServiceCatalogItemPage.tsx
+++ b/src/modules/service-catalog/ServiceCatalogItemPage.tsx
@@ -1,34 +1,13 @@
-import { XXXL } from "@zendeskgarden/react-typography";
-import type { ServiceCatalogItem } from "./data-types/ServiceCatalogItem";
 import styled from "styled-components";
 import { Button } from "@zendeskgarden/react-buttons";
-import { useState } from "react";
-import ChevronUp from "@zendeskgarden/svg-icons/src/16/chevron-up-fill.svg";
-import ChevronDown from "@zendeskgarden/svg-icons/src/16/chevron-down-fill.svg";
 import { getColorV8 } from "@zendeskgarden/react-theming";
 import { useTranslation } from "react-i18next";
 import { useItemFormFields } from "./components/useItemFormFields";
 import { ItemRequestForm } from "./components/service-catalog-item/ItemRequestForm";
 import type { Organization } from "../ticket-fields";
+import { CollapsibleDescription } from "./components/service-catalog-item/CollapsibleDescription";
+import { useServiceCatalogItem } from "./useServiceCatalogItem";
 
-const ItemTitle = styled(XXXL)`
-  font-weight: ${(props) => props.theme.fontWeights.semibold};
-`;
-
-const CollapsibleDescription = styled.div<{ expanded: boolean }>`
-  font-size: ${(props) => props.theme.fontSizes.md};
-  text-align: left;
-  display: -webkit-box;
-  -webkit-box-orient: vertical;
-  -webkit-line-clamp: ${(props) => (props.expanded ? "none" : 3)};
-  overflow: hidden;
-  margin-top: ${(props) => props.theme.space.md};
-  padding-right: ${(props) => props.theme.space.xl};
-
-  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    padding-right: 0;
-  }
-`;
 const Container = styled.div`
   display: flex;
   flex-direction: row;
@@ -45,17 +24,6 @@ const LeftColumn = styled.div`
   display: flex;
   flex-direction: column;
   gap: ${(props) => props.theme.space.lg};
-  margin-right: ${(props) => props.theme.space.xl};
-
-  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    margin-right: 0;
-  }
-`;
-
-const DescriptionWrapper = styled.div`
-  border-bottom: ${(props) => props.theme.borders.sm}
-    ${(props) => getColorV8("grey", 300, props.theme)};
-  padding-bottom: ${(props) => props.theme.space.lg};
   margin-right: ${(props) => props.theme.space.xl};
 
   @media (max-width: ${(props) => props.theme.breakpoints.md}) {
@@ -86,14 +54,6 @@ const RightColumn = styled.div`
   }
 `;
 
-const ToggleButton = styled(Button)`
-  margin-top: ${(props) => props.theme.space.sm};
-  font-size: ${(props) => props.theme.fontSizes.md};
-  &:hover {
-    text-decoration: none;
-  }
-`;
-
 const ButtonWrapper = styled.div`
   padding: ${(props) => props.theme.space.lg};
   border: ${(props) => props.theme.borders.sm}
@@ -117,7 +77,7 @@ const ButtonWrapper = styled.div`
 `;
 
 export interface ServiceCatalogItemPageProps {
-  serviceCatalogItem: ServiceCatalogItem;
+  serviceCatalogItemId: number;
   baseLocale: string;
   hasAtMentions: boolean;
   userRole: string;
@@ -128,7 +88,7 @@ export interface ServiceCatalogItemPageProps {
 }
 
 export function ServiceCatalogItemPage({
-  serviceCatalogItem,
+  serviceCatalogItemId,
   baseLocale,
   hasAtMentions,
   userRole,
@@ -136,7 +96,7 @@ export function ServiceCatalogItemPage({
   userId,
   brandId,
 }: ServiceCatalogItemPageProps) {
-  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const serviceCatalogItem = useServiceCatalogItem(serviceCatalogItemId);
   const { requestFields, handleChange } = useItemFormFields(
     serviceCatalogItem,
     baseLocale
@@ -147,27 +107,13 @@ export function ServiceCatalogItemPage({
       ? organizations[0]?.id?.toString()
       : null;
 
-  const toggleDescription = () => {
-    setIsExpanded(!isExpanded);
-  };
-
-  return (
+  return serviceCatalogItem ? (
     <Container>
       <LeftColumn>
-        <DescriptionWrapper>
-          <ItemTitle tag="h1">{serviceCatalogItem.name}</ItemTitle>
-          <CollapsibleDescription expanded={isExpanded}>
-            {serviceCatalogItem.description}
-          </CollapsibleDescription>
-          <ToggleButton isLink onClick={toggleDescription}>
-            {isExpanded
-              ? t("service-catalog.item.read-less", "Read less")
-              : t("service-catalog.item.read-more", "Read more")}
-            <Button.EndIcon>
-              {isExpanded ? <ChevronUp /> : <ChevronDown />}
-            </Button.EndIcon>
-          </ToggleButton>
-        </DescriptionWrapper>
+        <CollapsibleDescription
+          title={serviceCatalogItem.name}
+          description={serviceCatalogItem.description}
+        />
         <FromFieldsWrapper>
           <ItemRequestForm
             requestFields={requestFields}
@@ -189,5 +135,5 @@ export function ServiceCatalogItemPage({
         </ButtonWrapper>
       </RightColumn>
     </Container>
-  );
+  ) : null;
 }
--- a/src/modules/service-catalog/ServiceCatalogList.tsx
+++ b/src/modules/service-catalog/ServiceCatalogList.tsx
@@ -21,6 +21,8 @@ const StyledGrid = styled(Grid)`
   padding: 0;
 `;
 
+const PAGE_SIZE = 16;
+
 type Meta = {
   before_cursor: string;
   after_cursor: string;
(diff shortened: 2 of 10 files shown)


src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -147,27 +109,30 @@ export function ServiceCatalogItemPage({
...
+  useEffect(() => {
+    async function fetchServiceCatalogItem() {
+      try {
+        const response = await fetch(
+          `/api/v2/help_center/service_catalog/items/${serviceCatalogItemId}`
+        );
+        const data = await response.json();
+        if (response.ok) {
+          setServiceCatalogItem(data.service_catalog_item);
+        }
```
sergioazevedo, line 121: Should we move `const data = ....` to the if block?
```suggestion
        if (response.ok) {
            const data = await response.json();
          setServiceCatalogItem(data.service_catalog_item);
        }
```

sergioazevedo, line 121: I also notice that we are not handling the case where the request is not successful. Shouldn't we have at least log the issue?

gosiexon-zen: Yes we can do this, I moved it.

gosiexon-zen: I don't see this pattern in other places but we definitely will introduce some proper error handling with logging.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -0,0 +1,98 @@
...
+        getComputedStyle(descriptionRef.current).lineHeight
+      );
+      const maxHeight = lineHeight * 3;
+      if (descriptionRef.current.scrollHeight > maxHeight) {
+        setShowToggleButton(true);
+      } else {
+        setShowToggleButton(false);
+      }
+    }
+  }, [description]);
```
luis-almeida, line 74: Would it be possible to do handle the toggle based on the `length` of the description instead of the height?

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -90,12 +90,14 @@ export function useItemFormFields(
 
   useEffect(() => {
     const fetchAndSetFields = async () => {
```
luis-almeida, line 92: Same here: could be nice to extract this into a `useTicketFields` hook.

templates/service_page.hbs
```diff
@@ -14,13 +14,18 @@
   const settings = {{json settings}};
   const container = document.getElementById("service-catalog-item");
 
+  // Extract the ID from the URL
+  const url = window.location.href;
+  const id = url.substring(url.lastIndexOf('/') + 1);
```
luis-almeida, line 19: `window.location.href` will return query params as well if there's ever some and that might break this logic. 
A more resilient way would be `window.location.pathname.split("/").pop()`.
We might as well consider exposing it via curlybars at some point.

templates/service_page.hbs
```diff
@@ -14,13 +14,18 @@
...
+
   const props = {
     baseLocale: {{json help_center.base_locale}},
     hasAtMentions: {{json help_center.at_mentions_enabled}},
     userRole: {{json user.role}},
     userId: {{json user.id}},
     brandId: {{json brand.id}},
     organizations: {{json user.organizations}},
```
luis-almeida, line 27: Nit as it's not related to this PR but we might not be using all these properties in the app.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -128,15 +89,16 @@ export interface ServiceCatalogItemPageProps {
...
   baseLocale,
   hasAtMentions,
   userRole,
   organizations,
   userId,
   brandId,
 }: ServiceCatalogItemPageProps) {
-  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const [serviceCatalogItem, setServiceCatalogItem] =
+    useState<ServiceCatalogItem>({} as ServiceCatalogItem);
```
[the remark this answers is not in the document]
gosiexon-zen: Yes, I changed it

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
gosiexon-zen: I changed it to length but I'm not sure if the `DESCRIPTION_LENGTH_THRESHOLD = 250` is ok

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -147,27 +109,30 @@ export function ServiceCatalogItemPage({
       ? organizations[0]?.id?.toString()
       : null;
 
-  const toggleDescription = () => {
-    setIsExpanded(!isExpanded);
-  };
+  useEffect(() => {
```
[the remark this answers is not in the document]
gosiexon-zen: Ok I moved it to hook, also we can later move all of the hooks to the hooks folder as we were talking

src/modules/service-catalog/components/useItemFormFields.tsx
gosiexon-zen: Isn't it too much hooks? Because in this case useItemFormFields will only call another hook and pass fields and onChange method

templates/service_page.hbs
gosiexon-zen: When we will be sure what type of fields we will have we can revise this.

templates/header.hbs
```diff
@@ -85,6 +85,7 @@
         {{/if}}
         <li class="item">{{link 'community' role="menuitem"}}</li>
         <li class="item">{{link 'new_request' role="menuitem" class='submit-a-request'}}</li>
+        <li class="item"><a href="/hc/en-us/services">Request a service</a></li>
```
gosiexon-zen, line 88: I added this as I saw we are missing a button in mobile menu

src/modules/service-catalog/useServiceCatalogItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+  useEffect(() => {
+    const fetchServiceCatalogItem = async () => {
+      try {
+        const response = await fetch(
+          `/api/v2/help_center/service_catalog/items/${serviceItemId}`
+        );
+        const data = await response.json();
+        setServiceCatalogItem(data.service_catalog_item);
+      } catch (error) {
+        console.error("Error fetching service catalog item:", error);
```
sergioazevedo, line 20: Unless I'm missing something here, looks like we are covering only the request happy path, ie when fetch returns a 200 response. What would happen if we get `404` or `503`? 

Shouldn't we check for `!response.ok or !response.ok` and provide some error handling?

@luis-almeida

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -46,32 +46,13 @@ export function ServiceCatalogList({
       try {
         const response = await fetch(
           currentCursor
-            ? `/api/v2/custom_objects/service_catalog_item/records?page[size]=16&${currentCursor}`
-            : `/api/v2/custom_objects/service_catalog_item/records?page[size]=16`
+            ? `/api/v2/help_center/service_catalog/items?page[size]=16&${currentCursor}`
+            : `/api/v2/help_center/service_catalog/items?page[size]=16`
```
sergioazevedo, line 50: I notice that the page[size] parameter is hardcoded to `16` is that intended?

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
luis-almeida: We can revise it with design later with design 👍🏼

src/modules/service-catalog/components/useItemFormFields.tsx
luis-almeida: It's away to abstract the data loading part so the component can be a bit more simple and just hold the rendering logic. But it's fine to leave it here as well.

src/modules/service-catalog/useServiceCatalogItem.tsx
luis-almeida: That's true I missed that. Can we check `response.ok` and throw an error so it falls in `catch`?

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
gosiexon-zen: But it might differ, letters have different width so it might be more characters but with letters that are thin or the other way around. So I don't think description length based on characters count is not good.

src/modules/service-catalog/ServiceCatalogList.tsx
gosiexon-zen: I changed it to some const variable so we can change it easily.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export const CollapsibleDescription = ({
+  title,
+  description,
+}: CollapsibleDescriptionProps) => {
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const [showToggleButton, setShowToggleButton] = useState<boolean>(false);
+  const { t } = useTranslation();
+  const DESCRIPTION_LENGTH_THRESHOLD = 250;
```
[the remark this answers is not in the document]
gosiexon-zen: Ok I moved it

src/modules/service-catalog/useServiceCatalogItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+      }
+    };
+    fetchServiceCatalogItem();
+  }, [serviceItemId]);
+  if (serviceCatalogItem) {
+    return serviceCatalogItem;
+  } else return undefined;
```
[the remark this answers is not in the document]
gosiexon-zen: yes we can 😄 I changed that

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
Fredx87: Yeah I know, this is why I think we should set a higher value, or eventually go back to the previous approach.

If we set a higher value, let's say for example 350 characters, it may happen that we will have like 5 rows and we don't show the collapsible button. I don't think it is a big deal if we don't always show exactly 3 lines. 
The important thing is that we find a good enough threshold to avoid showing the button when there is no more content to show, and this can be tricky especially considering that we will going to support other languages with different characters sets in the future.

On the other hand, the previous approach is more complex, but if it is the only reliable one maybe we should use it. wdyt @luis-almeida?

luis-almeida: We're not using a monospace font so I don't think it's an issue that the rows we end up with might differ a bit.
It's considerably simpler to do the check based on the based on the `description.length` so I'd go with that approach. We could also just have it in a one-liner without the need for state and effect as mentioned in the other comment 👍🏼

luis-almeida: We can try `270` 👍🏼  I think this is something we can always change entirely if it doesn't work.
But I think still worth it to try start with the simpler approach.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -147,47 +107,37 @@ export function ServiceCatalogItemPage({
...
-      </LeftColumn>
-      <RightColumn>
-        <ButtonWrapper>
-          <Button isPrimary size="large" isStretched>
-            {t("service-catalog.item.submit-button", "Submit request")}
-          </Button>
-        </ButtonWrapper>
-      </RightColumn>
+      {serviceCatalogItem && (
+        <>
```
Fredx87, line 113: Do we always need to render the container and then render the content with an empty fragment (`<>`) as a wrapper? Can we do `return serviceCatalogItem ? <Container>...</Container> : null` or do we need the container for some reason?

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const showToggleButton = description.length > DESCRIPTION_LENGTH_THRESHOLD;
+
+  const toggleDescription = () => {
+    setIsExpanded(!isExpanded);
+  };
+
+  return (
+    <DescriptionWrapper>
+      <ItemTitle tag="h1">{title}</ItemTitle>
+      <CollapsibleText expanded={isExpanded}>{description}</CollapsibleText>
```
Fredx87, line 71: Should we set `expanded={isExpanded || !showToggleButton}`, so we avoid clamping the height to 3 lines if we don't show the toggle button?

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: Yes we can do this, I changed that.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
gosiexon-zen: Yep, maybe it will help with some edge cases, I added this.

Fredx87 (approved):

PR #547, opened 2024-11-22
Post request for service catalog item

gosiexon-zen:
## Description
Added a functionality to submit a service catalog item request. Added validation to the form. Reorganize the structure of `ServiceCatalogItemPage.tsx` so the submit button is with the corresponding form. We have fixed values for subject and description in the request and it is in English only. This is probably only for EAP.

[GG-3862](https://zendesk.atlassian.net/browse/GG-3862?atlOrigin=eyJpIjoiMmQxZjUzYWQ5NmMwNGM2OGEyYjJjOTA5NWM2ZDM2NjAiLCJwIjoiaiJ9)
## Screenshots
Form with validation: 
Screenshot 2024-11-22 at 16 47 45

Requested service view:
Screenshot 2024-11-22 at 16 55 34

## Checklist

[GG-3796]: https://zendesk.atlassian.net/browse/GG-3796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[GG-3881]: https://zendesk.atlassian.net/browse/GG-3881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[GG-3862]: https://zendesk.atlassian.net/browse/GG-3862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/service-catalog/ServiceCatalogItemPage.tsx
+++ b/src/modules/service-catalog/ServiceCatalogItemPage.tsx
@@ -1,79 +1,22 @@
 import styled from "styled-components";
-import { Button } from "@zendeskgarden/react-buttons";
-import { getColorV8 } from "@zendeskgarden/react-theming";
-import { useTranslation } from "react-i18next";
 import { useItemFormFields } from "./components/useItemFormFields";
 import { ItemRequestForm } from "./components/service-catalog-item/ItemRequestForm";
 import type { Organization } from "../ticket-fields";
-import { CollapsibleDescription } from "./components/service-catalog-item/CollapsibleDescription";
 import { useServiceCatalogItem } from "./useServiceCatalogItem";
+import { submitServiceItemRequest } from "./submitServiceItemRequest";
+import type { ServiceRequestResponse } from "./data-types/ServiceRequestResponse";
+import { addFlashNotification } from "../shared";
+import { useTranslation } from "react-i18next";
+import {
+  useToast,
+  Close,
+  Notification,
+  Title,
+} from "@zendeskgarden/react-notifications";
 
 const Container = styled.div`
   display: flex;
-  flex-direction: row;
-  justify-content: space-between;
-  gap: ${(props) => props.theme.space.xxl};
-
-  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    flex-direction: column;
-  }
-`;
-
-const LeftColumn = styled.div`
-  flex: 2;
-  display: flex;
   flex-direction: column;
-  gap: ${(props) => props.theme.space.lg};
-  margin-right: ${(props) => props.theme.space.xl};
-
-  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    margin-right: 0;
-  }
-`;
-
-const FromFieldsWrapper = styled.div`
-  margin-right: ${(props) => props.theme.space.xl};
-
-  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    margin-right: 0;
-  }
-`;
-
-const RightColumn = styled.div`
-  flex: 1;
-  margin-left: ${(props) => props.theme.space.xl};
-
-  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    position: sticky;
-    bottom: 0;
-    margin-left: 0;
-    width: 100%;
-    display: flex;
-    flex-direction: column;
-    align-items: center;
-  }
-`;
-
-const ButtonWrapper = styled.div`
-  padding: ${(props) => props.theme.space.lg};
-  border: ${(props) => props.theme.borders.sm}
-    ${(props) => getColorV8("grey", 300, props.theme)};
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-
-  @media (max-width: ${(props) => props.theme.breakpoints.md}) {
-    position: sticky;
-    top: 0;
-    background: ${(props) => props.theme.colors.background};
-    padding: ${(props) => props.theme.space.lg};
-    border: none;
-    border-top: ${(props) => props.theme.borders.sm}
-      ${(props) => getColorV8("grey", 300, props.theme)};
-    width: 100vw;
-    left: 0;
-    right: 0;
-  }
 `;
 
 export interface ServiceCatalogItemPageProps {
@@ -97,11 +40,78 @@ export function ServiceCatalogItemPage({
   brandId,
 }: ServiceCatalogItemPageProps) {
   const serviceCatalogItem = useServiceCatalogItem(serviceCatalogItemId);
-  const { requestFields, handleChange } = useItemFormFields(
-    serviceCatalogItem,
-    baseLocale
-  );
+  const {
+    requestFields,
+    associatedLookupField,
+    setRequestFields,
+    handleChange,
+  } = useItemFormFields(serviceCatalogItem, baseLocale);
   const { t } = useTranslation();
+  const { addToast } = useToast();
+  const notifyError = () => {
+    addToast(({ close }) => (
+      <Notification type="error">
+        <Title>
+          {t(
+            "service-catalog.item.service-request-error-title",
+            "Service couldn't be submitted"
+          )}
+        </Title>
+        {t(
+          "service-catalog.item.service-request-error-message",
+          "Give it a moment and try it again"
+        )}
+
+        <Close
+          aria-label={t("new-request-form.close-label", "Close")}
+          onClick={close}
+        />
+      </Notification>
+    ));
+  };
+
+  const handleRequestSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!serviceCatalogItem || !associatedLookupField) {
+      return;
+    }
+    const response = await submitServiceItemRequest(
+      serviceCatalogItem,
+      requestFields,
+      associatedLookupField,
+      baseLocale
+    );
+    if (!response?.ok) {
+      if (response?.status === 422) {
+        const errorData: ServiceRequestResponse = await response.json();
+        const invalidFieldErrors = errorData.details.base;
+        const updatedFields = requestFields.map((field) => {
+          const errorField = invalidFieldErrors.find(
+            (errorField) => errorField.field_key === field.id
+          );
+          return errorField
+            ? { ...field, error: errorField.description }
+            : field;
+        });
+        setRequestFields(updatedFields);
+      } else {
+        notifyError();
+      }
+    } else if (response && response.ok) {
+      addFlashNotification({
+        type: "success",
+        message: t(
+          "service-catalog.item.service-request-submitted",
+          "Service request submitted"
+        ),
+      });
+      const data = await response?.json();
+      const redirectUrl = "/hc/requests/" + data.request.id;
+      window.location.href = redirectUrl;
+    }
+  };
+
   const defaultOrganizationId =
     organizations.length > 0 && organizations[0]?.id
       ? organizations[0]?.id?.toString()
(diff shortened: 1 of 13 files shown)


gosiexon-zen:
@Fredx87 we are nor showing and passing value for the priority field. We need to think if we should add this.

Fredx87:
> @Fredx87 we are nor showing and passing value for the priority field. We need to think if we should add this.

Yeah, the same should apply to the `Type` field (and the linked `due_at` field). These fields need some different handling, like sending their values at the top level instead of inside the `custom_fields` array and probably other stuff. For the EAP we can focus on custom fields and add support for those fields later

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -27,16 +27,24 @@ const formatField = (field: TicketField): Field => {
...
+    return true;
+  }
+  return false;
 };
 
 const fetchTicketFields = async (
   form_id: string,
-  baseLocale: string
+  baseLocale: string,
+  setAssociatedLookupField: (field: Field | null) => void
```
Fredx87, line 47: I think that passing a setter callback to a function called `fetchTicketFields` and then again to a function called `isAssociatedLookupField` is a bit convoluted and not really clean, since these functions are more getter. Probably it would be better to return both form fields and the lookup field from this function and call the setter outside

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -97,43 +32,65 @@ export function ServiceCatalogItemPage({
...
+    associatedLookupField,
     baseLocale
   );
-  const { t } = useTranslation();
+
+  const handleRequestSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    const response = await submitServiceItemRequest();
+    if (response && !response.ok) {
```
Fredx87, line 52: This wasn't specified in the Jira story, but we can get different errors from the API, like a 422 when there are user errors or other errors like a 500 for other reasons, and we should handle them differently.

We should check the status code, if it is a 422 we should leverage the already implemented logic, otherwise, we should probably just show an error notification

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -97,43 +32,65 @@ export function ServiceCatalogItemPage({
...
+      const errorData = await response.json();
```
Fredx87, line 53: Nit: it is probably better to create a type for the entire response and cast it here instead of just creating a type for `RequestError` and using it later

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -97,43 +32,65 @@ export function ServiceCatalogItemPage({
...
+      if (errorData.error === "RecordInvalid") {
+        const invalidFieldErros = errorData.details.base;
```
Fredx87, line 55: There is a typo here (`Erros`)

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -97,43 +32,65 @@ export function ServiceCatalogItemPage({
...
+          return errorField
+            ? { ...field, error: errorField.description }
+            : field;
+        });
+        setRequestFields(updatedFields);
+      }
+    } else if (response && response.ok) {
+      const data = await response?.json();
+      const redirectUrl = "/hc/requests/" + data.request.id;
+      window.location.href = redirectUrl;
```
Fredx87, line 69: We also need to show a success notification on the request page, as shown in the designs. We can use the `addFlashNotification` function as we do in the answer bot modal: https://github.com/zendesk/copenhagen_theme/blob/a9a12f06f07d5919f36e0b33505d52a789a927d9/src/modules/new-request-form/answer-bot-modal/AnswerBotModal.tsx#L107

src/modules/service-catalog/useServiceFormSubmit.tsx
```diff
@@ -0,0 +1,57 @@
+import type { Field } from "../ticket-fields";
+import type { ServiceCatalogItem } from "./data-types/ServiceCatalogItem";
+
+export function useServiceFormSubmit(
```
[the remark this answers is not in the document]
gosiexon-zen: Ok I changed that

src/modules/service-catalog/useServiceFormSubmit.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const currentUserRequest = await fetch("/api/v2/users/me.json");
+    const currentUser = await currentUserRequest.json();
+    const customFields = requestFields.map((field) => {
+      if (field.type !== "subject" && field.type !== "description") {
+        return {
+          id: field.id,
+          value: field.value,
+        };
+      } else return;
+    });
```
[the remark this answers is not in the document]
gosiexon-zen: ok, changed that

src/modules/service-catalog/data-types/TicketField.ts
```diff
@@ -5,6 +5,7 @@ export interface TicketField {
   title: string;
   value?: string | string[] | boolean;
   required: boolean;
+  error: string | null;
```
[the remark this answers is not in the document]
gosiexon-zen: Right, I confused it with Field type

src/modules/service-catalog/components/useItemFormFields.tsx
gosiexon-zen: Right, I changed that

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: ok changed

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: Ok I added type for response

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: Ok added. I'm not sure if we will need it also for error but I will need to ask for copy.

src/modules/service-catalog/submitServiceItemRequest.tsx
```diff
@@ -0,0 +1,50 @@
+import type { Field } from "../ticket-fields";
+import type { ServiceCatalogItem } from "./data-types/ServiceCatalogItem";
+
+export async function submitServiceItemRequest(
+  serviceCatalogItem: ServiceCatalogItem,
+  requestFields: Field[],
+  associatedLookupField: Field | null | undefined,
```
Fredx87, line 7: What happens it the field is null? we should prevent this scenario otherwise the submission will not work

src/modules/service-catalog/submitServiceItemRequest.tsx
```diff
@@ -0,0 +1,50 @@
...
+  baseLocale: string
+): Promise<Response | void> {
+  if (!serviceCatalogItem) {
+    return;
+  }
```
Fredx87, line 12: We don't need this if since the item cannot be null

src/modules/service-catalog/submitServiceItemRequest.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const currentUserRequest = await fetch("/api/v2/users/me.json");
+  const currentUser =
+    currentUserRequest.ok && (await currentUserRequest.json());
```
Fredx87, line 15: we should throw an error if we cannot get the csrf token

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -97,43 +34,69 @@ export function ServiceCatalogItemPage({
...
+    associatedLookupField,
+    setRequestFields,
+    handleChange,
+  } = useItemFormFields(serviceCatalogItem, baseLocale);
   const { t } = useTranslation();
+  const handleRequestSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    const response =
+      serviceCatalogItem &&
```
Fredx87, line 48: Isn't there a case where we will call this function when `serviceCatalogItem` is null, right? We are showing the button only if it is not null.

If this is the case I would just add a `if (!serviceCatalogItem) { return; }` at the beginning of the function because using this pattern when the response can be null makes the code difficult to read with a lot of ifs.

I also think we need to handle these edge cases better, if we cannot fetch the custom object we are just not showing the form, and if we don't find the lookup field we don't do anything. These scenarios are blocking because we cannot do anything without this data. We can eventually address this in a separate PR where we introduce error handling

src/modules/service-catalog/ServiceCatalogItemPage.tsx
Fredx87: Do we have an answer to this? IMO if the submission fails for other reasons we would need to show an error somewhere

src/modules/service-catalog/submitServiceItemRequest.tsx
gosiexon-zen: I added check for associatedLookupField

src/modules/service-catalog/submitServiceItemRequest.tsx
gosiexon-zen: Ok added

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: Ok changed to your suggestion. I think I already created some task for adding ErrorBoundary if not I will create and I will add it everywhere it is needed.

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: I've just go the copy so I added the correct one

src/modules/service-catalog/ServiceCatalogItemPage.tsx
```diff
@@ -97,43 +41,99 @@ export function ServiceCatalogItemPage({
...
-  );
+  const {
+    requestFields,
+    associatedLookupField,
+    setRequestFields,
+    handleChange,
+  } = useItemFormFields(serviceCatalogItem, baseLocale);
   const { t } = useTranslation();
+  const { addToast } = useToast();
+  const notifyError = useCallback(() => {
```
Fredx87, line 52: There shouldn't be the need to wrap this function in a `useCallback` hook since it is used only in the `handleRequestSubmit` function

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -87,14 +102,17 @@ export function useItemFormFields(
...
     const fetchAndSetFields = async () => {
       if (serviceCatalogItem && serviceCatalogItem.form_id) {
         try {
-          await fetchTicketFields(serviceCatalogItem.form_id, baseLocale).then(
-            (ticketFields) => setRequestFields(ticketFields)
-          );
+          const { requestFields, associatedLookupField } =
+            await fetchTicketFields(serviceCatalogItem.form_id, baseLocale);
+          setRequestFields(requestFields);
+          setAssociatedLookupField(associatedLookupField);
```
Fredx87, line 115: We should save the error state somewhere if don't find the lookup field. We can tackle this separately when we are going to handle the error state, the important thing is that we don't forget to do it

src/modules/service-catalog/ServiceCatalogItemPage.tsx
gosiexon-zen: ok I removed the useCallback

Fredx87 (approved):

PR #548, opened 2024-11-29
fix: prevent text overflow in Japanese

melzreal:
## Description
In Japanese, having the single character broken into another line, provides another word meaning. We want to make sure the text is only broken at points that still preserve the message meaning of "Last activity"

## Screenshots

Before

image

After

image

## Checklist

### DIFF
--- a/style.css
+++ b/style.css
@@ -3494,6 +3494,7 @@ ul {
   white-space: pre;
 }
 .request-details dt {
+  line-break: strict;
   color: lighten($text_color, 20%);
   width: 40%;
 }
--- a/styles/_request.scss
+++ b/styles/_request.scss
@@ -150,6 +150,7 @@
     }
 
     dt {
+      line-break: strict;
       color: $secondary-text-color;
       width: 40%;
     }


luis-almeida (commented):
The style.css is the compiled "dist" file. These changes should go in the [sass partial](https://github.com/zendesk/copenhagen_theme/blob/master/styles/_request.scss#L127).

When it comes to the fix:
- Have we gotten any feedback from localization that `text-wrap: nowrap;` is ok with other locales as well? I don't see why it wouldn't be but maybe worth checking
- What happens when the text is larger than the container? Will it stretch and break the layout?

Fredx87 (commented):
We shouldn't edit the `style.css` directly, as mentioned in the readme: https://github.com/zendesk/copenhagen_theme#stylesheet-and-javascript. In fact it is automatically hidden as a generated file. 

We need to edit the source .scss file instead: https://github.com/zendesk/copenhagen_theme/blob/6f47597a003d207c1aaab4a3123debf4bedddbd2/styles/_request.scss#L135.

Do we also need to change the padding of the `dd` element?

melzreal:
@luis-almeida 

> The style.css is the compiled "dist" file. These changes should go in the [sass partial](https://github.com/zendesk/copenhagen_theme/blob/master/styles/_request.scss#L127).

Thank you, have discarded the changes, amended the correct file and run yarn to re-generate the files

> When it comes to the fix:
> 
> * Have we gotten any feedback from localization that `text-wrap: nowrap;` is ok with other locales as well? I don't see why it wouldn't be but maybe worth checking
> * What happens when the text is larger than the container? Will it stretch and break the layout?

Good question, I had checked quite a few languages https://pandora-v2.zende.sk/resources/accounts/184478 - and just realised the overflow wrap breaks for Greek 🤦 
 but for it to look good in both languages only if the container gets bigger, because the text needs more space.. I'm going to have a go at it again

luis-almeida:
Would [`line-break: strict;`](https://developer.mozilla.org/en-US/docs/Web/CSS/line-break) address the issue? 
Would be nice if we could keep the text-wrap not to have to change the design.

melzreal:
> Would [`line-break: strict;`](https://developer.mozilla.org/en-US/docs/Web/CSS/line-break) address the issue? Would be nice if we could keep the text-wrap not to have to change the design.

No, I did try but it had no effect. Have changed this to slighly reduce the container padding for the text by 5px while increasing the container 3%, so the text now looks good on most viewpoints - should I put this to Guntis for approval? Or would you recommend a different approach?

image
image

melzreal:
> Did you try it along with the text-wrap: nowrap? It's meant to be used without setting any text-wrap, and if so it will keep the text-wrapping but respect asian languages.

I had expected the rule to not allow it to break at all even with nowrap disabled, hence finding it wasn't affecting the text - I didn't realise the difference

> If we can confirm with a Japanese speaker that the second image reads correctly, then IMO that would be the correct fix.

Confirmed with [Yuko ](https://zendesk.atlassian.net/browse/GG-3997?focusedCommentId=2065725)(native Japanese speaker) that its indeed fine to break at that point

luis-almeida (approved):

luis-almeida (approved):

PR #554, opened 2024-12-16
fix: ensure we handle null values on description

melzreal:
## Description

Currently the page crashes when a description does not exist, due to a length check on a null value.
Have changed the logic to handle nulls and abstracted the function to be able to test it in isolation.

JIRA: https://zendesk.atlassian.net/jira/software/c/projects/GG/boards/1258?selectedIssue=GG-4011

## Screenshots

Before, service without description:

image

Now
image

## Checklist

### DIFF
--- a/__mocks__/svgMock.js
+++ b/__mocks__/svgMock.js
@@ -1,0 +1,1 @@
+export default 'SvgMock';
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -5,6 +5,9 @@ const config = {
   transform: {
     "^.+.tsx?$": ["ts-jest", {}],
   },
+  moduleNameMapper: {
+    '\\.svg$': '<rootDir>/__mocks__/svgMock.js',
+  },
   setupFilesAfterEnv: ["@testing-library/jest-dom"],
 };
 
--- a/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
@@ -50,11 +50,14 @@ interface CollapsibleDescriptionProps {
   description: string;
 }
 
-const DESCRIPTION_LENGTH_THRESHOLD = 270;
-
-export const shouldShowToggleButton = (description: string | null | undefined): boolean => {
+export const shouldShowToggleButton = (
+  description: string | null | undefined
+): boolean => {
   const DESCRIPTION_LENGTH_THRESHOLD = 270;
-  return typeof description === 'string' && description.length > DESCRIPTION_LENGTH_THRESHOLD;
+  return (
+    typeof description === "string" &&
+    description.length > DESCRIPTION_LENGTH_THRESHOLD
+  );
 };
 
 export const CollapsibleDescription = ({
--- a/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
@@ -52,6 +52,11 @@ interface CollapsibleDescriptionProps {
 
 const DESCRIPTION_LENGTH_THRESHOLD = 270;
 
+export const shouldShowToggleButton = (description: string | null | undefined): boolean => {
+  const DESCRIPTION_LENGTH_THRESHOLD = 270;
+  return typeof description === 'string' && description.length > DESCRIPTION_LENGTH_THRESHOLD;
+};
+
 export const CollapsibleDescription = ({
   title,
   description,
@@ -59,8 +64,7 @@ export const CollapsibleDescription = ({
   const [isExpanded, setIsExpanded] = useState<boolean>(false);
   const { t } = useTranslation();
 
-  const showToggleButton = description.length > DESCRIPTION_LENGTH_THRESHOLD;
-
+  const showToggleButton = shouldShowToggleButton(description);
   const toggleDescription = () => {
     setIsExpanded(!isExpanded);
   };
--- a/assets/service-catalog-bundle.js
+++ b/assets/service-catalog-bundle.js
@@ -312,11 +312,14 @@ const ToggleButton = styled(Button) `
     text-decoration: none;
   }
 `;
-const DESCRIPTION_LENGTH_THRESHOLD = 270;
+const shouldShowToggleButton = (description) => {
+    const DESCRIPTION_LENGTH_THRESHOLD = 270;
+    return typeof description === 'string' && description.length > DESCRIPTION_LENGTH_THRESHOLD;
+};
 const CollapsibleDescription = ({ title, description, }) => {
     const [isExpanded, setIsExpanded] = reactExports.useState(false);
     const { t } = useTranslation();
-    const showToggleButton = description.length > DESCRIPTION_LENGTH_THRESHOLD;
+    const showToggleButton = shouldShowToggleButton(description);
     const toggleDescription = () => {
         setIsExpanded(!isExpanded);
     };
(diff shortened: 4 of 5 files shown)


src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -50,7 +50,15 @@ interface CollapsibleDescriptionProps {
   description: string;
 }
 
-const DESCRIPTION_LENGTH_THRESHOLD = 270;
+export const shouldShowToggleButton = (
+  description: string | null | undefined
+): boolean => {
+  const DESCRIPTION_LENGTH_THRESHOLD = 270;
+  return (
+    typeof description === "string" &&
```
gosiexon-zen, line 58: I don't think we need to check if type is string but just to check if it is not null. In `ServiceCatalogItem` type we assign type string to description. 
Other question is for product if we want to allow to save service catalog item with empty description? If yes so we need to change the type.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -59,8 +67,7 @@ export const CollapsibleDescription = ({
   const [isExpanded, setIsExpanded] = useState<boolean>(false);
   const { t } = useTranslation();
 
-  const showToggleButton = description.length > DESCRIPTION_LENGTH_THRESHOLD;
-
+  const showToggleButton = shouldShowToggleButton(description);
```
gosiexon-zen, line 70: We don't need additional function, something like this should be enough: 
`const showToggleButton = description != null && description.length > DESCRIPTION_LENGTH_THRESHOLD;`

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
BrunoBFerreira, line 58: (nit) Maybe we can check for the presence of the description here instead of using typeof. I believe when you do something like 
```typescript
if (description) {
...
}
```

It returns false for both `null` and `undefined`. No strong feelings here though, since we know the exact type of the description.

__mocks__/svgMock.js
```diff
@@ -0,0 +1 @@
+export default 'SvgMock';
```
BrunoBFerreira, line 1: Out of curiosity, why are we adding this?

BrunoBFerreira (commented):
Just a nitpick and a question. Thank you!

jest.config.mjs
```diff
@@ -5,6 +5,9 @@ const config = {
   transform: {
     "^.+.tsx?$": ["ts-jest", {}],
   },
+  moduleNameMapper: {
+    '\\.svg$': '<rootDir>/__mocks__/svgMock.js',
```
gosiexon-zen, line 9: Why we need mock for svg?

__mocks__/svgMock.js
melzreal: image

without it - the test blows up essentially, with this error

image

the __mocks__ folder is the default used by jest, when paired with the change to the jest config it fixes the issue with svgs breaking tests

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
melzreal: the typeof description being string was a suggestion from copilot, in that instead of checking for null by checking for the typeOf being string then we are more safe to call "length" on it because its a method you can only call on strings

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
melzreal: I extracted it so I could test it, otherwise it can't be tested in isolation

jest.config.mjs
melzreal: see https://github.com/zendesk/copenhagen_theme/pull/554/files#r1888290869

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
gosiexon-zen: We don't have tests in this repo, we usually write e2e tests for CPH theme in Theming Center

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
gosiexon-zen: I would change that on the API side so it will be a string or empty string as Luis suggested in the JIRA ticket.
We know that if it won't be null it will be string.

__mocks__/svgMock.js
BrunoBFerreira: Did we know why it started breaking? Was it due to a version update or similar?

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
melzreal: its not a dependency install, no new packages were added - just a mock, and this is the same we do for testing in other jest cases. As soon as we increase our test coverage - which I don't understand quite why we don't have more tests on this repo already - it becomes needed, so unsure why not add it now?

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
melzreal: if we change it on the backend its a larger change as it affects every single ticketfield, we cant limit it to service catalog items? Because every item would then need to have a default of empty string on the TicketField initialization, but thats not how we are treating other fields.

so instead of initializing the ticket fields as we currently do:

```

  def initialize(params = {})
    @id = params[:id]
    @type = params[:type]
    @title_in_portal = params[:title_in_portal]
    @description = params[:description]
...
  end
  
  ```
  
  we'd have to prefill to a string on each initialization:
  
    def initialize(params = {})
   ...
    @description = params.fetch(:description, "")
```

My problem with this approach is that it affects every single part of the codebase using ticketfields, instead of just our own isolated code - and I'm uncertain if there won't be any code thats running a check on whether this value is present or not. Surely changing just our own code in the front end is a more isolated, and thus safer change?

Unless we have some other layer here I could insert the change on that im missing..? @luis-almeida

melzreal: however instead of api level maybe what is actually meant is at api request level, where we fallback to some default values should they not be available? In which case the change would still be in this repo where we are placing the request

__mocks__/svgMock.js
melzreal: it breaks when you attempt to test a file that contains an svg. If you check guide-client where we have many more tests, you can see that the jest config already contains a  moduleNameMapper for svg

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
luis-almeida: The problem doesn't seem to be with the `description` of ticket fields. That is already an empty string when not set from what I can see.
This seems to be the `description` of the Service Catalog Item we're fetching from `/api/v2/help_center/service_catalog/items/:id` and we're in full control of that end point in Help Center.

__mocks__/svgMock.js
BrunoBFerreira: Was just wondering why we didn't see the error before. But I see that it might be because of the new tests we are adding as well. Thank you for clarifying.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
melzreal: makes sense thanks, closing this one

melzreal:
closing in favour of https://github.com/zendesk/help_center/pull/30955/files

melzreal:
closing in favour of https://github.com/zendesk/help_center/pull/30955/files

PR #559, opened 2025-01-03
i18n scripts improvements

Fredx87:
## Description

This PR adds some improvements to the scripts used for downloading and extracting translation strings for the JS modules. The scripts were added when we had only one module, so they needed some tweaks to make it easier to work with multiple modules.

### bin/update-modules-translations.mjs
- Removed the `MODULES` mapping. Now the script searches for all the files matching the `src/modules/**/translations/en-us.yml` glob and downloads the translation files.

### bin/extract-strings.mjs
- Removed the need to specify the package name, source file glob, and output file as args.  The script now searches for all the files matching the `src/modules/**/translations/en-us.yml` glob and extracts strings for each module using the `src/modules/[MODULE_NAME]/**/*.{ts,tsx}` glob.
- Added the ability to extract strings for all modules or a single module, passing a `--module` parameter.
- The script now doesn't mark removed strings as obsolete by default, but it has a new `--mark-obsolete` flag. Sometimes we remove some strings on a branch but we want to deprecate them later when we merge the branch on `master`
- We have new i18n guidelines requiring us to provide strings for all possible plural form values (zero, one, two, few, many, other). The script has been updated to extract all the needed strings for plurals.

### Note
I ran the extract-strings script on `master` and `service-catalog` branches and found some inconsistencies in the strings used in the source code and the translation files, for example:
- there is a typo here: https://github.com/zendesk/copenhagen_theme/blob/6840432aa20893fd18b5c6648217e4e403a25dcf/src/modules/new-request-form/fields/attachments/FileListItem.tsx#L104 (`arria-label`) while the key in the yml file is correct
- we are using strings from another module in some places (e.g. https://github.com/zendesk/copenhagen_theme/blob/d29494dc07c31f0d826fa4ada1515b2377648715/src/modules/service-catalog/useNotifyError.tsx#L19 where we use a string from the `new-request-form` module in the `service-catalog` module.

We will need to open other PRs to fix these issue.

## Checklist

### DIFF
--- a/bin/extract-strings.mjs
+++ b/bin/extract-strings.mjs
@@ -83,10 +83,10 @@ const OUTPUT_BANNER = `#
 
 /** @type {import("i18next-parser").UserConfig} */
 const config = {
-  // Our translation system requires that we add ".zero", ".one", ".other" keys for plurals
+  // Our translation system requires that we add all 6 forms (zero, one, two, few, many, other) keys for plurals
   // i18next-parser extracts plural keys based on the target locale, so we are passing a
-  // locale that need exactly the ".zero", ".one", ".other" keys, even if we are extracting English strings
-  locales: ["lv"],
+  // locale that need exactly the 6 forms, even if we are extracting English strings
+  locales: ["ar"],
   keySeparator: false,
   namespaceSeparator: false,
   pluralSeparator: ".",
@@ -132,7 +132,9 @@ class SourceYmlTransform extends Transform {
       }
 
       // Add new keys to the source YAML or log mismatched value
-      for (const [key, value] of Object.entries(strings)) {
+      for (let [key, value] of Object.entries(strings)) {
+        value = this.#fixPluralValue(key, value, strings);
+
         const existingPart = outputContent.parts.find(
           (part) => part.translation.key === key
         );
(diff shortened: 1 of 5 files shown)


bin/extract-strings.mjs
```diff
@@ -50,10 +83,10 @@ const OUTPUT_BANNER = `#
 
 /** @type {import("i18next-parser").UserConfig} */
 const config = {
-  // Our translation system requires that we add ".zero", ".one", ".other" keys for plurals
+  // Our translation system requires that we add all 6 forms (zero, one, two, few, many, other) keys for plurals
   // i18next-parser extracts plural keys based on the target locale, so we are passing a
-  // locale that need exactly the ".zero", ".one", ".other" keys, even if we are extracting English strings
-  locales: ["lv"],
+  // locale that need exactly the 6 forms, even if we are extracting English strings
+  locales: ["ar"],
```
melzreal, line 89: I wonder why make ar the default instead of en

melzreal (approved):
LGTM as far as I understand it, likely could do with a second set of 👀 for testing

bin/extract-strings.mjs
Fredx87: It is explained in the comment above the line:
- our translation system requires us to define 6 keys for plurals (`[key].zero`,`[key].one`, `[key].two`, `[key].few`, `[key].many`, `[key].other`)
- the tool we use (`i18next-parsers`) extracts these plural keys based on the source locale. Since English has only `one` and `other` forms, if we set `en` as the source locale it would extract only `[key].one` and `[key].other`
- `ar` has all six forms, so setting it as the source locale has the effect of letting the tool extract all the needed strings

gosiexon-zen:
> I ran the extract-strings script on master and service-catalog branches and found some inconsistencies in the strings used in the source code and the translation files, for example:

My bad, sorry. I have PR opened with fetching translations for this `new-request-form.attachments.remove-file-aria-label` so I can update the label in the code.

gosiexon-zen (approved):

PR #560, opened 2025-01-03
fix: added useNotify hook and fixed shared close translation

Fredx87:
## Description

This PR moves the `useNotifyError` hook to the shared modules and generalizes it to `useNotify` to be used for all types of notifications.
- The `service-catalog-module` had a `useNotifyError` hook which was using the `new-request-form.close-label` translation from another module
- The `useWysiwyg` hook in the `ticket-fields` module was using the `new-request-form.close-label` translation from another module
- The `useNotifyError` hook has been moved to the shared package, renamed to `useNotify`, and an option to pass a `type` has been added to show different notifications
- The hook uses a new `cph-theme-shared.close-label` from the shared module
- Added the loading of the shared module translations in all modules

## Checklist


src/modules/flash-notifications/renderFlashNotifications.tsx
```diff
@@ -31,10 +38,7 @@ export function renderFlashNotifications(
 
     render(
       <ThemeProviders theme={createTheme(settings)}>
-        <FlashNotifications
-          notifications={parsedNotifications}
-          closeLabel={closeLabel}
-        />
+        <FlashNotifications notifications={parsedNotifications} />
```
BrunoBFerreira, line 41: Don't we need the `closeLabel`? Are we removing it for any specific reason?

src/modules/shared/translations/en-us.yml
```diff
@@ -25,3 +26,7 @@ parts:
       title: "Error state go to the homepage link"
       screenshot: "https://drive.google.com/file/d/1rJ9c4mW0YWnL5d6KWWZaY9TJBEH5otLL/view?usp=drive_link"
       value: "Go to the homepage"
+  - translation:
```
BrunoBFerreira, line 29: Is the addition of this translation the reason why we are removing the close label?

BrunoBFerreira (approved):
LGTM 👍  left just a couple of questions.

src/modules/shared/translations/en-us.yml
Fredx87: Yes, before we were reading the `closeLabel` from Curlybars using the `t` helper: https://github.com/zendesk/copenhagen_theme/blob/a808e3c7e9b6614c2364f9172245923523461f9f/templates/document_head.hbs#L29 and we were passing it to the JS module.
This was done just because we had only 1 string to load in the `flash-notifications` module and seemed overkill to create a JSON translation file with just that string

Now we have a translation file for the shared module with this string, and we can fetch it like the other strings

kubraokcu (approved):
👍🏼🌐

src/modules/shared/translations/en-us.yml
BrunoBFerreira: Thank you for explaining. I think this answers my other comment as well.

PR #567, opened 2025-01-22
fix: removed link to service catalog documentation

Fredx87:
## Description

This PR removes the link to the service catalog documentation, which is not needed for end users.
It also marks unused strings in the service catalog module as deprecated

## Screenshots

Screenshot 2025-01-22 at 16 24 23

## Checklist


src/modules/service-catalog/translations/en-us.yml
```diff
@@ -76,6 +77,7 @@ parts:
       title: "[A11Y] Hidden label for screen readers for close buttons (e.g. in modals and notifications)"
       screenshot: ""
       value: "Close"
+      obsolete: "2025-04-22"
```
BrunoBFerreira, line 80: Quick question: are we marking this one as obsolete to do a bit of clean up or is there any other reason?

BrunoBFerreira (approved):
LGTM 👍 left a small question but nothing that would block this.

src/modules/service-catalog/translations/en-us.yml
Fredx87: It is a cleanup and also because we don't want to have these strings that are not used to be included in the json files we download for each locale, otherwise we increase the bundle size with unuseful strings.

When the obsolete date passes our translation system will stop to include them in the JSON files. The suggested practice is to put a date 2/3 months in the future just a precaution if eventually the string is reused somewhere

kubraokcu (approved):

PR #568, opened 2025-01-23
Change type of ticket form id in service catalog item

gosiexon-zen:
## Description
We needed to use type `text` for `brand_id` and `ticket_form_id` as type integer has limitations. Now when the limit is raised we can change it to `integer`. 

Jira issue: https://zendesk.atlassian.net/browse/GG-4073
## Checklist


src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -108,7 +108,10 @@ export function useItemFormFields(
       if (serviceCatalogItem && serviceCatalogItem.form_id) {
         try {
           const { requestFields, associatedLookupField } =
-            await fetchTicketFields(serviceCatalogItem.form_id, baseLocale);
+            await fetchTicketFields(
+              serviceCatalogItem.form_id.toString(),
```
Fredx87, line 112: Nit: I would rather change the `form_id` parameter of the function to be a number, and rename it to `formId` for coherence

Fredx87 (approved):

src/modules/service-catalog/components/useItemFormFields.tsx
BrunoBFerreira: Updated. Could you re-review @Fredx87 ?

Fredx87 (approved):

PR #572, opened 2025-01-29
fix: set default value for Service Catalog fields

BrunoBFerreira:
## Description

This PR fixes a bug where Dropdown fields for Service Catalog Items did not use their default value (if they had one defined) when first accessing the page to request a service. 

### Before

Screenshot 2025-01-29 at 17 54 00

### After

Screenshot 2025-01-29 at 17 54 41

## Checklist


src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -4,6 +4,11 @@ import type { TicketField } from "../data-types/TicketField";
 import type { Field } from "../../ticket-fields/data-types/Field";
 import { getCustomObjectKey } from "../../ticket-fields/fields/LookupField";
 
+const get_field_value = (field: TicketField) =>
```
melzreal, line 7: should be camelCase like all the other functions here?

melzreal (commented):
You can avoid regenerating the bundles with one of the package.json scripts that run rollup - I think yarn build

Fredx87 (approved):

melzreal (approved):

PR #578, opened 2025-02-11
fix: breadcrumbs for long titles

gosiexon-zen:
## Description
We are using the title of the Services in the breadcrumbs for Service Catalog pages. However, when the title is too long, the breadcrumbs break. This PR fixes this issue.

[GG-4108](https://zendesk.atlassian.net/browse/GG-4108)
## Screenshots
Before:
Screenshot 2025-02-11 at 14 00 55
After:
Screenshot 2025-02-11 at 14 00 45

## Checklist

[GG-4108]: https://zendesk.atlassian.net/browse/GG-4108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Fredx87 (commented):
This change is going to impact all the breadcrumbs in Help Center and not only the ones in the Service Catalog, are we sure we want to do that?

For example, this is what happens when a section has a very long title: 
Screenshot 2025-02-11 at 14 41 17

The breadcrumbs for Service Catalog are also different from the ones shown in articles and posts, where the article or post title is not shown in the breadcrumbs.

Moreover, we shouldn't edit the `style.css` file directly, but we should edit the SCSS files

gosiexon-zen:
@Fredx87 so should we create our own breadcrumbs style for SC?

Fredx87:
No, because otherwise, they will be inconsistent with the other pages. For now, let's keep the current breadcrumbs even if they don't look that good with long items

PR #598, opened 2025-03-05
feat: add search to Services List

BrunoBFerreira:
## Description

We want to give the users the ability to search by the services they have available. 

To achieve that, this PR adds a search box to the Services list page, allowing users to input a string a search for Services that contain that input in their text based fields.

### Before

Screenshot 2025-03-05 at 11 25 12

### After

Screenshot 2025-03-05 at 11 18 54

When loading with search input

Screenshot 2025-03-05 at 11 31 48

Search without results

Screenshot 2025-03-05 at 11 41 40

## Checklist


kubraokcu (approved):
👍🏼🌐 for the wip strings

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -109,17 +127,20 @@ export function ServiceCatalogList({
             <Row wrap="wrap">
               {serviceCatalogItems.length !== 0 &&
                 serviceCatalogItems.map((record: ServiceCatalogItem) => (
-                  <StyledCol key={record.id} xs={12} sm={6} md={4} lg={3}>
+                  <Col key={record.id} xs={12} sm={6} md={4} lg={3}>
```
Fredx87, line 130: Why are we removing the styled col with the margin bottom here?
It removes the space between items when there are multiple rows
Screenshot 2025-03-05 at 15 11 27

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -50,14 +51,26 @@ export function ServiceCatalogList({
   }
 
   useEffect(() => {
+    async function fetchServiceCatalogItems() {
+      let requestUrl = "/api/v2/help_center/service_catalog/items";
+
+      if (searchInputValue) {
+        requestUrl += `/search?query=${searchInputValue}&`;
+      } else {
+        requestUrl += "?";
```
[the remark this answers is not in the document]
BrunoBFerreira: For the first question, I believe we can, given that we end up calling the same endpoint in the backend's backend, we either just send the query parameter or not.

For the second point, it is possible to have a minimum number of characters typed before sending the query (it was even shown in the demo UI) but what would happen if the client has a service with a two letter name? Is it unsearchable? As for the caching, given that Services can be created quite frequently (specially in the beginning), I believe we can skip it for now. We can always re-assess this later and add it if we need it.

I will add the debounce (I completely forgot 🙈 )

src/modules/service-catalog/ServiceCatalogList.tsx
BrunoBFerreira: My bad. I will fix it.

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -25,72 +27,58 @@ const StyledGrid = styled(Grid)`
...
-      }
-    }
-    fetchData();
-  }, [currentCursor, notify, t]);
+    return () => debouncedUpdateServiceCatalogItems.cancel();
+  }, [debouncedUpdateServiceCatalogItems]);
+
+  useEffect(() => {
+    debouncedUpdateServiceCatalogItems(searchInputValue, currentCursor);
+  }, [searchInputValue, currentCursor, debouncedUpdateServiceCatalogItems]);
```
Fredx87, line 81: There are some issues with this implementation of the debouncing:
- When clicking on the `next` and `prev` buttons, we use the denounced function as well, this seems to cause the API to be called multiple times, and the UI is flickering
- When the user is on a page other than the first, searching for items don't reset the page number, so the user will not see any result because he is in the wrong page
- This chain of `useEffect` called when different values are changing are difficult to track and reason about

https://github.com/user-attachments/assets/2b3d0e3c-6abb-4606-81fc-3285597096ab

We should simplify things and reduce the need to use `useEffect`, we don't need it for most of the cases (see https://react.dev/learn/you-might-not-need-an-effect). We need to fetch the services:
- When the user types something in the search bar. We can handle it in the onChange event and call the denounced function
- When prev/next buttons are clicked, we can handle it in the respective event handler and call the normal function
- At first page load, we can use a `useEffect` to do it on component mount

We should remove the parameters from the hook and keep them just on the function, and we should remove the `useEffect`. I am not even sure we need the `currentCursor` at all, since we can check the `meta` instead.

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -25,72 +27,58 @@ const StyledGrid = styled(Grid)`
...
+    throw errorFetchingItems;
   }
 
+  const debouncedUpdateServiceCatalogItems = useMemo(
+    () =>
+      debounce(
+        (searchInputValue: string, currentCursor: string | null) =>
+          fetchServiceCatalogItems(searchInputValue, currentCursor),
+        300
+      ),
```
Fredx87, line 71: This can be simplified as
```ts
debounce(fetchServiceCatalogItems, 300)
```

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -0,0 +1,46 @@
...
+  setSearchInputValue,
+}: SearchProps) => {
+  const { t } = useTranslation();
+
+  return (
+    <StyledField>
+      <Label hidden>
+        {t("service-catalog.search-services", "Search for services")}
+      </Label>
+      <MediaInput
```
Fredx87, line 34: We should also set `type="search"` for better a11y and `autoComplete="off"` to avoid browsers presenting autocomplete values

src/modules/service-catalog/translations/en-us.yml
```diff
@@ -118,3 +118,21 @@ parts:
       title: "Label indicating number of services currently being listed, the default plural form."
       screenshot: "https://drive.google.com/file/d/1uP3fyDR1vE2Z8lHXiNXaZD7LeTEgLCSw/view?usp=drive_link"
       value: "{{count}} services"
+  - translation:
+      key: "service-catalog.search-services"
+      title: "Search for services in the Service List"
+      screenshot: "https://drive.google.com/file/d/1JHDB_jxDaL0ozmpUTYIhYMaCm4-4SN27/view?usp=drive_link"
+      value: "Search for services"
+      wip: true
```
Fredx87, line 126: Are we waiting for approval of these strings?
Otherwise we can remove the `wip` tag

src/modules/service-catalog/translations/en-us.yml
```diff
@@ -118,3 +118,21 @@ parts:
...
+  - translation: 
+      key: "service-catalog.empty-state.search-description"
+      title: "Empty state with search description"
+      screenshot: "https://drive.google.com/file/d/10DiQB3DDf9x13owivlJs9QMNWVXCKdOd/view?usp=drive_link"
+      value: "Enter your keywords in the search field."
+      wip: true
+  - translation:
+      key: "service-catalog.clear-search"
```
Fredx87, line 134: Do we need this string, given that we are not shipping this feature now?

src/modules/service-catalog/useServiceCatalogItems.tsx
```diff
@@ -0,0 +1,76 @@
...
+      setIsLoading(true);
+
+      let requestUrl = "/api/v2/help_center/service_catalog/items/search?";
+
+      requestUrl +=
+        searchInputValue.length >= 3 ? `query=${searchInputValue}&` : "";
+
+      requestUrl += currentCursor
+        ? `page[size]=${PAGE_SIZE}&${currentCursor}`
+        : `page[size]=${PAGE_SIZE}`;
```
Fredx87, line 40: Nit: I think it is more readable if we use [UrlSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) instead of string interpolation

src/modules/service-catalog/translations/en-us.yml
BrunoBFerreira: Forgot to remove this one. Thank you 🙏

src/modules/service-catalog/translations/en-us.yml
BrunoBFerreira: I'm not sure if the design will change. However, considering that the changes we need are related with the clear all button, I think we can assume the rest of the strings should be final. I'll remove the wip.

BrunoBFerreira:
@Fredx87 I believe I have addressed all the comments. Could you take another look?
@kubraokcu I have also removed the WIP from the strings, so not sure if you want to review again as well.

Thank you both 🙏

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -25,82 +27,75 @@ const StyledGrid = styled(Grid)`
...
+        searchInputValue,
+        "page[before]=" + meta.before_cursor
+      );
+    }
+  };
+
+  const handleInputChange = (value: string) => {
+    setSearchInputValue(value);
+    if (value.length >= 3) {
+      debouncedUpdateServiceCatalogItems(value, null);
```
Fredx87, line 98: I suggested this change in the last review, but now trying it I realize it is wrong 😄 
If the user searches for something with more than 3 characters and then removes them, we are not fetching the data anymore. This means that we continue to show old results, even when the field is empty.

The previous version was better even if not ideal, maybe we should consult with Guntis on how it should work or maybe we can remove the limits on the characters

Fredx87 (approved):

PR #599, opened 2025-03-05
[GG-4166] Add service count to service list

BrunoBFerreira:
## Description

In addition to the information already available to the user in the Services list page, we also want to display the total number of Services available in the list.

This PR adds an extra element in the UI to display the total count returned by the endpoint. 

### Before

Screenshot 2025-03-05 at 13 23 41

### After

Screenshot 2025-03-05 at 13 24 48

## Checklist


src/modules/service-catalog/translations/en-us.yml
```diff
@@ -88,3 +88,9 @@ parts:
       title: "Error notification when there is problem with loading the list of services"
       screenshot: "https://drive.google.com/file/d/1vIeLVSOOV9-vUj-Kh9-ZwyFSC9jN6AFs/view?usp=drive_link"
       value: "Give it a moment and try it again"
+  - translation:
+      key: service-catalog.services-count
+      title: "Number of services currently being listed"
+      screenshot: "https://drive.google.com/file/d/1uP3fyDR1vE2Z8lHXiNXaZD7LeTEgLCSw/view?usp=drive_link"
+      value: "{{count}} services"
```
kubraokcu, line 95: Fyi, when you'll unwip the string we'll ask if you could create 5 variations to accommodate all our languages plural forms: https://zendeskdev.zendesk.com/hc/en-us/articles/6853841965722-How-do-I-manage-plurals-in-translation-strings-for-internationalization-i18n

kubraokcu (approved):
👍🏼🌐 for the wip string

src/modules/service-catalog/translations/en-us.yml
BrunoBFerreira: After speaking with @Fredx87, we decided we can add the plural forms already given that, although there are some parts of the design that might not be final, the count is not one of them.

Could you take another look? 🙏

BrunoBFerreira:
> Left a couple of comments. Does it make sense to have this PR separate from the search? I see them as two closely related things, but it is ok also to have them separate

I thought the same thing. However, since we are still waiting to confirm a couple of things related with the search, we can ship these change already given it will work independently from that. We will have the count even if the search is empty or not being used at all.

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -40,6 +43,7 @@ export function ServiceCatalogList({
   >([]);
   const [meta, setMeta] = useState<Meta | null>(null);
   const [currentCursor, setCurrentCursor] = useState<string | null>(null);
+  const [count, setCount] = useState<number>(0);
```
Fredx87, line 46: Nit: we can omit `number` since it can be inferred from 0

src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -101,6 +106,17 @@ export function ServiceCatalogList({
...
   return (
     <Container>
+      <span>
+        {t("service-catalog.service-count", {
+          "defaultValue.zero": "{{count}} services",
+          "defaultValue.one": "{{count}} service",
+          "defaultValue.two": "{{count}} services",
+          "defaultValue.few": "{{count}} services",
+          "defaultValue.many": "{{count}} services",
+          "defaultValue.other": "{{count}} services",
```
Fredx87, line 116: There is no need to specify all the possible variants here since these are default values used only for English if the string is not present in the translation files. We can just specify the `one` and `other` variants, which are the only ones needed for English

src/modules/service-catalog/translations/en-us.yml
kubraokcu: Sure, looks good, thank you!

kubraokcu (approved):

Fredx87 (approved):

Fredx87 (approved):

PR #609, opened 2025-03-17
feat: handle missing forms for Service Catalog requests

BrunoBFerreira:
## Description

Currently request forms for Service Catalog can be seen and edited/deleted from the Guide Admin pages. If a service request form is deleted without deleting the service item, the item can still be accessed in the Services page and requested. However, when an employee submits a service request of that type, the request form will default to the id of the account’s default ticket form.

To prevent this from happening, this PR updates the implementation to check if the form being return has the `active: true` field or if the `form_id` being returned matches the id used to fetch the form. If one of these conditions is not met, we show an error in the service page, preventing the user from requesting the service incorrectly. 

## Screenshots

Screenshot 2025-03-17 at 17 28 43

Error screen in the service page when corresponding request form is missing. 

## Checklist


luis-almeida (commented):
> > If a service request form is deleted without deleting the service item, the item can still be accessed in the Services page and requested. However, when an employee submits a service request of that type, the request form will default to the id of the account’s default ticket form.

Wouldn't the fetch call to `/api/v2/ticket_forms/${form_id}` fail with `404` in this scenario?

BrunoBFerreira:
> > If a service request form is deleted without deleting the service item, the item can still be accessed in the Services page and requested. However, when an employee submits a service request of that type, the request form will default to the id of the account’s default ticket form.
> 
> Wouldn't the fetch call to `/api/v2/ticket_forms/${form_id}` fail with `404` in this scenario?

That is a good question. The issue here is that forms are soft deleted. So although the form is "deleted" it will still be available and will be fetched but with the `active` property set to false. 

That is why I added a check for the active field as well.

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -71,6 +71,10 @@ const fetchTicketFields = async (
   const formData = await formResponse.json();
   const fieldsData = await fieldsResponse.json();
 
+  if (!formData.ticket_form.active || formData.ticket_form.id !== form_id) {
```
luis-almeida, line 74: Moving this convo here:

> > Wouldn't the fetch call to `/api/v2/ticket_forms/${form_id}` fail with 404 in this scenario?

> That is a good question. The issue here is that forms are soft deleted. So although the form is "deleted" it will still be available and will be fetched but with the active property set to false.
> That is why I added a check for the active field as well. Hope this answers your question.

That makes sense for the first check: `formData.ticket_form.active`.
I'm trying to understand the second one: `formData.ticket_form.id !== form_id`.
Is there a case where a call to `/api/v2/ticket_forms/${form_id}` would return a different form id? 😕

luis-almeida: It'd be really really weird to do fetch for a form but get another 🤨 At this point nothing really surprises me though 😅 

Nonetheless, we should get to the bottom as these checks add quite a lot of distrust in the APIs and from my experience we'll never dare to remove them.

BrunoBFerreira: I was checking and @Fredx87 is right. We fallback to other forms when we create a request with a form that is inactive/soft deleted. This check is not needed. We already handle the errors when the response is a 404, so we should cover the rest already. I'll update the code and the tests.

BrunoBFerreira: I was able to test both scenarios (inactive form and missing form) and in fact we did not need that check. With the code as it is now, we should be able to handle both situations.

@luis-almeida and @Fredx87 can you re-review the PR? 🙏 
Also thank you both for the input.

src/modules/service-catalog/components/useItemFormFields.spec.ts
```diff
@@ -0,0 +1,394 @@
+import { act, renderHook } from "@testing-library/react-hooks";
+import type { ServiceCatalogItem } from "../data-types/ServiceCatalogItem";
+import { useItemFormFields } from "./useItemFormFields";
+
+describe("useItemFormFields", () => {
+  const mockedItem: ServiceCatalogItem = {
```
luis-almeida, line 6: Nit: no need to call it `mocked*` naming in the specs. `item` or `serviceCatalogItem` would be simpler as we do in the following variables.

src/modules/service-catalog/components/useItemFormFields.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+    description: "Test Description 3",
+    label: "Test Title 3",
+    options: [],
+    required: true,
+    relationship_target_type: undefined,
+    error: null,
+    value: "Test Value 3",
+  };
+
+  const mockedBaseLocale = "en-us";
```
luis-almeida, line 68: Nit: same here

src/modules/service-catalog/components/useItemFormFields.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("should initialize with default values", async () => {
+    const mockedFormResponse = {
```
luis-almeida, line 75: Nit: same throughout the spec.

src/modules/service-catalog/components/useItemFormFields.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+    await waitForNextUpdate();
+
+    expect(result.current.requestFields).toEqual([expectedTextField]);
+
+    expect(result.current.associatedLookupField).toEqual(expectedLookupField);
+    expect(result.current.error).toBe(null);
+  });
+
+  it("should return an error if form data fetch fails", async () => {
+    global.fetch = jest.fn(() =>
```
luis-almeida, line 120: Nit: We could also "make it" a `jest.Mock` before the tests and then just mock the resolved value when needed:

``` ts
const globalFetch = ((globalThis.fetch as jest.Mock) = jest.fn());

//...

globalFetch.mockResolvedValue({ status: 500, ok: false })

```

src/modules/service-catalog/components/useItemFormFields.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+
+    global.fetch = jest.fn((url) => {
+      if (url.includes("/ticket_forms")) {
+        return Promise.resolve({
+          json: () => Promise.resolve(mockedFormResponse),
+          status: 200,
+          ok: true,
+        });
+      } else {
+        return Promise.resolve({
```
luis-almeida, line 95: Can we target the URLs more specifically so we don't rely on the current implementation:
```ts
return Promise.resolve({
  json: () => Promise.resolve("/api/v2/ticket_forms/1"
    ? formResponse
    : "/api/v2/ticket_fields?locale=en-us"
    ? ticketFieldResponse
    : {}),
  status: 200,
  ok: true,
});
```

Same in the following ones.

src/modules/service-catalog/components/useItemFormFields.tsx
```diff
@@ -71,6 +71,10 @@ const fetchTicketFields = async (
   const formData = await formResponse.json();
   const fieldsData = await fieldsResponse.json();
 
+  if (!formData.ticket_form.active) {
+    throw new Error("Error fetching form data");
```
luis-almeida, line 75: Can we be more specific in the error here? Would help when debugging the issue cos we'd see it in the console right away.

Similar to what we do above with "Associated lookup field not found", e could say "Associated ticket form is not active".

luis-almeida (approved):

PR #610, opened 2025-03-19
feat: conditionally render pagination in Services page

BrunoBFerreira:
## Description

Previously, we were always rendering the pagination elements, even when there were no more elements to show. The services page would always show the pagination, but it will be deactivated when no cursors were available.

To make the experience consistent with what we have in Guide Admin for Service Catalog, we are removing the elements altogether when there are no more elements to show. 

### Before

Screenshot 2025-03-19 at 14 54 36

### After

Screenshot 2025-03-19 at 14 55 08

## Checklist


src/modules/service-catalog/ServiceCatalogList.tsx
```diff
@@ -135,17 +135,17 @@ export function ServiceCatalogList({
               searchInputValue={searchInputValue}
             />
           )}
-          {serviceCatalogItems.length > 0 && (
+          {meta && (meta.before_cursor || meta.after_cursor) && (
```
luis-almeida, line 138: Nit: would `meta?.before_cursor || meta?.after_cursor` be enough?

luis-almeida (approved):

src/modules/service-catalog/ServiceCatalogList.tsx
BrunoBFerreira: I included the `meta` there to avoid using the `?` both here and when fetching the cursors for the Next and Previous elements. By having it like this, we can infer that `meta` is not null in this scope.
Not sure which version is the better approach or common practice. Let me know what you think @luis-almeida.

PR #614, opened 2025-03-25
feat: add clear all to Services search bar

BrunoBFerreira:
## Description

In a [previous PR](https://github.com/zendesk/copenhagen_theme/pull/598) we have added a search bar element to the Services page, available to the Service Catalog EAP users.

However, we did not implement the proposed design completely since we were having issues with some components made available from Garden.

After reaching out to the Garden team, we received guidance on how to implement the missing element, the clear all button, and we have now implemented it in this PR. 

## Screenshots

service-catalog clear-search

## Checklist


gosiexon-zen (approved):

olivia-tsao (approved):

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -13,6 +18,17 @@ const StyledField = styled(Field)`
...
 
+const StyledEndIconContainer = styled.div`
+  position: absolute;
+  top: 50%;
+  left: 99%;
+  transform: translate(-99%, -50%);
+`;
+
+const StyledMediaInput = styled(MediaInput)`
+  padding-right: ${(props) => props.theme.space.md};
```
Fredx87, line 29: It feels to me that the input is a bit too close to the icon. Why are we setting it specifically to `space.md`? The container has a padding of 10px, we are adding 20px more reaching 30px, but the icon button is 32px

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -26,12 +42,34 @@ export const Search = ({
 }: SearchProps) => {
   const { t } = useTranslation();
 
+  const SearchEndIcon = () => {
```
Fredx87, line 45: We shouldn't define a component inside a component like this, but externally. It probably make sense to move the condition out of the component and just not render it when the condition is not met

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -26,12 +42,34 @@ export const Search = ({
...
+    if (!isLoading && searchInputValue) {
+      return (
+        <Tooltip
+          content={t("service-catalog.clear-search", "Clear search")}
+          placement="bottom"
+          size="small"
+        >
+          <IconButton
```
Fredx87, line 53: I think we need an aria-label for a11y, even if we already have the tooltip. In the Admin part we omitted aria labels when there is already a tooltip, but then we received some reports from a11y that we need an aria-label as well.

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -43,6 +81,9 @@ export const Search = ({
         )}
         onChange={(event) => onChange(event.target.value)}
       />
+      <StyledEndIconContainer>
```
Fredx87, line 84: Nit: maybe we can apply the styles directly to the `SearchEndIcon` component itself instead of creating a container component, but it is a minor thing, we can also keep the container

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -1,10 +1,15 @@
...
 import { Dots } from "@zendeskgarden/react-loaders";
+import { Tooltip } from "@zendeskgarden/react-tooltips";
 import SearchIcon from "@zendeskgarden/svg-icons/src/16/search-stroke.svg";
+import EndIcon from "@zendeskgarden/svg-icons/src/16/x-stroke.svg";
 import { useTranslation } from "react-i18next";
 import styled from "styled-components";
 
 const StyledField = styled(Field)`
+  display: flex;
+  flex-direction: row;
```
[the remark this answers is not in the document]
BrunoBFerreira: Leftover from something I was experimenting with. I'm going to remove it 😄 Thank you for pointing this out.

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -13,6 +18,17 @@ const StyledField = styled(Field)`
   }
 `;
 
+const StyledEndIconContainer = styled.div`
+  position: absolute;
+  top: 50%;
+  left: 99%;
+  transform: translate(-99%, -50%);
```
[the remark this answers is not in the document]
BrunoBFerreira: Yeah I was a bit on the fence about using hardcoded values but you raise a good point. I'll update this.

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
BrunoBFerreira: I believe I saw the incorrect values on the design. I will fix it 👍

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -13,6 +16,16 @@ const StyledField = styled(Field)`
...
 `;
 
+const StyledIconButton = styled(IconButton)`
+  position: absolute;
+  top: 4px;
+  right: 4px;
+`;
+
+const StyledMediaInput = styled(MediaInput)`
+  padding-right: 22px;
```
Fredx87, line 26: I think we should add a bit more space, maybe 4px, the padding is just the minimum needed

Screenshot 2025-03-26 at 17 57 48

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -26,12 +39,31 @@ export const Search = ({
...
 
+  const SearchClearIcon = () => {
+    return (
+      <Tooltip
+        content={t("service-catalog.clear-search", "Clear search")}
+        placement="bottom"
+        size="small"
+      >
+        <StyledIconButton
+          aria-label={t("service-catalog.clear-search")}
```
Fredx87, line 50: We also need to pass "Clear Search" again as the second parameter, otherwise, until the string is back, we have `service-catalog.clear-search` as a value for the attribute

src/modules/service-catalog/components/service-catalog-list-item/Search.tsx
```diff
@@ -13,6 +14,10 @@ const StyledField = styled(Field)`
   }
 `;
 
+const StyledMediaInput = styled(MediaInput)`
+  padding-right: ${(props) => props.theme.space.base * 7}px;
```
Fredx87, line 18: I just realized that this doesn't work for RTL languages. We should use `padding-inline-end` instead

src/modules/service-catalog/components/service-catalog-list-item/SearchClearIcon.tsx
```diff
@@ -0,0 +1,36 @@
+import ClearIcon from "@zendeskgarden/svg-icons/src/16/x-stroke.svg";
+import { Tooltip } from "@zendeskgarden/react-tooltips";
+import { IconButton } from "@zendeskgarden/react-buttons";
+import { useTranslation } from "react-i18next";
+import styled from "styled-components";
+
+const StyledIconButton = styled(IconButton)`
+  position: absolute;
+  top: 4px;
+  right: 4px;
```
Fredx87, line 10: I just realized that this doesn't work for RTL languages. We should use `inset-inline-end` instead

Fredx87 (approved):

PR #622, opened 2025-04-02
chore: add services search strings for translation

BrunoBFerreira:
## Description

We have added new strings related to the Services search feature in our beta branch. However, given that our translation tool only picks up new strings from master, we did not get the required translation variants.

This PR adds the strings to master so we are able to get the translations, which will then be added to the beta branch in a future PR. 

## Checklist


src/modules/service-catalog/translations/en-us.yml
```diff
@@ -117,3 +117,18 @@ parts:
       title: "Label indicating number of services currently being listed, the default plural form."
       screenshot: "https://drive.google.com/file/d/1uP3fyDR1vE2Z8lHXiNXaZD7LeTEgLCSw/view?usp=drive_link"
       value: "{{count}} services"
+  - translation:
+      key: "service-catalog.search-services"
```
melzreal, line 121: are these used anywhere yet? would we not add them at the same time?

melzreal (approved):

kubraokcu (approved):

PR #630, opened 2025-05-02
Fix lookup relationship fields (lurf) in service catalog forms

luis-almeida:
## Description

1. `organizationField` was not passed to the `TicketField` component as it was marked as an "optional"
2. The problem was that inside the `TicketField` component we checked if `organizationField !== null` when building up the API endpoint and since `organizationField` was `undefined` it resulted in broken API calls:

image

This PR addresses this issue and just passed `organizationField={null}` since the "organization field" is a standard field and can't be added to a Service Catalog Item form.

## Checklist


luis-almeida:
> Are there any unit tests we could add for this?

Where would you add you add it?

luis-almeida:
Sorry for the late reply @akuttruff and thanks for addressing this one!
I wound up forgetting about merging it so was just getting back to it now.

PR #633, opened 2025-05-07
feat: display any file validation errors that can occur when the "Allowed file types" setting is on

oorion:
## Description

feat: display any file validation errors that can occur when the "Allowed file types" setting is on
## Checklist


src/modules/new-request-form/fields/attachments/Attachments.tsx
```diff
@@ -61,8 +61,20 @@ export function Attachments({ field }: AttachmentProps): JSX.Element {
   const { addToast } = useToast();
   const { t } = useTranslation();
 
+  const errorMessage = (file: File, xhr: XMLHttpRequest) => {
+    if (xhr.response?.error == "RecordInvalid") {
+      return xhr.response?.details?.base?.[0]?.description;
```
gosiexon-zen, line 66: Is this description translated on backend? 
Are we sure it is always first element here: `base?.[0]`?

oorion: Yes, it's translated on the backend but you're correct in thinking that there could be more than 1 element.  The other error messages that I've checked so far also appear to be translated on the backend.  I'm going to have to reach out to product to see how they want these errors to be displayed in the ui when there are more than 1 as it also affects this line of another pr https://github.com/zendesk/zendesk_console/pull/44689/files#diff-f29f8e4f8475555212f6d45c2b4a805e32a9b33ed3938582c5cd5b297400289fR456.  Perhaps we can display them separated with a `; ` but we'll see.  Thanks for asking the question

src/modules/new-request-form/fields/attachments/Attachments.tsx
```diff
@@ -61,8 +61,20 @@ export function Attachments({ field }: AttachmentProps): JSX.Element {
   const { addToast } = useToast();
   const { t } = useTranslation();
 
+  const errorMessage = (file: File, xhr: XMLHttpRequest) => {
+    if (xhr.response?.error == "RecordInvalid") {
```
oorion, line 65: Note to self: also check that `xhr.response?.details?.base` is present here because there may be an InvalidRecord response without that key in the details

src/modules/new-request-form/fields/attachments/Attachments.tsx
gosiexon-zen: Ok, so maybe adding additional check to be sure it is present will help and also if you want maybe adding some type for that like:
```javascript
interface UploadErrorResponse {
  error?: string;
  details?: {
    base?: Array<{ description: string }>;
  };
}
```

src/modules/new-request-form/translations/en-us.yml
```diff
@@ -185,4 +186,14 @@ parts:
...
-      value: "Search {{label}}"    
+      value: "Search {{label}}"
+  - translation:
+      key: "new-request-form.attachments.upload-failed-title"
+      title: "Title of the notification shown when the upload of an attachment failed"
+      screenshot: "https://drive.google.com/file/d/1HaYFIjlr89IWfu-HweJbTRjC23u3bLP-/view?usp=sharing"
+      value: "{{fileName}} upload failed"
+  - translation:
+      key: "new-request-form.attachments.error-separator"
+      title: "Separator between errors for the same file (usually '; ')"
```
kubraokcu:
```suggestion
      title: "Separator between errors for the same file - please localize according to your language's punctuation rules"
```

gosiexon-zen (approved):

PR #639, opened 2025-05-21
Update dev dependencies

Fredx87:
## Description

This PR updates some development dependencies to fix vulnerability issues.

- Updated Node version to v22.15.0
   - Node 18 is not supported anymore and some updated dependencies need a new version. v22 is the [current LTS version](https://nodejs.org/en/about/previous-releases)
- Updated `zcli` to the latest version
  - Fixes https://github.com/zendesk/copenhagen_theme/security/dependabot/106
  - Note: this version bump includes this change: https://github.com/zendesk/zcli/pull/238, some old logins using passwords instead of API tokens could stop working. Re logging in via ZCLI using an API token should fix the issue
- Updated `lighthouse`
  - Fixes https://github.com/zendesk/copenhagen_theme/security/dependabot/107
  - The new version returned some new violations that have been added to the ignored rules
- Updated `puppeteer`
  -   Uses an updated version of the `ws` package, so we can remove the resolution to the specific `^8.17.1` version
- Updated `wait-on` 
  - Fixes https://github.com/zendesk/copenhagen_theme/security/dependabot/106
- Updated `i18next-parser`
  - Fixes https://github.com/zendesk/copenhagen_theme/security/dependabot/93 and https://github.com/zendesk/copenhagen_theme/security/dependabot/77

I manually ran all the scripts to check that they still work properly with the new node versions and the new dependencies.

## Checklist


yarn.lock
```diff
@@ -13890,7 +13963,12 @@ write-file-atomic@^4.0.0, write-file-atomic@^4.0.1, write-file-atomic@^4.0.2:
     imurmurhash "^0.1.4"
     signal-exit "^3.0.7"
 
-ws@8.13.0, ws@8.17.0, ws@^7.0.0, ws@^8.11.0, ws@^8.13.0, ws@^8.17.1:
+ws@^7.0.0:
```
melzreal, line 13966: this version that has a CVE vulnerability was re-introduced when you removed the ws from the resolutions? Why remove it?

bin/lighthouse/config.js
```diff
@@ -51,10 +51,28 @@ module.exports = {
       ],
       "td-has-header": [
         {
-          path: "/hc/en-us/subscriptions",
+          path: "/hc/:locale/subscriptions",
```
BrunoBFerreira, line 54: Quick question: do we need this change due to the library updates?

BrunoBFerreira (approved):

yarn.lock
Fredx87: The lockfile sets the version `7.5.10` for the range `^7.0.0`. The version doesn't have a CVE vulnerability: https://github.com/zendesk/copenhagen_theme/security/dependabot/76. I also checked with `yarn audit` locally, and no audit is reported for the `ws` package.

Setting a fixed value in `resolution` could cause issues in the future. For example, the fixed version we choose could have some new vulnerabilities, or an updated dependency will require a new major version and we would force it to use the v8.x. I think it is better not to use it unless there is no other way.

bin/lighthouse/config.js
Fredx87: No. The URLs that are checked by the tool are [configurable by the user](https://github.com/zendesk/copenhagen_theme/blob/master/README.md#accessibility-testing). Somebody may put URLs with a different locale than `en-us`, so it is better to define the ignored rules with a placeholder. This one was added in #637 without the placeholder, probably by mistake

yarn.lock
melzreal: oh I see, it doesn't matter that this reads ws@^7.0.0, because the resolution is 7.5.10. Thanks for explaining 👍

melzreal (approved):

PR #642, opened 2025-06-02
feat: includes dynamic filter params for lookup autocomplete calls

akuttruff:
@zendesk/vikings

## Description
Includes query parameters from form field values to [support dynamic filters](https://support.zendesk.com/hc/en-us/articles/6941856507930-Announcing-dynamic-filtering-for-lookup-relationship-fields-on-tickets) for lookup autocomplete calls. We currently allow lookup autocomplete results to be filtered based on dynamically-selected assignee, organization, and requester fields. This work expands that functionality to support all field types (except for multi-selects, support for which will come at a later date).

## Screenshots
Screenshot 2025-06-02 at 9 52 11 AM

## Checklist


src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -14,6 +14,30 @@ import debounce from "lodash.debounce";
...
+  filters: Filter[],
+  fields: Field[] = []
+) {
+  return (
+    filters &&
+    filters.reduce<string[]>((acc, filter) => {
+      const parsedFilterId = filter.value.split("ticket_fields_")[1];
+
+      const field = fields.find((field) => {
+        const parsedFieldId = field.name
```
Fredx87: There is no need to extract the `id` from the `name` in this way, the `id` on `Field` already contains that. The `fields` array also contains standard fields (subject, description, and eventually type and priority). How about these fields? Are they supported as conditions?

src/modules/new-request-form/data-types/Field.ts
```diff
@@ -1,3 +1,14 @@
+export interface Filter {
```
Fredx87: The name `Filter` is too generic, and eventually can become confusing in the future.

I am not sure about the naming, but it feels to me that this one is more of a `RelationshipFilter` and the following one is a collection of different filters for a lookup field, so maybe something like `LookupFieldRelationshipFilters`?)

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -54,6 +86,19 @@ export function LookupField({
   const [isLoadingOptions, setIsLoadingOptions] = useState<boolean>(false);
   const { t } = useTranslation();
 
+  const dynamicFilters = field.relationship_filter
```
Fredx87: `dynamicFilters` is used only inside `buildAdvancedDynamicFilterParams`, there is no need to calculate it here on every re-render, it can be moved to the function itself

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -103,12 +148,17 @@ export function LookupField({
...
+        visibleFields
+      );
+
       if (organizationId !== null)
         searchParams.set("organization_id", organizationId);
       setIsLoadingOptions(true);
       try {
         const response = await fetch(
-          `/api/v2/custom_objects/${customObjectKey}/records/autocomplete?${searchParams.toString()}`
+          `/api/v2/custom_objects/${customObjectKey}/records/autocomplete?${searchParams.toString()}&${dynamicFilterParams}}`
```
Fredx87: We are using `URLSearchParams` to collect the various params to pass to the autocomplete endpoints and then appending them to the URL. I feel it would be better to continue to use the same approach also for these dynamic filter parameters instead of getting a string from `buildAdvancedDynamicFilterParams` and appending it at the end

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -136,7 +186,14 @@ export function LookupField({
         setIsLoadingOptions(false);
       }
     },
-    [customObjectKey, fieldId, organizationId, selectedOption, userId]
+    [
```
Fredx87, line 186: ESLint reports that there is a missing dependency in the array

src/modules/new-request-form/fields/helpers.test.ts
```diff
@@ -0,0 +1,101 @@
+import type { Field, Filter } from "../data-types";
```
Fredx87: It's a good thing you added unit tests. Unfortunately, we don't have that many right now, but in general, the convention is to have the test file named `[SOURCE_FILE].test.ts`. The name `helpers.test.ts` doesn't make it clear where the implementation leaves.

src/modules/new-request-form/fields/helpers.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { buildAdvancedDynamicFilterParams } from "./LookupField";
+describe("buildAdvancedDynamicFilterParams", () => {
+  it("builds dynamic filter params correctly when fields match filters", () => {
+    const filters: Filter[] = [
+      {
+        field: "some_field",
+        operator: "equals",
```
Fredx87: This doesn't seem aligned with the implementation, where the used operators are `matches` and `not_matches`

src/modules/new-request-form/fields/helpers.test.ts
```diff
@@ -0,0 +1,101 @@
+import type { Field, Filter } from "../data-types";
+import { buildAdvancedDynamicFilterParams } from "./LookupField";
+describe("buildAdvancedDynamicFilterParams", () => {
+  it("builds dynamic filter params correctly when fields match filters", () => {
```
Fredx87: What field types are supported by the dynamic filter feature? Is it just text fields or any field type? 
If more field types are supported, you may want to test how they behave (either manually or by adding more unit tests), since the value is stored differently for each field type

src/modules/new-request-form/fields/helpers.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        description: "",
+        type: "text",
+        options: [],
+      },
+    ];
+
+    const result = buildAdvancedDynamicFilterParams(filters, fields);
+
+    expect(result).toEqual([
+      "filter[dynamic_values][ticket_fields_0000]=undefined",
```
Fredx87: I am not familiar with the API, is this correct? If a field is missing on the form we need to pass `undefined` as a value? I would expect not to pass anything

Fredx87 (commented):
I left some comments.

Can we also add more info on the PR description on what this is about and how it works? I had some struggles understanding the changes due to missing context.

This is also an open-source repo used as a reference to build custom themes. I think more context can be useful also for third-party developers

src/modules/new-request-form/fields/LookupField.tsx
akuttruff: Good call, I overlooked this. Addressed [here](https://github.com/zendesk/copenhagen_theme/pull/642/commits/43895a0a5905035fdd4cc6f429a2c878c9c3f10c)

src/modules/new-request-form/data-types/Field.ts
akuttruff: Agreed -- I've [adjusted the types](https://github.com/zendesk/copenhagen_theme/pull/642/commits/43895a0a5905035fdd4cc6f429a2c878c9c3f10c) to `LookupRelationshipFieldFilter` and `FilterCondition`

src/modules/new-request-form/fields/LookupField.tsx
akuttruff: Addressed [here](https://github.com/zendesk/copenhagen_theme/pull/642/commits/43895a0a5905035fdd4cc6f429a2c878c9c3f10c)

src/modules/new-request-form/fields/LookupField.tsx
akuttruff: Addressed [here](https://github.com/zendesk/copenhagen_theme/pull/642/commits/43895a0a5905035fdd4cc6f429a2c878c9c3f10c)

src/modules/new-request-form/fields/helpers.test.ts
akuttruff: I've updated the test data

src/modules/new-request-form/fields/helpers.test.ts
akuttruff: Since all field values are expressed as strings and stored as such regardless of field type (checkbox values are passed as `true`), this test coverage should suffice.

src/modules/new-request-form/fields/helpers.test.ts
akuttruff: I've pivoted to use your `SearchParams` pattern, and will only pass the query params when the field value is truthy.

akuttruff:
@Fredx87 Thanks for the thorough review, I appreciate it. I've pushed updates to address your comments -- please let me know if I've missed anything.

src/modules/new-request-form/fields/LookupField.tsx
```diff
@@ -8,12 +8,45 @@ import {
...
+  const parsedFilterId =
+    dynamicFilters && dynamicFilters[0]?.value.split("ticket_fields_")[1];
+
+  const field = fields.find((field) => {
+    const parsedFieldId = field.id.toString();
+
+    return parsedFieldId === parsedFilterId;
+  });
+
+  return [dynamicFilters[0]?.value, field?.value];
```
Fredx87, line 47: With this implementation, at most one dynamic filter will be returned. Is this correct? I think the previous implementation was returning all matching filters, but I am not completely familiar with how the feature should work.

src/modules/new-request-form/fields/helpers.test.ts
Fredx87: That's not completely true: https://github.com/zendesk/copenhagen_theme/blob/6b15e6632367bc556fdebf017c0aebe00864c047/src/modules/new-request-form/data-types/Field.ts#L4. In case of multi-select fields, the value is an array of strings.

If the feature doesn't support this field type, it probably should work, considering everything else as a string. If you need to support this field type, you need to pass the value in the get in some way (maybe joining the values with a `,`?). 

The multiselect field also [is not updating ](https://github.com/zendesk/copenhagen_theme/blob/6b15e6632367bc556fdebf017c0aebe00864c047/src/modules/new-request-form/NewRequestForm.tsx#L310)the React state on change, so you would need to fix that as well. This is already fixed on the [beta](https://github.com/zendesk/copenhagen_theme/blob/d9b666b335544bcc1dc2d6cc36bbc8d8cf498b60/src/modules/ticket-fields/TicketField.tsx#L89) branch but not integrated in the master branch yet

akuttruff: @Fredx87 thanks for pointing that out, but dynamic filters for multiselects is out of scope for this piece of work

src/modules/new-request-form/fields/LookupField.tsx
akuttruff: Correct -- since we'll only be returning one dynamic filter per field, I simplified the logic here.

akuttruff:
@Fredx87 thanks again -- I've squashed the commits for clarity and responded to your questions about multi select fields. I'll have a PR up shortly for the v1-v3 implementation. Are you okay if I merge this work in? I'd like to get your approval on it, since you've raised some concerns.

src/modules/new-request-form/fields/helpers.test.ts
akuttruff: We have yet to implement support for that on the backend, but once that work is complete, there will be a follow-up story to implement on the frontend.

Fredx87 (approved):

PR #645, opened 2025-06-13
Refactor Service Catalog folder structure and namings

Fredx87:
## Description

This PR brings some refactoring to the `modules/service-catalog` and `modules/ticket-fields` folders.

### Commit 45ea1aa824b978e56236eed10e911a422b0ae7d7
The current folder structure for Service Catalog is a bit confusing and difficult to work with. Components, hooks, and functions related to the Service List page and the Service Item page are placed around without a particular structure, making it difficult to understand where they are used.

This commit consolidates the structure, putting all the things related to the list page in a `service-catalog-list` folder and all the things related to the service item page in a `service-catalog-item` folder. Both folders contain subfolders for components and hooks. The commit contains mainly file renamings and changes in the imports.

### Commit 3d5ac5ca30b42fa248f9abeae27ab7f04bc215b7
Currently, we have two different types for ticket fields:
- A [`Field`](https://github.com/zendesk/copenhagen_theme/blob/8a650cc681c734722a27902235923701f56340fa/src/modules/ticket-fields/data-types/Field.ts) type in the `ticket-fields` module
- A [`TicketField`](https://github.com/zendesk/copenhagen_theme/blob/8a650cc681c734722a27902235923701f56340fa/src/modules/service-catalog/data-types/TicketField.ts) type in the `service-catalog` module

Moreover, we also have a [`TicketField`](https://github.com/zendesk/copenhagen_theme/blob/8a650cc681c734722a27902235923701f56340fa/src/modules/ticket-fields/TicketField.tsx) component that has the same name as the type, but it is a completely different thing.

This is very confusing. We need different types because in the New Request Form we fetch the ticket fields from Curlybars, while in the Service Catalog Form we fetch them from public APIs. This is not clear from the code, so I made some changes:
- Renamed the `Field` type to `TicketFieldObject` (as it is called in the documentation) and added a link to the documentation
- Moved the `TicketField` type to the `ticket-fields` module and added a comment
- Added a `BaseTicketField` type with common fields
- Renamed the `TicketField` component to `RequestFormField`, since it is used in the New Request Form and in the Service Request Form

### Commit aecf77e95ddea95b09ed9ee13da0837547bc76d3
In #642, a new feature has been added to lookup fields. This feature uses the `relationship_filter` property of the ticket field, and we need to handle it in the Service Catalog form

## Checklist

### DIFF
--- a/src/modules/ticket-fields/data-types/TicketFieldObject.ts
+++ b/src/modules/ticket-fields/data-types/TicketFieldObject.ts
@@ -12,10 +12,10 @@ export interface TicketFieldObject extends BaseTicketField {
   error: string | null;
   label: string;
   required: boolean;
-  options: FieldOption[];
+  options: TicketFieldOptionObject[];
 }
 
-export interface FieldOption {
+export interface TicketFieldOptionObject {
   name: string;
   value: string;
 }
--- a/src/modules/ticket-fields/fields/LookupField.tsx
+++ b/src/modules/ticket-fields/fields/LookupField.tsx
@@ -10,7 +10,7 @@ import {
 import { useCallback, useEffect, useMemo, useState } from "react";
 import type {
   TicketFieldObject,
-  FieldOption,
+  TicketFieldOptionObject,
 } from "../data-types/TicketFieldObject";
 import { Span } from "@zendeskgarden/react-typography";
 import debounce from "lodash.debounce";
@@ -82,10 +82,9 @@ export function LookupField({
     relationship_target_type,
   } = field;
 
-  const [options, setOptions] = useState<FieldOption[]>([]);
-  const [selectedOption, setSelectedOption] = useState<FieldOption | null>(
-    null
-  );
+  const [options, setOptions] = useState<TicketFieldOptionObject[]>([]);
+  const [selectedOption, setSelectedOption] =
+    useState<TicketFieldOptionObject | null>(null);
   const [inputValue, setInputValue] = useState<string>(value as string);
   const [isLoadingOptions, setIsLoadingOptions] = useState<boolean>(false);
   const { t } = useTranslation();
@@ -168,7 +167,8 @@ export function LookupField({
           );
           if (selectedOption) {
             fetchedOptions = fetchedOptions.filter(
-              (option: FieldOption) => option.value !== selectedOption.value
+              (option: TicketFieldOptionObject) =>
+                option.value !== selectedOption.value
             );
             fetchedOptions = [selectedOption, ...fetchedOptions];
           }
--- a/src/modules/ticket-fields/fields/useNestedOptions.tsx
+++ b/src/modules/ticket-fields/fields/useNestedOptions.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useMemo, useState } from "react";
 import { type IOptionProps } from "@zendeskgarden/react-dropdowns.next";
-import type { FieldOption } from "../data-types/TicketFieldObject";
+import type { TicketFieldOptionObject } from "../data-types/TicketFieldObject";
 
 interface OptionProps {
   value: string;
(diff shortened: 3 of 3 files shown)


src/modules/ticket-fields/data-types/TicketFieldObject.ts
```diff
@@ -0,0 +1,21 @@
...
+ * available in the New Request Page template.
+ */
+export interface TicketFieldObject extends BaseTicketField {
+  id: number;
+  name: string;
+  value: string | string[] | boolean | null;
+  error: string | null;
+  label: string;
+  required: boolean;
+  options: FieldOption[];
```
melzreal: would this also account for empty options? 

Maybe unrelated to this representation but I was playing around with the api and options is only present on a dropdown type, so for ex this dropdown:
https://z3nmelsku-test.zendesk-staging.com/api/v2/ticket_fields/8722747260797
responds to an options call:
https://z3nmelsku-test.zendesk-staging.com/api/v2/ticket_fields/8722747260797/options

but a ticket field of another type, say a subject, returns"not found" when hitting the options endpoint
https://z3nmelsku-test.zendesk-staging.com/api/v2/ticket_fields/8504240744701/options

so im not sure thats a case we should also account for, test account [here](https://pandora-v2.zende.sk/resources/accounts/308999)

melzreal (approved):
definitely much better naming then Field 🙆‍♀️ ✨

src/modules/ticket-fields/data-types/TicketFieldObject.ts
Fredx87: This type is for the [ticket field object](https://developer.zendesk.com/api-reference/help_center/help-center-templates/objects/#ticket-field-object) returned in Curlybars as reported in the comment, so it is different from what we return in the APIs. Checking what we pass from the template, I can see that `options` can be `null` for most of the ticket fields.

In theory, a better type would be declaring a specific type for dropdowns and multiselect, where `options` is an array of `FieldOptions` and `null` for other types, but I didn't want to change too many things in this PR. I'll check if I can add the `| null` without causing the addition of too many checks around

Fredx87:
> In the service-catalog app we're repeating the "components" folder down through the folder structure. This is at odds with what we do in the "approval-requests" folder so maybe a bit confusing for people that clone the repo and work across apps. In other apps at Zendesk we most often stick to a single parent level `components` folder so it could be nice to align with that (?) And also maybe something we could consider for the `new-request-form` app in the future (?)

My reasoning for that is that the `service-catalog-list` and `service-catalog-item` are in fact two different "apps" rendered independently on two different pages, so it made sense to divide them in the folder structure as well.
The `approval-requests` folder currently has the same confusing structure that we have in the `service-catalog` app, probably they just followed the same pattern.

I am not a big fan of these top-level `components` and `hooks` folders, where we end up mixing multiple unrelated things together by type. But it is true that what we do in most repos, and also in repos where things are rendered in different views, and it is better to be consistent. I will give it a try to see if I can come up with a structure that is more aligned

luis-almeida:
> My reasoning for that is that the service-catalog-list and service-catalog-item are in fact two different "apps" rendered independently on two different pages, so it made sense to divide them in the folder structure as well.
The approval-requests folder currently has the same confusing structure that we have in the service-catalog app, probably they just followed the same pattern.

That makes sense and I think it's the single `service-catalog` bundle that could arguably make it "appear" as if it was just one app. I don't think we'd want to split that bundle though.

> I am not a big fan of these top-level components and hooks folders, where we end up mixing multiple unrelated things together by type. But it is true that what we do in most repos, and also in repos where things are rendered in different views, and it is better to be consistent. I will give it a try to see if I can come up with a structure that is more aligned

Yeah I'm not a fan of having a single top level `components` and `hooks` folder either. I think it's easier to navigate when the folder tree reflects the app/component structure. However, since we're constantly jumping from app to app I think it would make a lot of sense to try and keep it consistent.

src/modules/ticket-fields/data-types/TicketFieldObject.ts
Fredx87: I tried to extend the type with `| null`, but it will require us to add a lot of null checks around. As mentioned, it would be better to define the type in a better way, where `options` is available or not based on the ticket field type. But I think that this can be eventually addressed later in a separate PR

PR #647, opened 2025-06-27
GG-4255: fix: Use colours from team settings in the services page

dmuneras:
## Description

Use the colours from the team settings in the services list item cards. The main changes are:

- We will use the brand colour (`primaryHue`) as the border colour on hover and maintain the same.
- Use the text color (`foreground `) as the text colour for the cards.
- Updates `nwsapi` to fix a problem preventing the tests for on hover user events to work when implemented using [userEvent](https://github.com/testing-library/user-event). You can check [this issue](https://github.com/testing-library/jest-dom/issues/594).

## Video

https://github.com/user-attachments/assets/344bdbda-259f-4332-ac7d-f8bf87f01718

## Checklist

[GG-4255]: https://zendesk.atlassian.net/browse/GG-4255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,8 @@
     "wait-on": "8.0.3"
   },
   "resolutions": {
-    "@types/react": "^17.x"
+    "@types/react": "^17.x",
+    "nwsapi": "^2.2.20"
   },
   "commitlint": {
     "extends": [
--- a/src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
+++ b/src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
@@ -90,8 +90,6 @@ describe("ServiceCatalogListItem", () => {
       const itemTitle = screen.getByTestId("service-catalog-list-item");
       const defaultBorderColor = "#d8dcde";
 
-      await user.unhover(itemTitle);
-
       expect(itemTitle).toHaveStyle(`border-color: ${defaultBorderColor}`);
 
       await user.hover(itemTitle);
--- a/src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
+++ b/src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
@@ -1,0 +1,102 @@
+import { render, screen } from "@testing-library/react";
+import { ThemeProvider } from "styled-components";
+import { DEFAULT_THEME } from "@zendeskgarden/react-theming";
+import ServiceCatalogListItem from "./ServiceCatalogListItem";
+import type { ServiceCatalogItem } from "../../data-types/ServiceCatalogItem";
+import { userEvent } from "@testing-library/user-event";
+
+jest.mock("@zendeskgarden/svg-icons/src/16/shapes-fill.svg", () => {
+  return function MockShapesIcon() {
+    return <div data-testid="atl-nacional" />;
+  };
+});
+
+const theme = {
+  ...DEFAULT_THEME,
+  colors: {
+    ...DEFAULT_THEME.colors,
+    foreground: "#ff0000",
+    primaryHue: "#0000ff",
+  },
+};
+
+const renderWithTheme = (component: React.ReactElement) => {
+  return render(<ThemeProvider theme={theme}>{component}</ThemeProvider>);
+};
+
+describe("ServiceCatalogListItem", () => {
+  const mockServiceItem: ServiceCatalogItem = {
+    id: 1989,
+    name: "Atl Nacional keyboard",
+    description: "This is a keyboard from Atl Nacional",
+    form_id: 456,
+  };
+
+  const mockHelpCenterPath = "/hc/en-us";
+
+  beforeEach(() => jest.clearAllMocks());
+
+  describe("Rendering", () => {
+    it("should render the service item with correct content", () => {
+      renderWithTheme(
+        <ServiceCatalogListItem
+          serviceItem={mockServiceItem}
+          helpCenterPath={mockHelpCenterPath}
+        />
+      );
+
+      expect(screen.getByText(mockServiceItem.name)).toBeInTheDocument();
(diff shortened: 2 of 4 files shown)


gosiexon-zen (approved):

package.json
```diff
@@ -109,7 +109,8 @@
     "wait-on": "8.0.3"
   },
   "resolutions": {
-    "@types/react": "^17.x"
+    "@types/react": "^17.x",
+    "nwsapi": "^2.2.20"
```
BrunoBFerreira, line 113: I don't know how you added the resolution but this might be relevant: https://github.com/zendesk/guide-app-framework/pull/476#pullrequestreview-2868645490

BrunoBFerreira (commented):
Overall looks good. Thank you for adding the tests. 

I will just ask you to remove the JIRA links from the description considering this is an open source repo. 

Otherwise looks good 👍

package.json
dmuneras: I will check that one out. Thanks 🙏

dmuneras: Hi @BrunoBFerreira 

I looked at it and it looks like I can't use the yarn resolution command without updating the version of yarn used to generate the lock file.

BrunoBFerreira: That is fine. Just wanted to be sure 😄 thanks for confirming.

src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { DEFAULT_THEME } from "@zendeskgarden/react-theming";
+import ServiceCatalogListItem from "./ServiceCatalogListItem";
+import type { ServiceCatalogItem } from "../../data-types/ServiceCatalogItem";
+import { userEvent } from "@testing-library/user-event";
+
+jest.mock("@zendeskgarden/svg-icons/src/16/shapes-fill.svg", () => {
+  return function MockShapesIcon() {
+    return <div data-testid="atl-nacional" />;
+  };
+});
```
Fredx87, line 12: Usually, we don't mock sub-components in tests, unless there is a good reason.

Maybe we can instead do `<ShapesIcon data-test-id="service-icon" />` in the component and check that. I am not sure it works. If it doesn't, we can keep the mock

src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+      renderWithTheme(
+        <ServiceCatalogListItem
+          serviceItem={mockServiceItem}
+          helpCenterPath={mockHelpCenterPath}
+        />
+      );
+
+      const user = userEvent.setup();
+      const itemTitle = screen.getByTestId("service-catalog-list-item");
+      const defaultBorderColor = "#d8dcde";
```
Fredx87: We can read this value from `DEFAULT_THEME.colors.grey[300]` instead of hardcoding it here

src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    it("should use the theme foreground color as text color", () => {
+      renderWithTheme(
+        <ServiceCatalogListItem
+          serviceItem={mockServiceItem}
+          helpCenterPath={mockHelpCenterPath}
+        />
+      );
+
+      const itemTitle = screen.getByTestId("service-catalog-list-item");
```
Fredx87: Nit: this is not a title but something else, maybe a container? or a tile?

src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
dmuneras: Hi @Fredx87,

I am not sure if it is possible based on [this docs](https://zendeskgarden-v8.netlify.app/components/theme-object#colors).

I could get it like this: `theme?.palette?.grey?.[300]` this is [the ref](https://garden.zendesk.com/components/palette).

Fredx87: Yeah it should be `DEFAULT_THEME.palette.grey?.[300]`

src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
dmuneras: Hi @Fredx87 

I tried , but it get an error with the SVG import in the component. It looks like we have an extra Jest config to handle non javascript imports.

Fredx87: Yeah, that's true. Maybe it would be better to change the base jest config, adding support for SVGs as we do in other repos, to avoid the need to mock SVGs like this every time. But it is beyond the scope of this PR so we can eventually address it later

src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+      renderWithTheme(
+        <ServiceCatalogListItem
+          serviceItem={mockServiceItem}
+          helpCenterPath={mockHelpCenterPath}
+        />
+      );
+
+      const user = userEvent.setup();
+      const itemTitleContainer = screen.getByTestId("service-catalog-list-item-container");
+      const defaultBorderColor = DEFAULT_THEME?.palette?.grey?.[300];
```
Fredx87: Nit: there is no need to put all these `?`, only `grey` can be undefined

Fredx87 (approved):

PR #650, opened 2025-07-02
feat: display thumbnails in the services item list

zendesk-maciej-lorens:
## Description

Display thunbnails in the Service list page - https://zendesk.atlassian.net/browse/GG-4327
Display thunbnails in the Service page - https://zendesk.atlassian.net/browse/GG-4328

I joined both tickets as they are very similar, using same components and BL. 

### Design

Index:
image

Show:
image

### Before

Index:
image

Show:
image

### After

Index:
image

Show with thumbnail:
image

Show with fallback icon:
image

## Checklist

### DIFF
--- a/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
@@ -6,8 +6,7 @@ import ChevronDown from "@zendeskgarden/svg-icons/src/16/chevron-down-fill.svg";
 import { useTranslation } from "react-i18next";
 import { getColorV8 } from "@zendeskgarden/react-theming";
 import { XXXL } from "@zendeskgarden/react-typography";
-import { Avatar } from "@zendeskgarden/react-avatars";
-import ShapesIcon from "@zendeskgarden/svg-icons/src/16/shapes-fill.svg";
+import ItemThumbnail from "../service-catalog-list/ItemThumbnail";
 
 const DescriptionWrapper = styled.div`
   border-bottom: ${(props) => props.theme.borders.sm}
@@ -54,18 +53,6 @@ const HeaderContainer = styled.div`
   gap: ${(props) => props.theme.space.md};
 `;
 
-const StyledAvatar = styled(Avatar)`
-  width: 72px !important;
-  height: 72px !important;
-  background-color: ${(props) => getColorV8("grey", 100, props.theme)};
-
-  && > svg {
-    width: 28px;
-    height: 28px;
-    color: ${(props) => getColorV8("grey", 600, props.theme)};
-  }
-`;
-
 interface CollapsibleDescriptionProps {
   title: string;
   description: string;
@@ -91,13 +78,7 @@ export const CollapsibleDescription = ({
   return (
     <DescriptionWrapper>
       <HeaderContainer>
-        <StyledAvatar isSystem>
-          {thumbnailUrl ? (
-            <img src={thumbnailUrl} alt="" />
-          ) : (
-            <ShapesIcon aria-hidden="true" />
-          )}
-        </StyledAvatar>
+        <ItemThumbnail size="large" url={thumbnailUrl} />
         <ItemTitle tag="h1">{title}</ItemTitle>
       </HeaderContainer>
       {description && (
--- a/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
@@ -6,7 +6,7 @@ import ChevronDown from "@zendeskgarden/svg-icons/src/16/chevron-down-fill.svg";
 import { useTranslation } from "react-i18next";
 import { getColorV8 } from "@zendeskgarden/react-theming";
 import { XXXL } from "@zendeskgarden/react-typography";
-import ItemThumbnail from "../service-catalog-list/ItemThumbnail";
+import { ItemThumbnail } from "../item-thumbnail/ItemThumbnail";
 
 const DescriptionWrapper = styled.div`
   border-bottom: ${(props) => props.theme.borders.sm}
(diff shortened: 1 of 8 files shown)


src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.tsx
```diff
@@ -51,14 +52,15 @@ const TextContainer = styled.div`
...
-  width: ${(props) => props.theme.space.xl};
-  height: ${(props) => props.theme.space.xl};
-  text-align: center;
-  align-content: center;
+`;
+
+const StyledShapesIcon = styled(ShapesIcon)`
+  height: 16px !important;
+  width: 16px !important;
+  color: ${(props) => getColorV8("grey", 600, props.theme)} !important;
```
Fredx87: Nit: we could also do, inside the `StyledAvatar` styled-components:
```
  && > svg {
    width: 16px;
    height: 16px;
    color: ${(props) => getColorV8("grey", 600, props.theme)};
  }
```

This would allow us to avoid the usage of `!important`, but it is a minor thing, and we need to override the styles anyway, so feel free to ignore

zendesk-maciej-lorens: I was looking for something that removes this ugly !important. Thanks!

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -44,16 +47,37 @@ const ToggleButton = styled(Button)`
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: ${(props) => props.theme.space.md};
+`;
+
+const StyledAvatar = styled(Avatar)`
+  width: 72px !important;
+  height: 72px !important;
+  background-color: ${(props) => getColorV8("grey", 100, props.theme)};
```
gosiexon-zen, line 60: We need to take color from theme like `${(props) => props.theme.colors.....` because getColorV8 doesn't work with theme customization and Daniel recently fixed it in other places.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -44,16 +47,37 @@ const ToggleButton = styled(Button)`
...
 
+const HeaderContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: ${(props) => props.theme.space.md};
+`;
+
+const StyledAvatar = styled(Avatar)`
+  width: 72px !important;
```
gosiexon-zen: why we need important?

gosiexon-zen: Ok I see that you agreed that we need to override styles and this part
`&& > svg {
    width: 16px;
    height: 16px;
    color: ${(props) => getColorV8("grey", 600, props.theme)};
  }`
  allow us to remove important so maybe it is not necessary now?

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -66,7 +90,16 @@ export const CollapsibleDescription = ({
...
     <DescriptionWrapper>
-      <ItemTitle tag="h1">{title}</ItemTitle>
+      <HeaderContainer>
+        <StyledAvatar isSystem>
+          {thumbnailUrl ? (
+            <img src={thumbnailUrl} alt="" />
+          ) : (
+            <ShapesIcon aria-hidden="true" />
+          )}
+        </StyledAvatar>
```
Fredx87: We have two different components that are doing basically the same thing: checking if the service has a thumbnail and showing it inside the Avatar component, or using the default shapes icon. The only difference is the size.

I think it would be better to create a new `ServiceThumbnail` component with a size prop (small, large), and reuse it in both places

zendesk-maciej-lorens: done!

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
zendesk-maciej-lorens: unfortunately Avatar has width and height with !Important, so I have to use it here also.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -6,6 +6,7 @@ import ChevronDown from "@zendeskgarden/svg-icons/src/16/chevron-down-fill.svg";
 import { useTranslation } from "react-i18next";
 import { getColorV8 } from "@zendeskgarden/react-theming";
 import { XXXL } from "@zendeskgarden/react-typography";
+import ItemThumbnail from "../service-catalog-list/ItemThumbnail";
```
Fredx87: The `ItemThumbnail` component is reused both in the `ServiceCatalogList` and `ServiceCatalogItem` component, it is better to put it in a new folder under components, like `components/item-thumbnail`.

Moreover, we usually don't use default exports, so it is better to export it as a named variable for coherence

src/modules/service-catalog/components/service-catalog-list/ItemThumbnail.tsx
```diff
@@ -0,0 +1,31 @@
+import ShapesIcon from "@zendeskgarden/svg-icons/src/16/shapes-fill.svg";
+import styled from "styled-components";
+import { getColorV8 } from "@zendeskgarden/react-theming";
+import { Avatar } from "@zendeskgarden/react-avatars";
+
+const StyledAvatar = styled(Avatar)<{ size: string }>`
```
Fredx87: We can define the `size` prop as `"medium" | "large"` here as well, instead of a simple string

Fredx87 (commented):
Looks good!

I left a couple of small comments

PR #651, opened 2025-07-02
feat: add support for end user conditions in SC item page

BrunoBFerreira:
## Description

In the near future it will be possible to define End User Conditions in Service Catalog Forms. We use these conditions to determine which fields should be visible in the Request Forms, but we were not checking these for Service Catalog forms yet, so it was not possible to hide fields in the Service Catalog Item page.

This PR is adding support to have these conditions in the Service Catalog Item page. It moves the `getVisibleFields` logic to the `ticket-fields` module, given this will be now more widely used, and updates the logic that fetches the ticket fields for the Service Catalog Item page to check and filter with the `endUserConditions`. 

## Checklist


src/modules/service-catalog/hooks/useItemFormFields.tsx
```diff
@@ -105,10 +110,14 @@ const fetchTicketFields = async (
       return null;
     })
     .filter(Boolean);
+
   if (!associatedLookupField) {
     throw new Error("Associated lookup field not found");
   }
-  return { requestFields, associatedLookupField };
+
+  const visibleFields = getVisibleFields(requestFields, endUserConditions);
```
Fredx87: This is not going to work properly. We need to evaluate the visible fields every time a value of a field changes, computing the visible fields in the `fetchTicketFields` means that they are evaluated only at the first mount.
They also don't have any value set: the ticket fields API just returns a list of ticket fields, and the value is set by the user in the form

src/modules/ticket-fields/getVisibleFields.tsx
```diff
@@ -1,5 +1,5 @@
-import type { TicketFieldObject } from "../ticket-fields/data-types/TicketFieldObject";
-import type { EndUserCondition } from "./data-types";
+import type { TicketFieldObject } from "./data-types/TicketFieldObject";
+import type { EndUserCondition } from "../new-request-form/data-types/EndUserCondition";
```
Fredx87: Let's move the `EndUserCondition` type to the `ticket-fields` module as well. It is a bit strange that modules import the `getVisibleFields` function from the `ticket-fields` module and the function imports the type from the `new-request-form` module

src/modules/service-catalog/hooks/useItemFormFields.spec.ts
```diff
@@ -392,4 +392,92 @@ describe("useItemFormFields", () => {
       },
     ]);
   });
+
+  it("should apply conditional visibility to fields based on endUserConditions", async () => {
```
Fredx87: Do we really need such a test for this hook? We are basically re-testing what has already been tested in the `getVisibleFields` function. The test scenario is also wrong, since as I explained in the other comment the ticket fields API doesn't return any value for the fields.

This is more an integration test then an unit test, maybe it is worth having it, but I am not sure

src/modules/service-catalog/hooks/useItemFormFields.tsx
BrunoBFerreira: Ah you are right. I missed this 🙈 I will review the implementation.

BrunoBFerreira: I have updated the approach. However, I am also filtering in the first mount given that dropdowns might have default values that affect visibility. 

Let me know what you think.

src/modules/new-request-form/data-types/index.ts
```diff
@@ -1,5 +1,5 @@
 export * from "./RequestForm";
-export * from "./EndUserCondition";
+export * from "../../ticket-fields/data-types/EndUserCondition";
```
Fredx87: I think we can remove this export from this file, we shouldn't need it

src/modules/service-catalog/hooks/useItemFormFields.tsx
```diff
@@ -105,28 +112,38 @@ const fetchTicketFields = async (
...
+          const { requestFields, associatedLookupField, endUserConditions } =
             await fetchTicketFields(serviceCatalogItem.form_id, baseLocale);
-          setRequestFields(requestFields);
           setAssociatedLookupField(associatedLookupField);
+          setEndUserConditions(endUserConditions);
+          const filteredFields = getVisibleFields(
+            requestFields,
+            endUserConditions
+          );
+          setRequestFields(filteredFields);
```
Fredx87: We should use the same approach used in the NewRequestForm: https://github.com/zendesk/copenhagen_theme/blob/31a39ae21c1d6858b0e66ec7273185a5389609e6/src/modules/new-request-form/NewRequestForm.tsx#L109

The `requestFields` state should contain all fields, then we apply the conditions and use these fields in the rendering, instead of updating the state in the `useEffects`.

gosiexon-zen: Changed that, let me know if it is better now.

src/modules/service-catalog/hooks/useItemFormFields.tsx
```diff
@@ -138,22 +154,27 @@ export function useItemFormFields(
...
+          : ticketField
       );
+
+      setAllRequestFields(updatedFields);
     },
-    [requestFields]
+    [allRequestFields]
+  );
+
+  const requestFields = useMemo(
```
Fredx87: This works, but maybe `useMemo` is overkill in this case. I would keep it simple and coherent with the [implementation](https://github.com/zendesk/copenhagen_theme/blob/31a39ae21c1d6858b0e66ec7273185a5389609e6/src/modules/new-request-form/NewRequestForm.tsx#L109) in the `NewRequestForm` that doesn't use `useMemo`. 

We shouldn't have that many conditions to evaluate so there shouldn't be any performance concern with not using `useMemo`

gosiexon-zen: ok, I removed `useMemo`. I was thinking about moving this line `const requestFields = getVisibleFields(allRequestFields, endUserConditions);` to ServiceCatalogItem but here in this hook we get these user conditions so maybe it is a better place.

PR #654, opened 2025-07-10
feat: wip

oorion:
## Checklist


src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -247,7 +247,9 @@ export function NewRequestForm({
             );
           }
         })}
-        {attachments_field && <Attachments field={attachments_field} />}
+        {attachments_field && (
+          <Attachments field={attachments_field} locale={locale} />
```
Fredx87: It is better to pass the `baseLocale` instead ([documented here](https://developer.zendesk.com/api-reference/help_center/help-center-templates/objects/#help_center-object))

Help Center has a [huge list of locales](https://support.zendesk.com/hc/en-us/articles/4408821324826-Zendesk-language-support-by-product#h_01JT0A0Q6D3WMTAQ53PRNMS1EW), some of which are variants of a standard locale that we have translations for. As an example, the `nb-no` (Norwegian Bokmål - Norway) locale is a variant of the `no` (Norwegian) locale. 

The uploads API in Support _should be able_ to fallback to the correct locale, but I think it is just better if we already pass the fallback locale since we have it

src/modules/new-request-form/fields/attachments/Attachments.tsx
```diff
@@ -127,6 +128,7 @@ export function Attachments({ field }: AttachmentProps): JSX.Element {
 
         const url = new URL(`${window.location.origin}/api/v2/uploads.json`);
         url.searchParams.append("filename", file.name);
+        url.searchParams.append("locale", locale);
```
Fredx87: For completeness, we should pass the `locale` as a dependency of the hook. It is never going to change, but it's better to follow the rule of hooks as reported by ESLint

src/modules/new-request-form/NewRequestForm.tsx
oorion: Thanks.  I saw both but wasn't entirely sure what the difference was.  I'll make the change.

Fredx87 (approved):

PR #658, opened 2025-07-23
fix: allow html formatting to display as text

melzreal:
## Description

- Quotes currently get displayed as &nbsp instead of the visual for quote,
using dangerouslySetInnerHTML is safe for
 this context given we are already
 processing the description via the rendering pipeline

Jira: https://zendesk.atlassian.net/browse/GG-4510

## Screenshots

Before
image

After
image

## Checklist


src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.tsx
```diff
@@ -66,7 +66,9 @@ const ServiceCatalogListItem = ({
       <ItemThumbnail size="medium" url={serviceItem.thumbnail_url} />
       <TextContainer>
         <ItemTitle>{serviceItem.name}</ItemTitle>
-        <ItemDescription>{serviceItem.description}</ItemDescription>
+        <ItemDescription
+          dangerouslySetInnerHTML={{ __html: serviceItem.description }}
```
dmuneras, line 70: If you update [this test](https://github.com/zendesk/copenhagen_theme/blob/master/src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.spec.tsx#L31) to add some HTML tags in the description, you will cover this case and prevent future regressions. 😄

dmuneras (approved):

src/modules/service-catalog/components/service-catalog-list/ServiceCatalogListItem.tsx
```diff
@@ -66,7 +66,9 @@ const ServiceCatalogListItem = ({
       <ItemThumbnail size="medium" url={serviceItem.thumbnail_url} />
       <TextContainer>
         <ItemTitle>{serviceItem.name}</ItemTitle>
-        <ItemDescription>{serviceItem.description}</ItemDescription>
+        <ItemDescription
```
BrunoBFerreira, line 69: Quick question on this: considering that we have the `description` being parsed in the rendering pipeline, shouldn't this be formatted correctly already? Just wandering if the fix should be here or before.

dmuneras: I asked the same question to Gianluca when looking at it. I think we could check the Rendering pipeline objects and do it from there. It could be a follow up 😄

BrunoBFerreira: I think it could be a good improvement yes. Lets try to keep track of it.

PR #659, opened 2025-07-25
fix: avoid unnecessary toggle

kristkvch-zen:
## Description

This PR improves the logic for collapsing long descriptions in the CollapsibleDescription component.
Previously, the ToggleButton visibility was based solely on the `description.length > DESCRIPTION_LENGTH_THRESHOLD` condition. However, this led to incorrect behavior in real scenarios:
* Even short visible descriptions (e.g. a couple of `<p>` tags) could contain long HTML content like images or attachments, inflating the `description.length` and incorrectly triggering the toggle button.
* The toggle would appear even when there was nothing to collapse.

### References

- https://zendesk.atlassian.net/browse/GG-4486

## What's changed

Replaced the static `DESCRIPTION_LENGTH_THRESHOLD` check with a dynamic calculation of actual visual overflow. The new logic uses the CSS `line-clamp` property to compare:

`scrollHeight` - the full content height

`getBoundingClientRect().height` - the visible height of the clamped container

Now toggle button is shown only when real overflow occurs making UI more accurate and intuitive.

## Checklist

### DIFF
--- a/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState, useRef } from "react";
 import styled from "styled-components";
 import { Button } from "@zendeskgarden/react-buttons";
 import ChevronUp from "@zendeskgarden/svg-icons/src/16/chevron-up-fill.svg";
@@ -24,11 +24,11 @@ const ItemTitle = styled(XXXL)`
   margin-bottom: 0;
 `;
 
-const CollapsibleText = styled.div<{ expanded: boolean }>`
+const CollapsibleText = styled.div<{ isCollapsed: boolean }>`
   font-size: ${(props) => props.theme.fontSizes.md};
   display: -webkit-box;
   -webkit-box-orient: vertical;
-  -webkit-line-clamp: ${(props) => (props.expanded ? "none" : 3)};
+  -webkit-line-clamp: ${(props) => (props.isCollapsed ? 3 : "none")};
   overflow: hidden;
   margin-top: ${(props) => props.theme.space.md};
   padding-inline-end: ${(props) => props.theme.space.xl};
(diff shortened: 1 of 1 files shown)


src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -59,20 +59,34 @@ interface CollapsibleDescriptionProps {
...
-const DESCRIPTION_LENGTH_THRESHOLD = 270;
-
 export const CollapsibleDescription = ({
   title,
   description,
   thumbnailUrl,
 }: CollapsibleDescriptionProps) => {
-  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const [isCollapsed, setIsCollapsed] = useState<boolean>(true);
+  const [isClamped, setIsClamped] = useState<boolean>(false);
```
luis-almeida, line 68: Nit: no need to to specify the type  `<boolean>` here

luis-almeida (approved):
Tested it on my account and this approach seems to work much better 👍🏼

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -59,20 +59,34 @@ interface CollapsibleDescriptionProps {
...
+      const checkClamped = () => {
+        const visibleBoxHeight = el.getBoundingClientRect().height;
+        const fullContentHeight = el.scrollHeight;
+        const clamped = fullContentHeight - visibleBoxHeight > 1;
+        setIsClamped(clamped);
+        if (!clamped) {
+          setIsCollapsed(false);
+        }
+      };
+      requestAnimationFrame(checkClamped);
```
Fredx87, line 84: Question: Why do we need `requestAnimationFrame` here?

Fredx87 (approved):
I tested it as well, and it is much better than before! 🚀 

I am just wondering if the approach with the line clamp makes sense now that we have also images and embeds

For example, if the content has one row, two big images or embeds, and then other content, the content will be clamped after the third row.
This creates a very big description even when it is not expanded:
Screenshot 2025-07-28 at 12 00 28

Maybe we could use a max-height instead, with some caveats:
- The text will not have the ellipsis
- images will be cut off in the middle when the description is not expanded

I am not sure which one is the best approach, maybe we could ask a feedback from design, but I am ok with this approach as well

kristkvch-zen:
@Fredx87 I did consider using the `max-height` approach initially, but ran into the drawback that images and other blocks can get cut off awkwardly especially with variable font sizes and multiple empty lines before/after. An added `fade-out` effect for this approach could also be an option, but that would require a more in-depth discussion with design.

There are definitely edge cases to consider: users might insert lots of images, blank lines, tables, etc. I checked the Figma design and it only covers a short text description, so I agree that this should be discussed with a designer. We need to present the full range of possible content and get guidance on how the collapsed/expanded view should behave in each case.

Also, I think a better long-term solution might be to parse the HTML rather than relying on `dangerouslySetInnerHTML`, but that’s a separate topic.

src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
kristkvch-zen: We use `requestAnimationFrame` to wait for the browser to finish rendering so we get accurate layout measurements. Without it the clamping check might run too early and give incorrect results

Fredx87: I am not 100% sure it is needed, at least trying to remove it locally didn't change anything. But it doesn't hurt to keep it, and be more sure that the measurement is right

Fredx87:
> @Fredx87 I did consider using the `max-height` approach initially, but ran into the drawback that images and other blocks can get cut off awkwardly especially with variable font sizes and multiple empty lines before/after. An added `fade-out` effect for this approach could also be an option, but that would require a more in-depth discussion with design.
> 
> There are definitely edge cases to consider: users might insert lots of images, blank lines, tables, etc. I checked the Figma design and it only covers a short text description, so I agree that this should be discussed with a designer. We need to present the full range of possible content and get guidance on how the collapsed/expanded view should behave in each case.
> 

Yeah, these are fairly good points. Since we are close to release, let's use this approach, which is better than the one we had before, and eventually we can revisit it later with the help of design 👍 

> Also, I think a better long-term solution might be to parse the HTML rather than relying on `dangerouslySetInnerHTML`, but that’s a separate topic.

Out of curiosity, what do you mean by this? The HTML is sanitized on the backend, so we `dangerouslySetInnerHTML` is safe to use here. We could eventually re-sanitize it on the client as well for an extra layer of security, if that is what you mean.

PR #661, opened 2025-07-28
fix: apply word-break to prevent overflow

kristkvch-zen:
## Description

On the Employee SC Item page, the title was not truncated, which caused layout issues when long words were used

## References

https://zendesk.atlassian.net/browse/GG-4479

## Changes

The issue was caused by long unbroken words not fitting within the container, which led to layout problems. To resolve this the `word-break: break-all` property was applied to allow words to break and wrap properly across lines.

### Before 
Screenshot 2025-07-28 at 17 39 22

### After
Screenshot 2025-07-28 at 17 39 37

## Checklist

### DIFF
--- a/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
@@ -13,7 +13,6 @@ const DescriptionWrapper = styled.div`
     ${(props) => getColorV8("grey", 300, props.theme)};
   padding-bottom: ${(props) => props.theme.space.lg};
   margin-inline-end: ${(props) => props.theme.space.xl};
-  word-break: break-all;
 
   @media (max-width: ${(props) => props.theme.breakpoints.md}) {
     margin-inline-end: 0;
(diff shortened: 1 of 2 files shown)


src/modules/service-catalog/components/service-catalog-item/CollapsibleDescription.tsx
```diff
@@ -13,6 +13,7 @@ const DescriptionWrapper = styled.div`
     ${(props) => getColorV8("grey", 300, props.theme)};
   padding-bottom: ${(props) => props.theme.space.lg};
   margin-inline-end: ${(props) => props.theme.space.xl};
+  word-break: break-all;
```
luis-almeida: Could we use `word-break: break-word;` here instead?
We seem to use it everywhere else in the theme and if I remember correctly I believe it's because it works better with asian locales (CJK)

kristkvch-zen: Thanks for the heads-up! I checked the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) and it looks like `word-break: break-word` is deprecated. Do you think we should consider replacing it?

luis-almeida: Good find! I didn't know it was marked as deprecated but it makes sense.
We should probably switch to the closest recommended replacement that also works well with asian languages.
Maybe `overflow-wrap: break-word;` ?

kristkvch-zen: Totally agree, thanks! I’ll update it here as part of this fix. If you think it makes sense I can follow up with a separate PR to update it elsewhere in the codebase for consistency. Let me know what works best!

luis-almeida (approved):

Fredx87 (approved):

PR #663, opened 2025-07-30
fix: hide deactivated Fields on the ServiceCatalog Item page

zendesk-maciej-lorens:
## Description

Deactivated CustomField is visible on Employee SC Item page

## Screenshots

before Field deactivation:

image

Field deactivation:
image

After Field deactivation:
image

## Checklist

### DIFF
--- a/src/modules/service-catalog/hooks/useItemFormFields.tsx
+++ b/src/modules/service-catalog/hooks/useItemFormFields.tsx
@@ -97,6 +97,7 @@ const fetchTicketFields = async (
         ticketField &&
         ticketField.type !== "subject" &&
         ticketField.type !== "description" &&
+        ticketField.active !== false &&
         ticketField.editable_in_portal
       ) {
         if (


src/modules/service-catalog/hooks/useItemFormFields.tsx
```diff
@@ -97,6 +97,7 @@ const fetchTicketFields = async (
         ticketField &&
         ticketField.type !== "subject" &&
         ticketField.type !== "description" &&
+        ticketField.active !== false &&
```
dmuneras: *nit:* if you are already getting a boolean value you could simple do:


```suggestion
        ticketField.active &&
```

zendesk-maciej-lorens: agreed! had to refactor some of specs for that but thats better

Fredx87 (approved):

PR #665, opened 2025-08-07
[GG-4482] Add error notification in case of service catalog request race condition

zendesk-kate-gasiorek:
## Description

This PR instructs the user to refresh the page after unsuccessful service catalog item request submission in case of a race condition where an admin edited the form with a new conditional field and the end user's view got stale.

Race condition reproduction:
1. In Admin Center create an empty service catalog item and publish it.
2. As an end user in Guide go to the newly created item.
3. In Admin Center in another tab add a new end user required field to the form.
4. Go back to the tab in Guide with stale service catalog form open and try to submit it.
5. With this PR you'll get an error notification instructing you to refresh the page.

## Screenshots

image

## Checklist


src/modules/shared/notifications/useNotify.tsx
```diff
@@ -7,17 +7,26 @@ import {
...
   const { t } = useTranslation();
 
   const notify = useCallback(
-    ({ title, message, type }: ToastNotification) => {
+    ({ title, message, type, link }: ToastNotification) => {
       addToast(({ close }) => (
         <Notification type={type}>
           {title && <Title>{title}</Title>}
           {message}
+          {link && <StyledLink href={link.href}>{link.text}</StyledLink>}
```
melzreal: quick question, was this design approved? as it looks slightly different then the usual it will need review if we're escaping the standard error notification, given we need to keep things consistent across

zendesk-kate-gasiorek: Yes it was, here is the related discussion: https://zendesk.slack.com/archives/GSHE498CE/p1754305058379989?thread_ts=1754304051.265709&cid=GSHE498CE

melzreal (approved):

src/modules/shared/notifications/useNotify.tsx
luis-almeida: Could the link be part of the message so we don't need to extend the notifications api?

zendesk-kate-gasiorek: Implemented

zendesk-kamil-zachradnik (approved):

PR #668, opened 2025-09-10
Kzachradnik/pdsc 9 overlapping notifications in copenhagen theme

zendesk-kamil-zachradnik:
## Description

This PR fixes an issue around the Copenhagen Theme where ToastProvider was being rendered in two separate DOM trees. The PR don't fix the same issue between Copenhagen and Help Center.

**Before**
- A ToastProvider was mounted once for flash notifications.
- Additional ToastProvider instances were created anywhere we used ThemeProvider in .hbs templates.
- As a result, multiple providers existed in parallel, causing errors and overlapping notifications.

**After**
We now mount a single global ToastProvider inside FlashNotifications, which is rendered in document_head.hbs — a template included on every view. Notifications can be triggered from anywhere in the application using events and hooks, but they are all displayed within the same DOM tree. This ensures consistency:
- No more duplicated or conflicting toasts.
- A single source of truth for global notifications.
- A simpler developer experience, since notifications can be dispatched from any component or view without worrying about provider scope.

## Screenshots

The notifications were triggered artificially to show the problem

**Before**

https://github.com/user-attachments/assets/aac8d0bb-deec-4f64-b063-6c47e4abc017

**After**

https://github.com/user-attachments/assets/c91ac79b-b6cd-4503-98af-1d3787f8529f

## Checklist


src/modules/shared/notifications/GlobalNotificationsProvider.tsx
```diff
@@ -0,0 +1,12 @@
+import { ToastProvider } from "@zendeskgarden/react-notifications";
+import { FlashNotifications } from "../../flash-notifications/FlashNotifications";
+import { GlobalNotifications } from "./GlobalNotifications";
+
+export function GlobalNotificationsProvider() {
+  return (
+    <ToastProvider zIndex={2147483647}>
```
kristkvch-zen: Can we use a variable here?

src/modules/shared/notifications/GlobalNotificationsProvider.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+  message: "Integration Message",
+  type: "info" as INotificationProps["type"],
+};
+
+function TestTrigger() {
+  const notify = useNotify();
+  return <button data-testid="trigger" onClick={() => notify(payload)} />;
+}
+
+describe("GlobalNotificationsProvider", () => {
```
kristkvch-zen: Great test!

src/modules/shared/notifications/GlobalNotificationsProvider.tsx
z-kzachradnik: It won't help much, we should take a look into the zIndex around the app and prepare some pattern for components and not use high indexes to avoid overlapping. I copy paste the code from the ThemeProvider so for now let's keep it like this and we should take a look into the topic sooner or later

kristkvch-zen (approved):

src/modules/flash-notifications/FlashNotifications.tsx
```diff
@@ -1,22 +1,31 @@
...
-export function FlashNotifications({
-  notifications,
-}: FlashNotificationsProps): ReactElement {
+export function FlashNotifications(): ReactElement {
   const notify = useNotify();
 
   useEffect(() => {
-    for (const notification of notifications) {
-      notify(notification);
+    const raw = window.sessionStorage.getItem(FLASH_NOTIFICATIONS_KEY);
```
gosiexon-zen, line 8: shouldn't we move try block before this line? Could sessionStorage be null or throw error? 🤔

src/modules/shared/notifications/GlobalNotificationsProvider.tsx
gosiexon-zen: Maybe we can get rid of that zIndex prop when I add new navigation for the help center navbar - I'll check that.

src/modules/shared/notifications/notifyBus.ts
```diff
@@ -0,0 +1,30 @@
+import type { Content } from "@zendeskgarden/react-notifications/dist/typings/elements/toaster/useToast";
+
+let listeners: ((render: Content) => void)[] = [];
+let queue: Content[] = [];
```
Fredx87: I think it would be better to store a `ToastNotification` in the queue and pass it to the listeners. Then the listener can transform it into a `Content` that will be passed to `addToast` with the Close button and so on.

src/modules/shared/notifications/notifyBus.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function emitNotify(render: Content): void {
+  if (listeners.length === 0) {
+    queue.push(render);
+  } else {
+    listeners.forEach((fn) => fn(render));
+  }
+}
+
+export function subscribeNotify(fn: (render: Content) => void): () => void {
```
Fredx87: I think that the fact that this function returns a function to unsubscribe is not easy to understand. It is a common pattern in the React world, but this bus is outside of React

I would prefer to have a more explicit API with two `subscribe` and `unsubscribe` methods, and then call the unsubscribe in the return of the `useEffect`

src/modules/shared/notifications/notifyBus.spec.tsx
```diff
@@ -0,0 +1,33 @@
+import { render, screen } from "@testing-library/react";
+import { emitNotify, subscribeNotify } from "./notifyBus";
+
+describe("notifyBus", () => {
```
Fredx87, line 9: I would also test the unsubscription mechanism

src/modules/shared/notifications/GlobalNotificationsProvider.tsx
```diff
@@ -0,0 +1,12 @@
+import { ToastProvider } from "@zendeskgarden/react-notifications";
+import { FlashNotifications } from "../../flash-notifications/FlashNotifications";
+import { GlobalNotifications } from "./GlobalNotifications";
+
+export function GlobalNotificationsProvider() {
```
Fredx87: I think we should use a different name, a `Provider` in React is something that provides a context for its children. This component doesn't have any children, it uses the `ToastProvider` and then it directly renders the components for the notifications

src/modules/shared/notifications/GlobalNotificationsProvider.tsx
Fredx87: We are already using this magic number in two places, also [here](https://github.com/zendesk/copenhagen_theme/blob/cd618fbd5dedb8c01faa20a975f042cdf9bc104c/src/modules/shared/garden-theme/modal-container/ModalContainerProvider.tsx#L8). 

Maybe we can take the opportunity to use a shared constant instead. I would keep the comment anyway, explaining why we need this.

The issue is that the navbar rendered in help center uses a stupidly high z-index of `2147483646`:
Screenshot 2025-09-25 at 17 25 03

So anything that needs to be rendered on top of it needs to have a higher value, which is also the maximum value accepted for z-index.

It would be very nice if we could modify the z-index in Help Center, but I am not sure if we can do that without the risk of breaking existing themes

src/modules/shared/notifications/GlobalNotifications.tsx
```diff
@@ -0,0 +1,6 @@
+import { useGlobalNotify } from "./hooks/useGlobalNotify";
+
+export const GlobalNotifications = () => {
+  useGlobalNotify();
```
Fredx87: Since this component doesn't do anything, maybe it is not worth separating the logic into a hook. We could just have the logic in the component itself

src/modules/flash-notifications/renderFlashNotifications.tsx
```diff
@@ -2,43 +2,27 @@ import { render } from "react-dom";
...
-  FLASH_NOTIFICATIONS_KEY,
   initI18next,
   loadTranslations,
 } from "../shared";
-import type { Settings, ToastNotification } from "../shared";
-import { FlashNotifications } from "./FlashNotifications";
+import type { Settings } from "../shared";
+import { GlobalNotificationsProvider } from "../shared/notifications/GlobalNotificationsProvider";
 
 export async function renderFlashNotifications(
```
Fredx87, line 17: Nit, I think the name of this function and the module is a bit misleading now:
- `flash` notification is intended to represent notifications [shown on the next page after a redirect](https://github.com/zendesk/copenhagen_theme/blob/cd618fbd5dedb8c01faa20a975f042cdf9bc104c/src/modules/shared/notifications/addFlashNotification.ts#L4) - named after the same concept in [Rails](https://api.rubyonrails.org/v7.2/classes/ActionDispatch/Flash.html) 
- This module is now responsible for rendering all notifications

That said, changing the module name and function would also require changing the document_head template: https://github.com/zendesk/copenhagen_theme/blob/0ed69bddb4892f49940d84d317d01ab0d8e95675/templates/document_head.hbs#L27-L33

This can make it more difficult for customers to update to a new version, so it is probably better if we keep the current name

src/modules/flash-notifications/renderFlashNotifications.spec.tsx
```diff
@@ -0,0 +1,36 @@
+import { cleanup, screen } from "@testing-library/react";
+import { renderFlashNotifications } from "./renderFlashNotifications";
+import { emitNotify, resetNotifyBus } from "../shared/notifications/notifyBus";
+import type { Settings } from "../shared";
+import type { Content } from "@zendeskgarden/react-notifications/dist/typings/elements/toaster/useToast";
+
+afterEach(() => {
+  cleanup();
+  document.body.innerHTML = "";
+  resetNotifyBus();
```
Fredx87: Is this reset needed? Shouldn't the component unsubscribe from the bus when unmounted?

Is this test needed at all? The function is just rendering the `GlobalNotificationsProvider` component which is already tested

Fredx87 (commented):
This PR solves a problem with the notifications in the theme, but it doesn't solve the problem that was reported initially, related to overlapping notifications from Help Center and the theme.

To be clear, this is another example that is still problematic:
Screenshot 2025-09-25 at 15 59 46

This can be triggered, for example, by deactivating Help Center, and it is caused by the fact that some notifications are emitted from the JS served by Help Center, and some are emitted from the JS in the theme. It is not an easy problem to solve, since the two implementations live in two separate worlds.

That said, we also have the problem that this PR is fixing, related to overlapping notifications in the theme. It is a problem that we needed to solve as well, so thank you for looking into it and covering the implementation with lots of tests.

The general approach is ok, I added some comments because I think it can be simplified a bit, and we can improve the naming of some things

src/modules/flash-notifications/FlashNotifications.tsx
z-kzachradnik: sessionStorage.getItem does not throw an exception — it simply returns null if the key is missing. The only place an error can actually occur is in JSON.parse, so the try/catch is only needed there. The current order (if (!raw) return; before try) is therefore correct.

src/modules/flash-notifications/renderFlashNotifications.tsx
z-kzachradnik:
```
/**
 * Note: Historically named "flash notifications" after Rails flash messages.
 * This function now renders all notifications, not only flash ones.
 * The name is kept for backward compatibility with document_head.hbs.
 */
```

I’ve added a comment to make the purpose clearer, so it’s easier to understand how this module is used. I fully understand that it’s not the ideal solution, but I agree it’s better to keep the current name for compatibility reasons.

src/modules/shared/notifications/hooks/useNotify.ts
```diff
@@ -0,0 +1,9 @@
+import type { ToastNotification } from "../ToastNotification";
+import { useCallback } from "react";
+import { emitNotify } from "../notifyBus";
+
+export const useNotify = () => {
```
Fredx87: This hook isn't doing anything other than wrapping the `emitNotify` function in a `useCallback`.

Probably at this point is just easier to use the `emitNotify` function directly instead of wrapping it in a hook. The function is defined outside of the React lifecycle, so there shouldn't be a need to wrap it inside a `useCallback`. 

I would just call it `notify` at this point and maybe export it from an `index.ts` file inside `notifications`, so then it can be imported as `import { notify } from "../shared/notifications";`

src/modules/shared/notifications/GlobalNotificationsRoot.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+  message: "Integration Message",
+  type: "info" as INotificationProps["type"],
+};
+
+function TestTrigger() {
+  const notify = useNotify();
+  return <button data-testid="trigger" onClick={() => notify(payload)} />;
+}
+
+describe("GlobalNotificationsRoot", () => {
```
Fredx87, line 16: Nit: I feel we are testing the same thing in too many places. For example, we already tested this scenario in the `GlobalNotifications` component.

This component doesn't do anything special other than wrapping two other components and doesn't have any special logic.

I would personally not test this thing here, but it's also ok to have it here as an integration test at a higher level. Just pointing out that we don't need to test every single component if there is no value in it

z-kzachradnik: We can just leave it as it is since the tests are already written and not causing any issues. Let’s just keep in mind for the future not to duplicate scenarios like this when they don’t add much value.

src/modules/shared/notifications/hooks/useNotify.ts
z-kzachradnik: I originally kept it as a hook because components were already using const notify = useNotify(), so this approach avoided touching multiple call sites and left the door open in case we needed context-aware logic in the future.
That said, your suggestion makes sense here since the hook doesn’t provide real value at the moment. I’ll refactor it to a plain notify function as you proposed.

src/modules/flash-notifications/FlashNotifications.tsx
```diff
@@ -1,22 +1,29 @@
 import type { ReactElement } from "react";
 import { useEffect } from "react";
-import type { ToastNotification } from "../shared/notifications/ToastNotification";
-import { useNotify } from "../shared/notifications/useNotify";
+import type { ToastNotification } from "../shared";
+import { notify } from "../shared";
+import { FLASH_NOTIFICATIONS_KEY } from "../shared";
```
Fredx87: There is an ESLint warning about importing from `shared/index.ts` multiple times

Fredx87 (approved):
Thank you for addressing the comments 🙏

PR #673, opened 2025-09-15
[KIAW-1468] Provide alt text for images in posts and articles 

xovnsi:
## Description

Improve accessibility by adding alt text to images in posts and articles if it's not already provided. 
Alt text is based on the image's position within the content (e.g., "Image 3 in article")

## Screenshots

image
image

## Checklist

### DIFF
--- a/script.js
+++ b/script.js
@@ -662,7 +662,6 @@
    * Adds default alt text to images that don't have one
    */
   function addAltTextToImages() {
-
     const articleBody = document.querySelector(".article-body");
     const postBody = document.querySelectorAll(".post-body");
 
@@ -674,7 +673,7 @@
 
       imagesWithoutAlt.forEach((img) => {
         const position = Array.from(allImages).indexOf(img) + 1;
-        const altText = window.articleImageAltText.replace("{{position}}", position);
+        const altText = window.articleImageAltText.replace("9999", position);
         img.setAttribute("alt", altText);
       });
     }
@@ -688,7 +687,7 @@
 
       imagesWithoutAlt.forEach((img) => {
         const position = Array.from(allImages).indexOf(img) + 1;
-        const altText = window.postImageAltText.replace("{{position}}", position);
+        const altText = window.postImageAltText.replace("9999", position);
         img.setAttribute("alt", altText);
       });
     }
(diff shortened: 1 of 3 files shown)


luis-almeida (commented):
This file is regenerated.
Please add these changes to the source files within [src](https://github.com/zendesk/copenhagen_theme/tree/master/src)

> Improve accessibility by adding alt text to images in posts and articles if it's not already provided.

So our editors ever add these alt texts? If not, should we fix those first?

luis-almeida:
Thanks for answering so quickly!

> Yes, it adds it sometimes. In the new article editor, you have the option to add alt text manually. In the old editor, some images may already have alt text embedded, possibly from metadata - I'm not sure how exactly it works

Got it! I still think it would makes sense to fix the editors as it seems like they're not generating accessible markup atm.
But I understand this would be a quicker fix.

One last question: These alt texts are being added in english only.
Isn't that a problem? Shouldn't we have translations for these?

xovnsi:
> Got it! I still think it would makes sense to fix the editors as it seems like they're not generating accessible markup atm. But I understand this would be a quicker fix.
 
Yes, I hope this approach is enough, maybe there's a better way to provide more descriptive text. 

> One last question: These alt texts are being added in english only. Isn't that a problem? Shouldn't we have translations for these?

That's right. I haven't thought of that. I'll create new translation strings. Thank you!

Fredx87:
A couple of considerations:

1. I am not sure it makes sense to implement this fix.
   - The Jira story reports the issue in the community posts page, not in the Article page
   - We are talking about content generated by the user, it should be up to the user to provide a valid alt text for the image. The editor should provide a way to define an ALT text. From my testing, it is possible to insert an alt text in articles, in community posts, but for some reason I cannot insert it for comments on posts.
   - An alt text "Image X of Y" is completely useless for the user
2. The `style` commit type is not the correct type for these type of changes
   - it [should be used](https://github.com/zendesk/copenhagen_theme?tab=readme-ov-file#commit-types) for "Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)"
   - It is not going to create a new version when the PR is merged
   - It is not going to add an entry in the changelog
   - the correct type is `fix` or `feat`

xovnsi:
@Fredx87 thank you for the feedback! 
Yeah, the changes don't really make sense. I'll flag this to accessibility team, maybe it's okay to close this issue.

PR #674, opened 2025-09-15
style: make sure code snippet is always aligned ltr

Jiyu-Zhang-Zendesk2:
## Description

Make sure code snippets are always aligned `ltr` even if we are in a `rtl` language like Arabic.

Jira: https://zendesk.atlassian.net/browse/OH-4030
## Screenshots

(incorrect)
incorrect_alignment

(correct)
correct_alignment

## Checklist

### DIFF
--- a/style.css
+++ b/style.css
@@ -1776,6 +1776,7 @@ ul {
   padding: 10px 15px;
   overflow: auto;
   white-space: pre;
+  direction: ltr;
 }
 .article-body blockquote {
   border-left: 1px solid #ddd;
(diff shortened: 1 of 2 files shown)


DariaRadu (approved):

Jiyu-Zhang-Zendesk2:
@luis-almeida I double checked and `pre` tags are always `ltr` in [the CKEditor](https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-code-block/theme/codeblock.css#L15) because `Code block are language direction–agnostic` so it doesn't seem like there should be text inside. Sune has also changed guide-markup (https://github.com/zendesk/guide-markup/pull/664) such that `pre` tags are always `ltr` there. This PR is to align HC with Guide Admin on that.

luis-almeida (approved):
@Jiyu-Zhang-Zendesk2 thanks for explaining
I can see that it is also the default direction in Support  👍🏼

PR #678, opened 2025-10-03
chore: fix theme upload script for lighthouse

Fredx87:
## Description

After upgrading to Yarn v4 in #677, the Lighthouse check started to fail, because yarn doesn't support the `--silent` flag anymore.

## Checklist

### DIFF
--- a/bin/theme-upload.js
+++ b/bin/theme-upload.js
@@ -1,9 +1,10 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
 /* eslint-env node */
 const brandId = process.env.BRAND_ID;
 const { execSync } = require("child_process");
 
 function zcli(command) {
-  const data = execSync(`yarn --silent zcli ${command} --json`);
+  const data = execSync(`yarn zcli ${command} --json`);
   return JSON.parse(data.toString());
 }
 


bin/theme-upload.js
```diff
@@ -1,9 +1,10 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
 /* eslint-env node */
 const brandId = process.env.BRAND_ID;
 const { execSync } = require("child_process");
 
 function zcli(command) {
-  const data = execSync(`yarn --silent zcli ${command} --json`);
+  const data = execSync(`yarn zcli ${command} --json`);
```
luis-almeida: I tested it locally and unfortunately without `--silent` the `--json` flag won't work with `yarn` as the "verbose" output includes more than just the json we want and it won't parse.

Maybe it's a bit weird because we use `yarn` in the repo but switching to `npx` in this script seems to work.

Fredx87: What do you mean by "the `--json` flag won't work"?

The lighthouse check is successful in this PR, and it means that the script works fine. To double-check, I ran the script locally as well and it works. I don't see any difference in the output if I use `yarn --silent zcli ...` with yarn 1, or `yarn zcli ...` or `npx zcli ...` with Yarn v4.

The only issue I see is that when something in the script fails, we don't output meaningful information. The issue was present even before the migration to Yarn v4, and I think it caused by the fact that we don't wrap the `execSync` in a try-catch

luis-almeida: Never mind! You're right I was missing a a yarn install 🤦🏼‍♂️   that's why `const data = execSync(`yarn zcli ${command} --json`);` was not returning proper json. The issue wasn't with the lack of `--silent` 👍🏼 

> The only issue I see is that when something in the script fails, we don't output meaningful information. The issue was present even before the migration to Yarn v4, and I think it caused by the fact that we don't wrap the execSync in a try-catch

That is also a good point and it would be nice to add. Just need to make sure we don't sallow the error and still break the build but with a more appropriate error.

PR #753, opened 2026-02-13
feat: add search to dropdown and multi-select fields

gosiexon-zen:
## Description

Added possibility to search in drop-down fields: Tagger and MultiSelect component. DropDown component is used only for system fields.

PR with translations: https://github.com/zendesk/copenhagen_theme/pull/758

Jira: [GG-3415](https://zendesk.atlassian.net/browse/GG-3415) & [GG-3416](https://zendesk.atlassian.net/browse/GG-3416)
## Screenshots

https://github.com/user-attachments/assets/72d5595a-c6af-4283-93fd-b68f4a39dc99

## Checklist

[GG-3415]: https://zendesk.atlassian.net/browse/GG-3415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GG-3416]: https://zendesk.atlassian.net/browse/GG-3416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,8 @@
   },
   "resolutions": {
     "@types/react": "^17.x",
-    "nwsapi": "^2.2.20"
+    "nwsapi": "^2.2.20",
+    "@zendeskgarden/container-utilities": "^1.0.0"
   },
   "commitlint": {
     "extends": [
--- a/src/modules/ticket-fields/fields/MultiSelect.test.tsx
+++ b/src/modules/ticket-fields/fields/MultiSelect.test.tsx
@@ -1,0 +1,255 @@
+import { render } from "../../test/render";
+import { screen, waitFor } from "@testing-library/react";
+import { userEvent } from "@testing-library/user-event";
+import { MultiSelect } from "./MultiSelect";
+import type { TicketFieldObject } from "../data-types/TicketFieldObject";
+
+describe("MultiSelect", () => {
+  const mockOnChange = jest.fn();
+
+  const baseField: TicketFieldObject = {
+    id: 1,
+    name: "test_multiselect",
+    label: "Test MultiSelect",
+    type: "multiselect",
+    value: [],
+    error: null,
+    required: false,
+    description: "",
+    options: [
+      { name: "Option 1", value: "option_1" },
+      { name: "Option 2", value: "option_2" },
+      { name: "Option 3", value: "option_3" },
+      { name: "Option 4", value: "option_4" },
+    ],
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.runOnlyPendingTimers();
+    jest.useRealTimers();
+  });
+
+  describe("Rendering", () => {
+    it("renders combobox with accessible label", () => {
+      render(<MultiSelect field={baseField} onChange={mockOnChange} />);
+
+      expect(
+        screen.getByRole("combobox", { name: /test multiselect/i })
+      ).toBeInTheDocument();
+    });
+
+    it("marks field as required when required prop is true", () => {
+      const requiredField = { ...baseField, required: true };
+      render(<MultiSelect field={requiredField} onChange={mockOnChange} />);
+
+      const combobox = screen.getByRole("combobox");
+      expect(combobox).toHaveAttribute("aria-required", "true");
+    });
+
+    it("displays description hint", () => {
+      const fieldWithDescription = {
+        ...baseField,
+        description: "<p>Select multiple options</p>",
+      };
+      render(
+        <MultiSelect field={fieldWithDescription} onChange={mockOnChange} />
+      );
+
+      expect(screen.getByText("Select multiple options")).toBeInTheDocument();
+    });
+
+    it("displays error message with error validation", () => {
+      const fieldWithError = {
+        ...baseField,
+        error: "At least one option is required",
+      };
+      render(<MultiSelect field={fieldWithError} onChange={mockOnChange} />);
+
+      expect(
+        screen.getByText("At least one option is required")
+      ).toBeInTheDocument();
+    });
+
+    it("renders hidden inputs for selected values for form submission", () => {
+      const fieldWithValues = {
+        ...baseField,
+        value: ["option_1", "option_2"],
+      };
+      const { container } = render(
+        <MultiSelect field={fieldWithValues} onChange={mockOnChange} />
+      );
+
+      const hiddenInputs = container.querySelectorAll(
+        'input[type="hidden"][name="test_multiselect[]"]'
+      );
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -114,7 +114,8 @@
   },
   "resolutions": {
     "@types/react": "^17.x",
-    "nwsapi": "^2.2.20"
+    "nwsapi": "^2.2.20",
+    "@zendeskgarden/container-utilities": "^1.0.0"
```
gosiexon-zen, line 118: It was to fix a Rollup build error. The `@zendeskgarden/react-dropdowns.legacy` package imports KEY_CODES from `@zendeskgarden/container-utilities`, but that export only exists in v1.x. Our package.json has `@zendeskgarden/container-utilities` at ^2.0.2 in dependencies, and v2.x removed the KEY_CODES export — so the legacy dropdowns package couldn't resolve it and the build failed.
- But this is already fixed in Melissa's PR: https://github.com/zendesk/copenhagen_theme/pull/751

SirishaNeduri (dismissed):

Fredx87 (commented):
Some general comments:

### Difference with Support
I noticed a couple of differences between this component and the one in support. It doesn't need to be exactly the same, but it would be nice to have some consistency. 

In our case, the search results are presented "flat" (both for tagger and multiselect):

Image

while in Support they are presented with the full path:

Image

I think it is nicer to see the full path, since we also show it when an item is selected. It also nice that it highlights the search string

Second difference, when an item is selected from search in the `Tagger` component, we don't show it as selected when the user opens the combobox, in some cases we show a completely different set of options where the selected one is not visible (in this case "precision" is selected):

Image

While support always show the right option as selected:

Image

I think this is a big problem for a11y. for a screen reader user there is no way to know which option is currently selected.

There are also some differences in how we present things (search placeholder, no results, ...). This may be ok since are coherent with the lookup fields

### Duplicated logic
There is a lot of duplicated logic in the two components around the search logic, like search state, filtered options, .... It would be better to consolidate it in a single hook, maybe even reusing the `useNestedOptions` hooks

### No tests
There are no tests for the search logic and the components. It is true that we don't have tests for these two components right now, but we can take the opportunity to add them

### Translation strings
Probably it is better to create a separate PR with the translations string only, so we avoid releasing this with untranslated strings

SirishaNeduri (dismissed):

SirishaNeduri (dismissed):

src/modules/ticket-fields/fields/MultiSelect.test.tsx
```diff
@@ -0,0 +1,188 @@
+import { render } from "../../test/render";
+import { screen } from "@testing-library/react";
+import { MultiSelect } from "./MultiSelect";
+import type { TicketFieldObject } from "../data-types/TicketFieldObject";
+
+describe("MultiSelect", () => {
```
Fredx87, line 7: There are some issues with these tests:

- they query elements with `getByText`. It would be better to query them by their accessible role as suggested in the Testing Library [doc](https://testing-library.com/docs/queries/about/#priority).
- there are no interaction tests. Just passing some props and check if there is some text in the rendered component. This doesn't offer any guarantee that the component works as intended, what happens when the user selects an item in the dropdown? does search work? We don't need to test any possible scenario but at least the main ones.

src/modules/ticket-fields/fields/searchOptions.ts
```diff
@@ -0,0 +1,63 @@
...
+  value: string;
+  label: string;
+  menuLabel?: string;
+}
+
+/**
+ * Builds a map of option values to their display labels.
+ * Nested options (with "::" in the name) are formatted as "Part1 > Part2 > Part3".
+ */
+export function buildOptionsMap(
```
Fredx87: These 2 functions are used only in the `useFieldSearch` hook, we can move them in same file, I don't think there is a use case to use them outside that hook

src/modules/ticket-fields/fields/Tagger.test.tsx
```diff
@@ -0,0 +1,152 @@
+import { render } from "../../test/render";
+import { screen } from "@testing-library/react";
+import { Tagger } from "./Tagger";
+import type { TicketFieldObject } from "../data-types/TicketFieldObject";
+
+describe("Tagger", () => {
```
Fredx87, line 7: Same issue as the `Multiselect` component tests

src/modules/ticket-fields/fields/Tagger.tsx
```diff
@@ -39,23 +85,66 @@ export function Tagger({ field, onChange }: TaggerProps): JSX.Element {
...
+    setIsSearching(false);
+    setFilteredOptions(currentGroup.options);
+  }, [
+    currentGroup.options,
+    debouncedFilterOptions,
+    setFilteredOptions,
+    setInputValue,
+    setIsSearching,
+  ]);
+
```
Fredx87: We shouldn't clear the input on focus, in my previous review I commented about keeping the group with the selected option selected, but also with that there is a problem with accessibility. 

Testing with VoiceOver, with this behavior of clearing the input, there is no way for a user to know if an option is selected or not. I think we should follow the [Garden examples](https://garden.zendesk.com/components/combobox#autocomplete) more closely, and don't delete the text automatically. If the user wants to search again they need to clear the input manually.

The current behavior is coherent with the Lookup Field, but I am not sure if that one was audited for accessibility.

src/modules/ticket-fields/fields/useFieldSearch.tsx
```diff
@@ -0,0 +1,103 @@
...
+  label: string;
+  menuLabel?: string;
+}
+
+interface UseFieldSearchProps {
+  fieldOptions: TicketFieldOptionObject[];
+  currentGroupOptions: OptionProps[];
+}
+
+interface UseFieldSearchResult {
```
Fredx87, line 72: This hook is exposing a lot of internal states to the consumer. it would be better to provide a cleaner interface

src/modules/ticket-fields/fields/Tagger.tsx
```diff
@@ -39,23 +85,66 @@ export function Tagger({ field, onChange }: TaggerProps): JSX.Element {
...
+      }
+
+      const newValue = changes.selectionValue;
+      const newLabel =
+        newValue === ""
+          ? EMPTY_OPTION_LABEL
+          : optionsMap.get(newValue) || newValue;
+      setSelectionValue(newValue);
+      setSelectedLabel(newValue === "" ? null : newLabel);
+      setInputValue(newLabel);
```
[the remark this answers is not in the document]
gosiexon-zen: Changed

PR #771, opened 2026-03-06
feat: add setting to enable and disable suggested articles

melzreal:
## Description
1. Add setting to enable and disable suggested articles
2. Update translations to pull the needed strings for the setting to display, by running yarn i18n:update-translations   and yarn download-locales then committing the results     

Jira: https://zendesk.atlassian.net/jira/software/c/projects/GG/boards/8177?selectedIssue=GG-4392&visitedUserSeg=true

## Screenshots

setting ON, suggested articles show

image

setting OFF, no suggested articles show

image

## Checklist


dmuneras (dismissed):

SirishaNeduri (dismissed):

Fredx87 (changes requested):
What is implemented in this PR is not what the Jira story says. The story is about
> hiding Suggested Articles component in New request form

This PR and the added setting are instead referring to the suggested articles in the community post page. I guess we will need to submit a new string for the description and then wait again for the translations

melzreal:
> I guess we will need to submit a new string for the description

yes, I'm so sorry. I somehow misread the acceptance criteria and failed to notice the new request form part. Have opened https://github.com/zendesk/copenhagen_theme/pull/783 with the modified string

src/modules/new-request-form/NewRequestForm.tsx
```diff
@@ -35,6 +35,7 @@ export interface NewRequestFormProps {
   userRole: string;
   userId: number;
   brandId: number;
+  showSuggestedArticles: boolean;
```
Fredx87: The code of the new-request-form module can be used by customers in their custom theme,  where the new added setting may not exist. They can do that by copying and pasting the bundled js assets, or copy and pasting the source code and recompile it. Did we check what happens with this code when the setting doesn't exist?

Hopefully it is not going to break anything, but I guess it will disable the feature by default. I am not sure it is the best way to handle it, since it was turned on by default before, so maybe we should add a default value of `true` so keep the existing behavior

Fredx87 (commented):
The approach is now correct, I left a comment about custom themes. I just realized that customers can use this code outside of the Copenhagen Theme

src/modules/new-request-form/NewRequestForm.tsx
melzreal: good point - I've made the chang in the new_request_page hbs template, since without the setting in the manifest it would break the upload trying to find {{ settings.show_suggested_articles }} . Now it defaults to true if the setting cannot be found - have tried uploading a theme containing the remaining code and without the new setting and that works with no issues

setting the prop to default to true made the setting be ignored and always default to true, hence making the change where its initialized instead of here

Fredx87 (approved):

PR #796, opened 2026-04-23
Add ask_ai button 

konradhopek-zen:
## Description

https://zendesk.atlassian.net/browse/PDSC-650

## Screenshots: 

Desktop: 

Screenshot 2026-05-06 at 12 39 26

Mobile: 

Screenshot 2026-05-06 at 12 39 40

### DIFF
--- a/styles/_header.scss
+++ b/styles/_header.scss
@@ -259,3 +259,69 @@ $header-height: 71px;
 #zd-modal-container ~ .skip-navigation {
   display: none;
 }
+
+// Garden-style primary "Ask AI" button used in the header navigation.
+// Mirrors `<Button isPrimary><Button.StartIcon><SparkleIcon /></Button.StartIcon> Ask AI</Button>`.
+// Uses the theme's `link_color` / `hover_link_color` settings so the button
+// stays coherent with link styling even when admins customize the theme.
+.ask-ai {
+  align-items: center;
+  background-color: $link_color;
+  border: 1px solid $link_color;
+  border-radius: 4px;
+  color: $brand_text_color;
+  display: inline-flex;
+  font-size: 14px;
+  font-weight: $font-weight-semibold;
+  gap: 6px;
+  line-height: 1.4286;
+  padding: 6px 14px;
+
+  &:visited {
+    color: $brand_text_color;
+  }
+
+  .ask-ai-icon {
+    flex-shrink: 0;
+  }
+}
+
+.nav-wrapper-desktop .ask-ai {
+  margin-inline-end: 16px;
+  padding: 6px 14px;
+
+  &,
+  &:visited {
+    color: $brand_text_color;
+  }
+
+  &:hover,
+  &:focus,
+  &:active {
+    background-color: $hover_link_color;
+    border-color: $hover_link_color;
+    color: $brand_text_color;
+    text-decoration: none;
+  }
+}
+
+.nav-wrapper-mobile .ask-ai {
+  margin: 4px 24px;
+  padding: 6px 14px;
+  width: auto;
+  height: auto;
+
+  &,
+  &:visited {
+    color: $brand_text_color;
+  }
+
+  &:hover,
+  &:focus,
+  &:active {
+    background-color: $hover_link_color;
+    border-color: $hover_link_color;
+    color: $brand_text_color;
+    text-decoration: none;
+  }
+}
--- a/templates/header.hbs
+++ b/templates/header.hbs
@@ -25,6 +25,12 @@
         {{/unless}}
       </ul>
     </nav>
+          {{#link 'chat' class='ask-ai'}}
+            <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16" aria-hidden="true" class="ask-ai-icon">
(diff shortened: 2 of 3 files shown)


translations.yml
```diff
@@ -423,3 +423,8 @@ parts:
       screenshot: "https://drive.google.com/file/d/18d736vhJIuaYHO_uPeg9E4IBaVnJCoVM/view?usp=sharing"
       value: "Requests list"
       obsolete: "2026-04-01"
+  - translation:
+      key: "txt.help_center_copenhagen_theme.ask_ai"
+      title: "Label for the Ask AI button"
+      screenshot: "https://drive.google.com/file/d/1ZimRsBMDU67fo-8CVzwDSd_M6coXnWtW/view?usp=drive_link"
```
zendesk-mariannabuchicchio: @konradhopek-zen could you please update the sharing settings to "anyone with the link can view"? Thanks!

konradhopek-zen: done, sorry

styles/_header.scss
```diff
@@ -259,3 +259,77 @@ $header-height: 71px;
 #zd-modal-container ~ .skip-navigation {
   display: none;
 }
+
+// Garden-style primary "Ask AI" button used in the header navigation.
+// Mirrors `<Button isPrimary><Button.StartIcon><SparkleIcon /></Button.StartIcon> Ask AI</Button>`.
+// Uses Garden's primary blue (blue-600/700) so the look matches the design system
+// regardless of the per-brand `brand_color` theme setting.
+$ask-ai-bg: #1F73B7;
+$ask-ai-bg-hover: #144A75;
```
Fredx87: I don't think we should define a static color here. The theme is customizable with the settings, ideally this color should respect a theme setting, otherwise it will look bad.

The buttons usually follows the [brand_color](https://github.com/zendesk/copenhagen_theme/blob/7a5d6491a57065fb60dc2e2b78aa9592c85ad29e/manifest.json#L12), but if instead we want this to be coherent with the links we could use the [link_color](https://github.com/zendesk/copenhagen_theme/blob/7a5d6491a57065fb60dc2e2b78aa9592c85ad29e/manifest.json#L33) instead.

Other comments:
- We want more or less the same feature for what we call "Milestone 2" for CX. For now the work is paused, but [our design](https://www.figma.com/design/NMYfnfN0XXFbb9C6jz6xC9/Conversational-Help-Center?node-id=1312-12958&t=IPKcD4hKRdHL35kj-0) was a bit different for this link. Can we check if we want to go to this "button" approach with both design teams?
- Can you add screenshots to the PR description showing how this looks like both on desktop and mobile?

translations.yml
```diff
@@ -423,3 +423,8 @@ parts:
...
+      value: "Ask AI"
```
Fredx87: This translation string doesn't belong here. This file is used only for the translations related to the theme settings in the manifest

templates/header.hbs
```diff
@@ -13,6 +13,14 @@
   <div class="nav-wrapper-desktop">
     <nav class="user-nav" id="user-nav" aria-label="{{t 'user_navigation'}}">
       <ul class="user-nav-list">
+        <li>
+          {{#link 'chat' class='ask-ai'}}
+            <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" focusable="false" viewBox="0 0 16 16" aria-hidden="true" class="ask-ai-icon">
+              <path fill="currentColor" d="M2.499 11a.5.5 0 0 1 .477.348l.256.797a1.01 1.01 0 0 0 .63.633l.789.248a.5.5 0 0 1 .001.954l-.79.252a1.007 1.007 0 0 0-.63.633l-.252.787a.5.5 0 0 1-.95.008l-.266-.79a1.034 1.034 0 0 0-.636-.639l-.781-.252a.5.5 0 0 1-.002-.95l.794-.26a1.023 1.023 0 0 0 .636-.634l.248-.786A.5.5 0 0 1 2.5 11ZM1 7.513a1 1 0 0 1 .69-.953l2.583-.844a3.95 3.95 0 0 0 2.465-2.457l.808-2.56A1 1 0 0 1 9.452.695l.832 2.598a3.906 3.906 0 0 0 2.448 2.453l2.569.811a1 1 0 0 1 .004 1.906l-2.572.823a3.896 3.896 0 0 0-2.449 2.454l-.82 2.565a1 1 0 0 1-1.9.014l-.866-2.567v-.002A3.971 3.971 0 0 0 4.24 9.284l-2.547-.821A1 1 0 0 1 1 7.513Z"/>
+            </svg>
+            <span>{{t 'ask_ai'}}</span>
```
Fredx87: To be able to use this translation in this way, you need first to expose an `ask_ai` translation in the `t` helper from Help Center: https://developer.zendesk.com/api-reference/help_center/help-center-templates/helpers/#t-translation-helper

konradhopek-zen: done: put here: https://github.com/zendesk/help_center/pull/33764

styles/_header.scss
```diff
@@ -259,3 +259,74 @@ $header-height: 71px;
...
+
+.nav-wrapper-desktop .ask-ai {
+  margin-right: 16px;
+  padding: 6px 14px;
+
+  [dir="rtl"] & {
+    margin-right: 0;
+    margin-left: 16px;
+  }
+
```
Fredx87: Nit: instead of redefining the margins with `[dir="rtl"]` we could use [CSS Logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Logical_properties_and_values/Margins_borders_padding) - so `margin-inset-end` instead of `left` and `right`.

It would be more readable, but feel free to ignore if you want, since this pattern is already used in other pages

Fredx87 (commented):
I added a minor comment that can be ignored.

But before merging we should at least squash all the commits into one with a message similar to `feat: add Ask AI button`

zendesk-mariannabuchicchio:
@konradhopek-zen it looks like the yml file for the translations was removed. If you don't need that, could please remove localization from the reviewers? Thanks!

Fredx87 (approved):

PR #801, opened 2026-05-04
feat: preview mode in sc item

z-kzachradnik:
## Description

Render a Service Catalog item in preview mode on the Help Center: pinned banner at the top, request submission disabled, drafts rendered without errors.

## Screenshots

image

## Checklist

### DIFF
--- a/src/modules/service-catalog/components/service-catalog-item/PreviewModeBanner.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/PreviewModeBanner.tsx
@@ -1,0 +1,116 @@
+import { useEffect, useRef, useState } from "react";
+import { createPortal } from "react-dom";
+import styled from "styled-components";
+import { GlobalAlert } from "@zendeskgarden/react-notifications";
+import { getColor } from "@zendeskgarden/react-theming";
+import { useTranslation } from "react-i18next";
+import { Z_INDEX_ABOVE_NAVBAR } from "../../../shared/notifications/constants";
+
+const StyledGlobalAlert = styled(GlobalAlert)`
+  background: ${({ theme }) => getColor({ theme, hue: "blue", shade: 200 })};
+  border-bottom: 1px solid
+    ${({ theme }) => getColor({ theme, hue: "blue", shade: 300 })};
+  color: ${({ theme }) => getColor({ theme, hue: "blue", shade: 700 })};
+  box-shadow: none;
+
+  svg {
+    color: ${({ theme }) => getColor({ theme, hue: "blue", shade: 600 })};
+  }
+`;
+
+const StyledGlobalAlertTitle = styled(GlobalAlert.Title)`
+  color: ${({ theme }) => getColor({ theme, hue: "blue", shade: 800 })};
+`;
+
+export function PreviewModeBanner() {
+  const { t } = useTranslation();
+  const [portalHost, setPortalHost] = useState<HTMLDivElement | null>(null);
+  const initialBodyPaddingTopRef = useRef<string>("");
+
+  useEffect(() => {
+    const host = document.createElement("div");
+    host.setAttribute("data-preview-mode-banner-host", "true");
+    host.style.position = "fixed";
+    host.style.top = "0";
+    host.style.left = "0";
+    host.style.right = "0";
+    host.style.zIndex = String(Z_INDEX_ABOVE_NAVBAR);
+
+    initialBodyPaddingTopRef.current = document.body.style.paddingTop;
+
+    document.body.prepend(host);
+    setPortalHost(host);
+
+    return () => {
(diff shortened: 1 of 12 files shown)


src/modules/service-catalog/components/service-catalog-item/PreviewModeBanner.tsx
```diff
@@ -0,0 +1,116 @@
...
+        lastPadding = nextPadding;
+      }
+    };
+
+    const scheduleApply = () => {
+      if (scheduled) {
+        return;
+      }
+      scheduled = true;
+      window.requestAnimationFrame(applyLayout);
```
kristkvch-zen: Here we have a callback, but don't keep its id and we never cancel it in a cleanup. There might be a risk of a stale callback after unmount

src/modules/service-catalog/components/service-catalog-item/PreviewModeBanner.tsx
```diff
@@ -0,0 +1,116 @@
...
+    if (typeof ResizeObserver !== "undefined") {
+      resizeObserver = new ResizeObserver(scheduleApply);
+      resizeObserver.observe(portalHost);
+    }
+
+    window.addEventListener("resize", scheduleApply);
+
+    return () => {
+      resizeObserver?.disconnect();
+      window.removeEventListener("resize", scheduleApply);
```
kristkvch-zen, line 91: Also, probably it's better to reset the padding here in cleanup, so that after deleting a banner we guarantee there's no paddingTop

src/modules/service-catalog/components/service-catalog-item/PreviewModeBanner.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const initialBodyPaddingTopRef = useRef<string>("");
+
+  useEffect(() => {
+    const host = document.createElement("div");
+    host.setAttribute("data-preview-mode-banner-host", "true");
+    host.style.position = "fixed";
+    host.style.top = "0";
+    host.style.left = "0";
+    host.style.right = "0";
+    host.style.zIndex = "2147483647";
```
kristkvch-zen: We already have this "magic number" as Z_INDEX_ABOVE_NAVBAR constant

kristkvch-zen (dismissed):

src/modules/service-catalog/components/service-catalog-item/PreviewModeBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+import styled from "styled-components";
+import { GlobalAlert } from "@zendeskgarden/react-notifications";
+import { useTranslation } from "react-i18next";
+import { Z_INDEX_ABOVE_NAVBAR } from "../../../shared/notifications/constants";
+
+const PREVIEW_BLUE_200 = "#CEE2F2";
+const PREVIEW_BLUE_300 = "#ADCCE4";
+const PREVIEW_BLUE_600 = "#2F80B5";
+const PREVIEW_BLUE_700 = "#144A75";
+const PREVIEW_BLUE_800 = "#0F3554";
```
gosiexon-zen: Why we are not using Garden theme colors?

src/modules/service-catalog/components/service-catalog-item/PreviewModeBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const StyledGlobalAlert = styled(GlobalAlert)`
+  background: ${PREVIEW_BLUE_200};
```
gosiexon-zen: In Garden GlobarAlert of type info is already blue so why we need to override it?

z-kzachradnik: Garden's info defaults to primaryHue.* shades, but in Copenhagen primaryHue is overridden with the merchant's brand_color in createTheme.ts. Without our override the preview banner would render in whatever brand color the merchant set (orange, green, etc.), which defeats the purpose of preview mode being a stable system-level signal. The banner is also admin-only (only shown during draft preview, not to end users), so merchant brand consistency doesn't really apply here. The Figma design intentionally uses fixed blue for that reason. Merchants can still override via custom CSS in style.css if they really want to — we're just providing a sensible default.

gosiexon-zen: Ok thanks, at first I was also thinking why not using our preview mode that we have for Theming Center editor but probably you want different design and not connect it with TC. Make sense :)

gosiexon-zen (dismissed):

gosiexon-zen (approved):

kristkvch-zen (approved):

PR #806, opened 2026-05-04
feat: filter ticket fields by current brand form

melzreal:
## Description
- Customers have requested ticket fields to only display if associated with the brand that is currently being viewed.
 To achieve this, the api for ticket fields does not provide an option but the api for ticket forms does - so we fetch ticket forms for the current brand, filter for active forms, then filter the ticket fields to only display if included in the active forms for the brand.

- Also added "**finally**" clause to make sure isLoading will always be set to false

https://zendesk.atlassian.net/jira/software/c/projects/GG/boards/8177?selectedIssue=GG-4963&visitedUserSeg=true

## Screenshots

before, fields for a form belonging to another domain displayed:
image

after, fields for a form belonging to another domain (BoopBrandCheckbox) is filtered out, as well as date field that is not currently used by any form:

image

## Checklist

### DIFF
--- a/src/modules/request-list/hooks/useTicketFields.ts
+++ b/src/modules/request-list/hooks/useTicketFields.ts
@@ -1,6 +1,6 @@
 import { useEffect, useState } from "react";
 import type { CursorPaginatedResponse } from "../utils/pagination/CursorPaginatedResponse";
-import type { TicketField } from "../data-types";
+import type { TicketField, TicketForm } from "../data-types";
 import { fetchAllCursorPages } from "../utils/pagination/fetchAllCursorPages";
 
 interface UseTicketFields {
@@ -16,6 +16,20 @@ async function listTicketFields(
   const response = await fetch(
     `/api/v2/ticket_fields.json?locale=${locale}&page[size]=${pageSize}`
   );
+
+  if (!response.ok) {
+    throw new Error(response.statusText);
+  }
+  return await response.json();
+}
+
+async function listTicketForms(
+  locale: string
+): Promise<{ ticket_forms: TicketForm[] }> {
+  const response = await fetch(
+    `/api/v2/ticket_forms?locale=${locale}&associated_to_brand=true`
+  );
+
   if (!response.ok) {
     throw new Error(response.statusText);
   }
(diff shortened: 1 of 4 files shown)


src/modules/request-list/hooks/useTicketFields.ts
```diff
@@ -16,6 +16,20 @@ async function listTicketFields(
...
+    throw new Error(response.statusText);
+  }
+  return await response.json();
+}
+
+async function listTicketForms(
+  locale: string
+): Promise<{ ticket_forms: TicketForm[] }> {
+  const response = await fetch(
+    `/api/v2/ticket_forms?locale=${locale}&associated_to_brand=true`
```
melzreal: this endpoint also allows for pagination, but since in [help_center](https://github.com/zendesk/help_center/blob/melreal/end-user-conditions/app/models/ticket_form_backend.rb#L94) we opted out I went with the same approach, happy to change if deemed necessary

luis-almeida: Could we just query with `active=true` to get the list of `active` forms "filtered" already?

I'm also not sure if we shouldn't pass `end_user_visible=true` as well. But maybe not if the goal is to filter out all fields associated to other brands even if those forms are not visible to end users. Maybe @Fredx87 has thought about this already.

Fredx87: I haven't thought about it, but this page is also accessible by agents/admins where they can see their requests, so I am not sure we should restrict for end_user_visible forms. The API already returns only the ones visible to end users if the logged in user is an end user.

But for sure we need to fetch all forms, including service catalog forms. This was not specified in the Jira story but I just remembered that by default the ticket forms api doesn't return service catalog forms, unless we pass `form_type=all`. And I think in that case the API is paginated so we need to perform multiple calls to fetch all pages

melzreal: thanks - have updated to use `form_type=all` and `ticket_forms=active` instead of filtering after, and finally to use cursor pagination like the ticket fields already did

src/modules/request-list/hooks/useTicketFields.ts
```diff
@@ -29,21 +44,33 @@ export function useTicketFields(locale: string): UseTicketFields {
 
   async function fetchTicketFields() {
     try {
-      const response = await fetchAllCursorPages(
-        () => listTicketFields(locale),
-        "ticket_fields"
+      const [response, tfResponse] = await Promise.all([
```
Fredx87: Nit: `response` and `tfResponse` are not quite clear. We could just call them `ticketFields` and `ticketForms`

src/modules/request-list/hooks/useTicketFields.ts
```diff
@@ -29,21 +44,33 @@ export function useTicketFields(locale: string): UseTicketFields {
...
       setError(error as Error);
+    } finally {
+      setIsLoading(false);
     }
   }
 
   useEffect(() => {
     fetchTicketFields();
-  }, []);
+  }, [locale]);
```
Fredx87: This is logically correct but triggers this warning from ESLint:
```
React Hook useEffect has a missing dependency: 'fetchTicketFields'. Either include it or remove the dependency array
```

Because the hook itself depends only on the function, and not on the locale. We can just bring the function inside the `useEffect` and in this case the hook will depends only on `locale`.

Moreover, this is not strictly needed right now, since there is no way to change the locale at runtime. When the Help Center locale changes, the whole page reloads. But I think it still could make sense to have it as a dependency if in the future this thing will change

Fredx87 (commented):
Added a couple of small comments.

With this approach (fetch all ticket form pages) there is a risk that we need perform a lot of calls to get the data. For example, if a customer has 1000 forms for a brand, we will need to make 10 calls to the ticket forms API.

Since we don't have any other alternatives right now, I think it is fine. I also tried to check some data and apparently the maximum number of forms for a brand is ~300. If in the future we will see that this approach can cause problems, we need to ask for a new API that returns directly the data we need

src/modules/request-list/hooks/useTicketFields.ts
melzreal: I wasnt sure if a full reload happened on locale change, in that case its indeed not needed will remove for simplicity

Fredx87 (approved):

PR #811, opened 2026-06-01
Resolve NPM vulnerabilities 

gosiexon-zen:
## Description

Resolve 13 open Dependabot security alerts by updating vulnerable dependencies using targeted resolutions.
**Used targeted `yarn set resolution` commands** to patch only specific vulnerable version ranges in `yarn.lock`, rather than forcing versions globally in `package.json`.

After upgrading lodash to 4.18.0 to fix security vulnerabilities, commitlint's CI checks failed because `git-raw-commits@4.0.0` (a dependency of commitlint) was incompatible with the new lodash version. Resolved this by:
  1. Adding a resolution for `git-raw-commits@5.0.1` (compatible with lodash 4.18.0)
  2. Upgrading commitlint from v19 to v20 (which supports the newer git-raw-commits)
  
  #### Commitlint v17 → v20 Breaking Changes Impact

  | Version | Breaking Change | Impact on This Project |
  |---------|----------------|----------------------|
  | v18 | Node.js v18+ required | ✅ None (have v20.15.1) |
  | v19 | Pure ESM migration | ✅ None (config in package.json) |
  | v20 | URL line length exception | ✅ None (rule more lenient) |

Most updates use Yarn's `resolutions` field to force patched versions of transitive (indirect) dependencies—packages we don't directly depend on but are pulled in by other packages—ensuring all instances across the dependency tree use the secure version.

JIRA [https://zendesk.atlassian.net/browse/GG-5037](https://zendesk.atlassian.net/browse/GG-5037)
## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,12 +113,7 @@
   },
   "resolutions": {
     "@types/react": "^17.x",
-    "nwsapi": "^2.2.20",
-    "handlebars": "4.7.9",
-    "lodash": "4.18.0",
-    "picomatch": "2.3.2",
-    "tar": "7.5.11",
-    "git-raw-commits": "5.0.1"
+    "nwsapi": "^2.2.20"
   },
   "commitlint": {
     "extends": [
--- a/package.json
+++ b/package.json
@@ -114,7 +114,8 @@
   "resolutions": {
     "@types/react": "^17.x",
     "nwsapi": "^2.2.20",
-    "handlebars": "4.7.9"
+    "handlebars": "4.7.9",
+    "lodash": "4.18.0"
   },
   "commitlint": {
     "extends": [
--- a/package.json
+++ b/package.json
@@ -53,8 +53,8 @@
     "styled-components": "^5.3.11"
   },
   "devDependencies": {
-    "@commitlint/cli": "17.3.0",
-    "@commitlint/config-conventional": "17.3.0",
+    "@commitlint/cli": "^20.0.0",
+    "@commitlint/config-conventional": "^20.0.0",
     "@jest/globals": "^29.6.4",
     "@rollup/plugin-commonjs": "^25.0.2",
     "@rollup/plugin-dynamic-import-vars": "^2.1.2",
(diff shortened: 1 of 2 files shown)


BrunoBFerreira (dismissed):
Maybe we can add a list of the dependencies and the target versions to the description, just to make it easier to find what was updated in the PR. 

Otherwise, looks good.

package.json
```diff
@@ -53,8 +53,8 @@
     "styled-components": "^5.3.11"
   },
   "devDependencies": {
-    "@commitlint/cli": "17.3.0",
-    "@commitlint/config-conventional": "17.3.0",
+    "@commitlint/cli": "^20.0.0",
+    "@commitlint/config-conventional": "^20.0.0",
```
Fredx87, line 57: These are major bumps. Did we check that there are no breaking changes we need to address, and that everything still works as before?

package.json
```diff
@@ -113,7 +113,12 @@
   },
   "resolutions": {
     "@types/react": "^17.x",
-    "nwsapi": "^2.2.20"
+    "nwsapi": "^2.2.20",
+    "handlebars": "4.7.9",
+    "lodash": "4.18.0",
+    "picomatch": "2.3.2",
+    "tar": "7.5.11",
+    "git-raw-commits": "5.0.1"
```
Fredx87: We shouldn't put these fixed resolutions in the package.json. These are very hard to maintain and could even prevent vulnerabilities fixes in the future.

We should use `yarn set resolution` instead. See my comment [here](https://github.com/zendesk/guide-app-framework/pull/476#pullrequestreview-2868645490)

package.json
gosiexon-zen: I added comparison to the description, we should be good. Claude checked breaking changes and didn't find any issues.

gosiexon-zen: This is what it wrote also:
 Additional Confidence Points:

  1. ✅ All tests pass with v20 (494 tests, 63 suites)
  2. ✅ Simple config - you're using the standard @commitlint/config-conventional with no custom rules
  3. ✅ No custom plugins - reduces risk of incompatibility
  4. ✅ CI passed locally with yarn commitlint --help working correctly

package.json
gosiexon-zen: Thanks! I changed to using yarn set resolutions.

gosiexon-zen: Also added comment to the description why I couldn't update Tar 6.x Vulnerability, can you check if it is ok?

Fredx87 (approved):

PR #812, opened 2026-06-01
fix(ticket-fields): show selected label for nested dropdown values

Fredx87:
## Description

Nested dropdown fields (single-select `Tagger` and `MultiSelect`) stopped displaying the label of an already-selected value after the Garden v9 upgrade (Copenhagen theme v4.18). The value was correctly selected and submitted, but the closed combobox showed the empty placeholder instead of the selected option's label. This was most visible when values are set externally, e.g. via prefilled URL parameters.

The previous implementation relied on rendering every option on the first render and trusting the Combobox to cache their labels so it could resolve a selected value that lives in a non-visible group. Garden v9 rebuilds that label cache from the currently rendered options on every render, so once the options were removed the selected label was lost.

Instead of depending on that internal caching, resolve the label from the full option list directly: Tagger computes the displayed label via `renderValue`, and `MultiSelect` renders hidden options for selected values so their tags resolve, without affecting menu navigation. Tagger now also opens on the group that contains the selected value, which is both clearer for the user and no longer required for correctness.

Adds component-level tests covering selected-value labels (root, nested and empty) and the existing navigation/selection behavior to guard against regressions.

## Screenshots

Before:
- when a nested dropdown has a value selected (either by submitting a form with errors or by prefilling via URL), the value is not shown
- when a nested multiselect has a value selected (either by submitting a form with errors or by prefilling via URL), the selected tags show the raw value instead of the label
- when navigating around in a nested multiselect, the selected tags show the raw value instead of the label for all the selected items that are not in the current selected level

https://github.com/user-attachments/assets/a394bb48-cda1-4c9e-abdf-2e21d547da12

After:
- when a nested dropdown has a value selected (either by submitting a form with errors or by prefilling via URL), the value is shown correctly
- when a nested multiselect has a value selected (either by submitting a form with errors or by prefilling via URL), the selected tags show the label of the selected items
- when navigating around in a nested multiselect, the selected tags show the label of the selected items

https://github.com/user-attachments/assets/73bcc8b6-b4a5-44b9-b5f4-88b7ef92580c

## Checklist

### DIFF
--- a/src/modules/ticket-fields/fields/useNestedOptions.tsx
+++ b/src/modules/ticket-fields/fields/useNestedOptions.tsx
@@ -190,39 +190,64 @@ function buildNestedOptions(
 }
 
 /**
- * When one or more options are selected, the Combobox component renders the label
- * for an option in the input, searching for an option passed as a child with the
- * same value as the selected option.
+ * Builds a map from each selectable option value to the label that should be
+ * displayed for it once selected (e.g. `"Bass > Fender > Precision"`).
  *
- * In the first render we are passing only the root group options as children,
- * and if we already have some selected values from a SubGroup, the component is not
- * able to find the label for the selected option.
- *
- * We therefore need to pass all the non-navigation options as children in the first render.
- * The passed options are cached by the Combobox component, so we can safely remove them
- * after the first render and pass only the root group options.
+ * The Combobox only renders the options for the currently displayed group, so a
+ * selected value that lives in a different group has no matching option child to
+ * derive its label from. Consumers use this map to resolve the label of a
+ * selected value regardless of which group is currently displayed.
  */
-function getInitialGroup(nestedOptions: NestedOptions): RootGroup {
-  const result: RootGroup = {
-    type: "RootGroup",
-    options: [],
-  };
+function buildOptionLabels(
+  nestedOptions: NestedOptions
+): Record<string, string> {
+  const result: Record<string, string> = {};
 
   Object.values(nestedOptions).forEach(({ options }) => {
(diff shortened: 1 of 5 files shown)


BrunoBFerreira (approved):

src/modules/ticket-fields/fields/useNestedOptions.tsx
```diff
@@ -235,13 +260,23 @@ export function useNestedOptions({
...
+    () => buildOptionLabels(nestedOptions),
+    [nestedOptions]
+  );
+
+  const initialGroup = useMemo(
+    () =>
+      findGroupForValue(nestedOptions, selectedValue) ??
+      nestedOptions[ROOT_GROUP_IDENTIFIER],
+    [nestedOptions, selectedValue]
   );
```
melzreal, line 273: is useMemo necessary? I guess it could be simply

const optionLabels = buildOptionLabels(nestedOptions);

const initialGroup =
  findGroupForValue(nestedOptions, selectedValue) ??
  nestedOptions[ROOT_GROUP_IDENTIFIER];

melzreal (approved):
LGTM left a small question on useMemo

src/modules/ticket-fields/fields/useNestedOptions.tsx
melzreal: asked this to chatgpt and the answer was that its reasonable because the values are derived from props instead of stored, and that for a dropdown UI this is "the right pattern" because of typing, parent component updates, hover focus changes and unrelated state changes 👍 so will resolve

PR #814, opened 2026-06-08
fix: update styling for wysiwyg resize

BrunoBFerreira:
## Description

When reducing the size of the window horizontally to a size used for mobile, the wysiwyg would not resize properly and present some overflow. This is a subpar experience to our users and some tickets have been raised around this.

The wysiwyg parameters seemed to be correct, allowing the toolbar collapse, but the behavior being shown was not. After some debugging, we found out the issue was not in the wysiwyg itself but in the elements that contain it, which were not resizing correctly.

This PR addresses this issue by updating the styling of the `main` container of the request page, allowing for the max width to be 100%. 

### Before

Screenshot 2026-06-08 at 14 45 05

### After

Screenshot 2026-06-08 at 14 42 48

## Checklist


Fredx87 (approved):

melzreal (approved):
:shipit:

PR #819, opened 2026-06-18
Update translations to pull latest

melzreal:
## Description

Update translations as requested by i18n team

## Checklist


BrunoBFerreira (dismissed):

bin/update-modules-translations.mjs
```diff
@@ -6,7 +6,7 @@
 import { writeFile, readFile, mkdir } from "node:fs/promises";
 import { resolve } from "node:path";
 import { glob } from "glob";
-import { load } from "js-yaml";
+import yaml from "js-yaml";
```
Fredx87: Why do we need this change? For me the script works without this change

melzreal: it does not work for me, when I checkout master it throws this error:
image
```
import { load } from "js-yaml";
         ^^^^
SyntaxError: Named export 'load' not found. The requested module 'js-yaml' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'js-yaml';
const { load } = pkg;

    at #asyncInstantiate (node:internal/modules/esm/module_job:302:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:405:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:654:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v25.3.0
```

melzreal: update: was a stale node modules cache, removed

Fredx87 (approved):

PR #823, opened 2026-06-25
feat: add MultiLookupField for multi-select lookup ticket fields (CD-4026)

byronhabana:
## Description

Adds `MultiLookupField` component for CD-4026, enabling end users to select up to 20 custom object records via a multi-select autocomplete dropdown.

**Changes:**
- New `MultiLookupField.tsx` using Garden's `Combobox` with `isMultiselectable` + `isAutocomplete`
- Routes `multi_lookup` field type in `RequestFormField`
- Hides `multi_lookup` from request list table columns
- Submits via hidden `name[]` inputs for Rails array params (same pattern as `MultiSelect`)
- Gated passively by backend — component only renders if BE returns a field with `type: "multi_lookup"`

## Screenshots

Mocked only — requires backend (CD-4025) to serve `multi_lookup` fields. UI uses Garden's native multiselect Combobox with tags.

Screenshot 2026-06-26 at 11 41 33 AM

https://github.com/user-attachments/assets/2605b1dc-971f-4662-974a-25e2fd357b41

## Checklist


SirishaNeduri (dismissed):

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,336 @@
...
+import { useTranslation } from "react-i18next";
+import {
+  buildAdvancedDynamicFilterParams,
+  getCustomObjectKey,
+} from "./LookupField";
+import type { CustomObjectRecord } from "../data-types/CustomObjectRecord";
+
+const MAX_SELECTIONS = 20;
+
+export interface MultiLookupFieldProps {
```
luis-almeida: Nit: seems to only be used here so no need to export.

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,336 @@
...
+                "cph-theme-ticket-fields.multi-lookup-field.max-reached",
+                "Maximum selections reached"
+              )
+            : t(
+                "cph-theme-ticket-fields.lookup-field.placeholder",
+                "Search {{label}}",
+                { label }
+              )
+        }
+        isDisabled={isAtLimit}
```
luis-almeida: Doesn't this mean that the Combox is disabled also for removing the fields?
And if that's the case, how will the user be able to modify the selection after reaching the limit?

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,336 @@
...
+              {renderOption ? renderOption(option) : option.name}
+            </Option>
+          ))}
+      </Combobox>
+      {error && <Field.Message validation="error">{error}</Field.Message>}
+      {selectedValues.map((val) => (
+        <input key={val} type="hidden" name={`${name}[]`} value={val} />
+      ))}
+      {selectedValues.length === 0 && (
+        <input type="hidden" name={`${name}[]`} value="" />
```
luis-almeida: Does the backend handle this situation correctly? I believe this will submit an array with empty string.

src/modules/request-list/components/requests-table/RequestsTable.tsx
```diff
@@ -31,6 +31,7 @@ const HIDDEN_FIELDS = [
   "assignee",
   "custom_status",
   "lookup",
+  "multi_lookup",
```
luis-almeida, line 34: I think we also want to add this field in [FilterPropertyDropdown](https://github.com/zendesk/copenhagen_theme/blob/master/src/modules/request-list/components/requests-toolbar/filter-modal/FilterPropertyDropdown.tsx#L11)

src/modules/ticket-fields/fields/MultiLookupField.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+  it("shows description when provided", () => {
+    render(
+      <MultiLookupField
+        {...defaultProps}
+        field={{ ...defaultField, description: "Pick multiple records" }}
+      />
+    );
+    expect(screen.getByText("Pick multiple records")).toBeInTheDocument();
+  });
+});
```
luis-almeida, line 234: Could be nice to have specs for adding and removing items but I see that most of the other test files also just cover the rendering part.

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,336 @@
...
+    id: "no-results",
+  };
+
+  const fetchSelectedOptions = useCallback(
+    async (recordIds: string[]) => {
+      const results: TicketFieldOptionObject[] = [];
+      for (const id of recordIds) {
+        try {
+          const res = await fetch(
+            `/api/v2/custom_objects/${customObjectKey}/records/${id}`
```
luis-almeida: From what I understand this can fire up to 20 sequential queries (?)
Would it be possible to hit `/api/v2/custom_objects/${customObjectKey}/records` and get them all at once using [`filter[ids]`](https://developer.zendesk.com/api-reference/custom-data/custom-objects/custom_object_records/#parameters)?

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,336 @@
...
+    <Field>
+      <Field.Label>
+        {label}
+        {required && <Span aria-hidden="true">*</Span>}
+      </Field.Label>
+      {description && (
+        <Field.Hint dangerouslySetInnerHTML={{ __html: description }} />
+      )}
+      <Combobox
+        inputProps={{ required: required && selectedOptions.length === 0 }}
```
luis-almeida: Have we made sure that this new component is accessible?
As an example, `MultiSelect` also sets an "aria-required" attribute

luis-almeida (commented):
Added a few comments but have one more question:

In the PR description it mentions support for Service Catalog but this seems to have been descoped (?) Will it be added later on? c.c. @zendesk/butlers

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: cleaned that up here https://github.com/zendesk/copenhagen_theme/pull/823/commits/d6aaeecc9316642c9100d99f0e1789cbc26ad190 thanks

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: good catch, fixed here https://github.com/zendesk/copenhagen_theme/pull/823/commits/d6aaeecc9316642c9100d99f0e1789cbc26ad190

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: ended up removing the fallback following the existing pattern for multiselect here https://github.com/zendesk/copenhagen_theme/pull/823/commits/d6aaeecc9316642c9100d99f0e1789cbc26ad190 thanks!

src/modules/request-list/components/requests-table/RequestsTable.tsx
byronhabana: added https://github.com/zendesk/copenhagen_theme/pull/823/changes/d6aaeecc9316642c9100d99f0e1789cbc26ad190

src/modules/ticket-fields/fields/MultiLookupField.test.tsx
byronhabana: added the tests here https://github.com/zendesk/copenhagen_theme/pull/823/changes/d6aaeecc9316642c9100d99f0e1789cbc26ad190

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: addressed this here https://github.com/zendesk/copenhagen_theme/pull/823/changes/d6aaeecc9316642c9100d99f0e1789cbc26ad190

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: yeah i missed one and addressed here https://github.com/zendesk/copenhagen_theme/pull/823/commits/d6aaeecc9316642c9100d99f0e1789cbc26ad190 thanks!

byronhabana:
Thanks for the feedback @luis-almeida , that's right service catalog is out of scope for this ticket hence I ended up removing it.

SirishaNeduri (dismissed):

konradhopek-zen (dismissed):

luis-almeida (dismissed):

hiroiguchi (dismissed):

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,354 @@
...
+            >
+              {renderOption ? renderOption(option) : option.name}
+            </Option>
+          ))}
+      </Combobox>
+      {error && <Field.Message validation="error">{error}</Field.Message>}
+      {isAtLimit && (
+        <Field.Message validation="error">
+          {t(
+            "cph-theme-ticket-fields.multi-lookup-field.limit-reached",
```
z-kzachradnik: You addded `cph-theme-ticket-fields.multi-lookup-field.max-reached` to en-us.yml file but you used `cph-theme-ticket-fields.multi-lookup-field.limit-reached`

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,354 @@
...
+            (record: CustomObjectRecord) => ({
+              name: record.name,
+              value: record.id,
+            })
+          );
+          setSelectedOptions(results);
+          onChange(results.map((o) => o.value));
+        }
+      } catch {
+        // skip failed fetch
```
z-kzachradnik: When the initial resolution request fails (network error or non-2xx), we currently leave selectedOptions empty. Since the hidden name[] inputs are derived from selectedOptions, a transient failure silently drops the field's previously-saved values, a subsequent submit would clear the field. Falling back to the raw record ids preserves the saved selection on submit (data-safe degradation); the ids are shown as labels until a later fetch resolves the real names. We intentionally don't call onChange in the fallback because the value is unchanged from what the parent passed in, so there's nothing to update and we avoid marking the form dirty

For example, before try/catch block you can define:
`const fallback = recordIds.map((id) => ({ value: id, name: id }));`

and then use it in catch

`setSelectedOptions(fallback);`

src/modules/ticket-fields/fields/MultiLookupField.tsx
```diff
@@ -0,0 +1,354 @@
...
+      <Field.Label>
+        {label}
+        {required && <Span aria-hidden="true">*</Span>}
+      </Field.Label>
+      {description && (
+        <Field.Hint dangerouslySetInnerHTML={{ __html: description }} />
+      )}
+      <Combobox
+        ref={wrapperRef}
+        inputProps={{ required }}
```
z-kzachradnik: Nit / worth double-checking: inputProps={{ required }} sits on the combobox text input, but we clear it with setInputValue("") after every selection, so the input is effectively always empty. On a form that doesn't set noValidate, native HTML validation might block submit with "please fill in this field" even when tags are selected (unlike the single LookupField, where inputValue holds the selected name so it passes). Since we already set aria-required manually on the combobox, we probably don't need required on the input at all.

Not blocking, just worth verifying the required-field submit flow before merge. 🙂

z-kzachradnik (changes requested):

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: good catch, addressed [here](https://github.com/zendesk/copenhagen_theme/pull/823/changes/530e4d370f695036832b3c4748b61210d01ec686)

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: addressed [here](https://github.com/zendesk/copenhagen_theme/pull/823/changes/530e4d370f695036832b3c4748b61210d01ec686)

src/modules/ticket-fields/fields/MultiLookupField.tsx
byronhabana: 😅 missed this one, addressed [here](https://github.com/zendesk/copenhagen_theme/pull/823/changes/530e4d370f695036832b3c4748b61210d01ec686)

SirishaNeduri (dismissed):

z-kzachradnik (dismissed):

z-kzachradnik (approved):

SirishaNeduri (approved):

luis-almeida (approved):
Approving the PR but can I please ask you to squash into just two or three well described commits to get a cleaner changelog entry? 🙏🏼  e.g.:

byronhabana:
Done — squashed the 13 commits down to 3:

1. `feat: add MultiLookupField for multi-select lookup ticket fields [CD-4026]` — the component, routing, and tests
2. `i18n: add MultiLookupField translation strings`
3. `chore: rebuild bundles for MultiLookupField`

Verified the resulting tree is byte-identical to the pre-squash HEAD (`git diff` came back empty), so no code changed — just history.

Heads up: force-pushing dismissed the existing approvals (yours, @z-kzachradnik's, @SirishaNeduri's), since this repo dismisses stale reviews on push. Since the diff is unchanged, hoping re-approval is quick — sorry for the churn.

[CD-4026]: https://zendesk.atlassian.net/browse/CD-4026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #830, opened 2026-07-08
Revert "feat: filter ticket fields by current brand form"

melzreal:
Reverts zendesk/copenhagen_theme#806

- There are different layers of permissions built on top of unfiltered objects, this change makes it so the default fields do not show unless the custom object have been specifically assigned to the brand. 

Other options:
- Make this a setting instead of reverting, disabled by default
- Disable the setting if view requests across brands is enabled


melzreal:
closing in favour of https://github.com/zendesk/copenhagen_theme/pull/833/commits

PR #833, opened 2026-07-10
feat: remove filter when requests across brands on

melzreal:
## Description
- ensure custom objects are not filtered per brand when viewing requests for all the brands

- also add default columns when the response for ticket forms is empty, to ensure we always show
the columns present in the default form

## Screenshots

with view requests across brands on:
image

with view requests across brands off:

image

## Checklist

### DIFF
--- a/src/modules/request-list/components/requests-list/RequestsList.test.tsx
+++ b/src/modules/request-list/components/requests-list/RequestsList.test.tsx
@@ -46,7 +46,13 @@ const renderComponent = async (params?: Partial<RequestListParams>) => {
     push,
   });
 
-  render(<RequestsList locale="en-us" customStatusesEnabled={false} />);
+  render(
+    <RequestsList
+      locale="en-us"
+      customStatusesEnabled={false}
+      viewRequestsAcrossBrandsEnabled={false}
+    />
+  );
 
   // This is needed because of a “bug” with the Jest fake timers
   // See: https://github.com/facebook/jest/issues/10221
@@ -94,6 +100,27 @@ describe("Rendering", () => {
     expect(screen.getAllByText("Mar 3, 2019")[0]).toBeInTheDocument();
     expect(screen.getAllByText("Jun 3, 2020")[0]).toBeInTheDocument();
     expect(screen.getAllByText("Open")[0]).toBeInTheDocument();
+  }, 10000);
+
+  test("renders default columns when ticketFields is empty", async () => {
+    (useTicketFields as jest.Mock).mockReturnValue({
+      ticketFields: [],
+    });
+
+    await renderComponent();
+
+    expect(screen.getByRole("button", { name: /id/i })).toBeInTheDocument();
+    expect(
+      screen.getByRole("button", { name: /created date/i })
+    ).toBeInTheDocument();
+    expect(
+      screen.getByRole("button", { name: /updated date/i })
+    ).toBeInTheDocument();
+    expect(screen.getByRole("button", { name: /status/i })).toBeInTheDocument();
+
+    expect(
+      screen.getByRole("link", { name: "My request" })
+    ).toBeInTheDocument();
   }, 10000);
 
   test("applies a default sort of updated at and descending", async () => {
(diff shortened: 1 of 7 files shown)


src/modules/request-list/components/requests-list/RequestsList.test.tsx
```diff
@@ -46,7 +46,13 @@ const renderComponent = async (params?: Partial<RequestListParams>) => {
...
   });
 
-  render(<RequestsList locale="en-us" customStatusesEnabled={false} />);
+  render(
+    <RequestsList
+      locale="en-us"
+      customStatusesEnabled={false}
+      viewRequestsAcrossBrandsEnabled={false}
+    />
+  );
```
BrunoBFerreira, line 55: Due to this change, should we add a test with the `viewRequestAcrossBrandsEnabled={true}`?

src/modules/request-list/hooks/useTicketFields.ts
```diff
@@ -37,28 +45,45 @@ async function listTicketForms(
...
-            ticketField.active && activeTicketFieldIds.has(ticketField.id)
-        )
-      );
+        setTicketFields(
+          ticketFields.filter(
+            (ticketField) =>
+              ticketField.active &&
+              (activeTicketFieldIds.size > 0
+                ? activeTicketFieldIds.has(ticketField.id)
+                : DEFAULT_FIELD_TYPES.includes(ticketField.type))
```
BrunoBFerreira: Could you explain this change a bit better? Feel I am missing some context.

melzreal: right now we call ticket fields and check for active fields, but this sometimes returns an empty array [] and results in a broken display for the requests page.

This happens when the customers have removed forms from the brand that is being viewed, so when we fetch the forms there are no active ticketfields

The workaround suggested for this was to make sure that we always display default columns [as done in this line 
](https://github.com/zendesk/copenhagen_theme/pull/833/changes#diff-2f205a36d078210ff96b3cb4aef19d99322f4e385e91b0060a9ef3dbb156c45eR92)
however since the default ticket form has these fields that you cannot remove, I thought I could also hardcode them to ensure they are present in the response and always available to the filters, even for a newly created brand:

Image

Image

I can also remove the hardcoded filters and only keep the logic that displays the columns, this was an "extra defensive" idea essentially

src/modules/request-list/components/requests-list/RequestsList.test.tsx
melzreal: I have added them, but to useTicketFields test - I can also add them here to showcase the functionality but all it would do is stub the call to ticket fields and the returned values - same as we are doing here:
https://github.com/zendesk/copenhagen_theme/pull/833/changes#diff-3ed1c95ab910dc68474478a76670d95821ecf567af2d1b3b0075b982e161d7cfR145

is it worth adding or just uneeded duplication? 🤔

BrunoBFerreira: Is there any code path (even as simple as it might be) that is not exercised? If the answer is no, then no duplication needed. I was reviewing again, and it seems we only added the parameter. If the only thing changed was the parameter pass, then maybe we can skip, assuming the point where it is passed and where it is used are tested.

src/modules/request-list/hooks/useTicketFields.ts
BrunoBFerreira: Although I think it is a good idea, having them hardcoded seems a bit of a risk since changes on mandatory fields can break this approach. If there was a field identifying mandatory fields or similar, it would be safer.

melzreal: not sure they will change, but we could argue its unecessary having filters for an empty list.. so I think I'll remove it, its overhead

src/modules/request-list/components/requests-list/RequestsList.test.tsx
melzreal: have added a test for the changes in request list that now display the default columns, indeed its not affected by the new prop but it is a separate behaviour that was not yet covered

BrunoBFerreira (approved):
Thank you for addressing all the comments 🙏

PR #842, opened 2026-07-21
fix: surface underlying service request submission errors

konradhopek-zen:
When a service catalog request fails validation, the backend 422 response no longer includes a field_key, which broke field-error mapping: every error was treated as a stale form ("Refresh the page") and no field lit up red. Handle the id-less format by surfacing the underlying error description(s) in the notification, and only show the refresh message when a field_key is present but unmatched.

Also add client-side required-field validation so blank required fields are flagged in red before submit, and clear a field's error when it is edited.

Z2 ticket: https://support.zendesk.com/agent/tickets/15014219 

backend changes: 
general PSA - Required fields are working now. I did have to change the response format in case of missing fields. Example response now:

```
{
    "error": "RecordInvalid",
    "description": "Record validation errors",
    "details": {
        "base": [
            {
                "description": "What are you looking for? cannot be blank",
                "error": "cannot be blank"
            }
        ]
    }
}
```
FE doesn't account for this yet and throws a generic error and the empty fields are not lightning up red in the form.

## Checklist


kristkvch-zen (dismissed):

ONBOARDING.md
```diff
@@ -0,0 +1,557 @@
+# Copenhagen Theme - Onboarding Guide
```
BrunoBFerreira: Why are we adding this file?

src/modules/service-catalog/components/service-catalog-item/ServiceCatalogItem.tsx
```diff
@@ -280,12 +287,20 @@ export function ServiceCatalogItem({
   async function handleValidationErrors(response: Response) {
     const errorData: ServiceRequestResponse = await response.json();
     const invalidFieldErrors = errorData?.details?.base ?? [];
-    const missingErrorFields = invalidFieldErrors.filter(
+
+    console.error("Service request validation failed:", errorData);
```
BrunoBFerreira: We should remove this.

src/modules/service-catalog/components/service-catalog-item/ServiceCatalogItem.spec.tsx
```diff
@@ -154,7 +154,7 @@ describe("ServiceCatalogItem", () => {
       label: "Field 1",
       required: true,
       options: [],
-      value: null,
+      value: "Field 1 value",
```
BrunoBFerreira, line 157: Shouldn't we keep some tests where this value is `null`?

kristkvch-zen: The actual null value validation is already tested in useValidateServiceItemForm.spec.ts. So adding a null here wouldn't test anything new.

BrunoBFerreira: If it does not exercise any new code path, then I think we can exclude it yes. I would still confirm if having the `null` here will not go through a code path that does something more than what we have in `useValidateServiceItemForm`.

src/modules/service-catalog/components/service-catalog-item/ServiceCatalogItem.tsx
```diff
@@ -280,12 +287,22 @@ export function ServiceCatalogItem({
   async function handleValidationErrors(response: Response) {
     const errorData: ServiceRequestResponse = await response.json();
     const invalidFieldErrors = errorData?.details?.base ?? [];
-    const missingErrorFields = invalidFieldErrors.filter(
+
+    // Log the raw error payload since the notification only shows a generic
+    // user-facing message, and there's no error-tracking service to capture it otherwise.
+    console.error("Service request validation failed:", errorData);
```
BrunoBFerreira: Having this console log is something we already do in other places? I know we can't log properly from here but would be interesting to confirm if this is an approach we should follow.

BrunoBFerreira (commented):
Just left some additional questions but nothing major.
Thank you for addressing the previous comments 🙇

src/modules/service-catalog/components/service-catalog-item/ServiceCatalogItem.spec.tsx
kristkvch-zen: Confirmed. It goes through more than `useValidateServiceItemForm`. That hook's validate() runs live here and gates handleRequestSubmit before submitServiceItemRequest is called. So mockRequestFields[0] needs a truthy value and a null would fail validation and block every test in that describe block before it reaches the server-error logic.

But there's a minor gap: no test covers ServiceCatalogItem blocking submission on a failed validation (only the validation logic itself is tested in isolation). Unrelated to this change - worth a follow-up or leave it?

BrunoBFerreira: You might want to check it out as a follow up then and see if it makes sense to fill that gap. Will mark this as resolved.

BrunoBFerreira (approved):
Thank you for addressing everything 🙏 
Before merging I would just check the usage of this string: `service-catalog.field-required-error`.
It appears to not be included in the translations. You might want to check it out and add it before this PR.

kristkvch-zen (approved):

PR #850, opened 2026-07-31
feat(templates): add applications_page with application_catalog helper

cakuki:
Adds `templates/applications_page.hbs` so the Applications catalog page can render via an independently-updatable app instead of requiring a native CPH build.

Keeps `{{breadcrumbs}}` in the same `.sub-nav` wrapper used by `service_list_page.hbs`/`category_page.hbs`. Calls a bare `{{application_catalog}}` helper with no local markup around it, so the app it renders can be updated without a CPH release — mirroring the existing `{{generative_answers}}` pattern in `search_results.hbs`.

This template alone has no effect until the corresponding backend helper is implemented and wired up. That work is tracked separately and isn't part of this PR.

Risks
risk:none

None: adds a new, unreferenced template file. No existing route maps to it yet, so this has zero effect on any live page until the backend-side wiring lands.

### DIFF
--- a/templates/applications_page.hbs
+++ b/templates/applications_page.hbs
@@ -1,0 +1,7 @@
+<div class="container">
+  <div class="sub-nav">
+    {{breadcrumbs}}
+  </div>
+
+  {{application_catalog}}
+</div>
--- a/templates/header.hbs
+++ b/templates/header.hbs
@@ -14,6 +14,7 @@
     <nav class="user-nav" id="user-nav" aria-label="{{t 'user_navigation'}}">
       <ul class="user-nav-list">
         <li>{{link 'community'}}</li>
+        <li>{{link 'applications'}}</li>
         <li>{{link 'service_catalog'}}</li>
         <li>{{link 'new_request' class='submit-a-request'}}</li>
         {{#unless signed_in}}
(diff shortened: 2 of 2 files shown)


cakuki:
@luis-almeida Added the Applications link to `header.hbs` (desktop + mobile nav), positioned after Community and before Service Catalog. No `{{#if}}` gating needed in the template — the link only renders when the feature is available for the account, same as the other conditionally-shown nav entries. Ready for another look.

cakuki:
Re: the failing Lighthouse check on this PR — confirmed it's pre-existing on `master`, unrelated to this PR's changes.

The single ❌ error (`meta-refresh` on `/hc/en-us/requests/new`, unrelated to `header.hbs`/`applications_page.hbs`) is byte-for-byte identical to failures already present on `master` itself, e.g. runs [30249004304](https://github.com/zendesk/copenhagen_theme/actions/runs/30249004304) and [30248902518](https://github.com/zendesk/copenhagen_theme/actions/runs/30248902518) from before this branch existed. This PR doesn't touch that template at all, so there's nothing in this diff to fix — will flag the underlying issue separately rather than patch around it here.

luis-almeida (approved):

PR #851, opened 2026-08-01
fix(service-catalog): preserve conditionally hidden fields on submission errors

petersonjdNIH:
## Description

Submitting the Service Catalog request form could permanently delete any ticket field currently hidden by an end-user condition from React state. Once dropped, the field never came back without a full page reload, so the form appeared to hang/fail and could become permanently unsubmittable.

**Root cause:** `useItemFormFields.tsx` stores the full field list in state (`allRequestFields`) but exposes only the currently-visible subset as `requestFields` (via `getVisibleFields(allRequestFields, endUserConditions)`). It hands callers `setRequestFields: setAllRequestFields` — the raw setter over the full list, under a name that reads like it operates on `requestFields`.

In `ServiceCatalogItem.tsx`, `handleValidationErrors()` (triggered on a 422 response) mapped over the visible-only `requestFields` and passed the result straight to `setRequestFields`. Since `requestFields` is only the visible subset, this call replaced the full state instead of merging into it — silently wiping any field hidden by a condition at that moment.

**Fix:** in `handleValidationErrors()`, merge onto the previous full list with a functional update instead of replacing it:

```ts
setRequestFields((prevFields) =>
  prevFields.map((field) => {
    const errorField = invalidFieldErrors.find(
      (errorField) => errorField.field_key === field.id
    );
    return { ...field, error: errorField?.description || null };
  })
);
```

Added coverage in `ServiceCatalogItem.spec.tsx`: asserts `setRequestFields` is called with a function, then invokes that function against a field list that includes one field not present in the mocked (visible) `requestFields`, asserting it's preserved unchanged in the result.

## Screenshots

N/A — this is a logic-only state management fix with no UI/markup changes.

## Checklist

- [x] :arrow_left: changes are compatible with RTL direction (no markup/UI changes)
- [x] :wheelchair: Changes to the UI are [tested for accessibility](./../README.md#accessibility-testing) and compliant with [WCAG 2.1](https://www.w3.org/TR/WCAG21/) (no UI changes)
- [x] :memo: changes are tested in Chrome, Firefox, Safari and Edge (no UI changes; covered by unit tests)
- [x] :iphone: changes are responsive and tested in mobile (no UI changes)

### DIFF
--- a/src/modules/service-catalog/components/service-catalog-item/ServiceCatalogItem.spec.tsx
+++ b/src/modules/service-catalog/components/service-catalog-item/ServiceCatalogItem.spec.tsx
@@ -502,13 +502,21 @@ describe("ServiceCatalogItem", () => {
       fireEvent.submit(form);
 
       await waitFor(() => {
-        expect(setRequestFields).toHaveBeenCalledWith([
-          expect.objectContaining({
-            id: 1,
-            error: "Product name: cannot be blank",
-          }),
-        ]);
+        expect(setRequestFields).toHaveBeenCalledWith(expect.any(Function));
       });
+
+      // validateForm's client-side check also calls setRequestFields
+      // (clearing stale errors) before handleValidationErrors does, so we
+      // need the last call, not the first.
+      const calls = setRequestFields.mock.calls;
+      const updateFn = calls[calls.length - 1][0];
+      const result = updateFn(mockRequestFields);
+      expect(result).toEqual([
+        expect.objectContaining({
+          id: 1,
+          error: "Product name: cannot be blank",
+        }),
+      ]);
     });
 
     const renderLastNotifyMessage = () => {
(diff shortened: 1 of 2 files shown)


kristkvch-zen:
Hi Joel — thanks for catching this, the fix is the right approach 🙏

Heads-up: we've recently merged [#842](https://github.com/zendesk/copenhagen_theme/pull/842), which touches the same block in `handleValidationErrors` and renamed the match key `field_key` → `field_id`. Your fix is still needed (the field-wiping bug is live on `master`), it just needs a quick rebase — keep your functional-update change, but match on `field_id` instead of `field_key`, and update the spec to match.

Thanks again for sending the fix back! One tip for next time: filing as a Support (Z2) ticket hits our on-call triage and gets eyes faster than GitHub notifs.

petersonjdNIH:
Thanks for the quick and thorough review, @kristkvch-zen. I appreciate you catching the overlap with #842! I've rebased on top of `master` and updated `handleValidationErrors()` (and the specs) to match on `field_id` instead of `field_key`, keeping the functional-update fix intact. The branch is rebased, tests are green, and the PR is ready for another look.

Noted on filing future issues as a Support (Z2) ticket for faster triage — will do next time!

kristkvch-zen:
Thank you!

kristkvch-zen (approved):

PR #856, opened 2026-08-17
fix(new-request-form): support lookup and multi_lookup field prefill via URL params

byronhabana:
## Summary
- `lookup` fields could only be prefilled via `?tf_<field_id>=<value>` by accident (falling through to the untyped default case), with no validation that the value was a real record ID.
- `multi_lookup` fields silently failed to prefill at all — `MultiLookupField` requires `field.value` to be a `string[]`, but the default case always assigned a plain string.
- Adds explicit `lookup`/`multi_lookup` cases that validate values as ULIDs before assigning, and split `multi_lookup` values into an array (comma-separated), mirroring the existing `multiselect` pattern.

## Test plan
- [x] `yarn test src/modules/new-request-form/usePrefilledTicketFields.test.tsx` — 20 passing tests covering standard fields, custom fields, multiselect, checkbox, date, partialcreditcard exclusion, lookup, multi_lookup, and the URL-length/`parent_id` bail-outs
- [x] `yarn eslint` clean on changed files
- [x] Manually verified in a locally-packaged theme upload: `?tf_<field_id>=<ULID>` populates a lookup field, and `?tf_<field_id>=<ULID1>,<ULID2>` populates a multi-lookup field (including the existing `max_selections` warning triggering correctly when over the limit)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/modules/new-request-form/usePrefilledTicketFields.test.tsx
+++ b/src/modules/new-request-form/usePrefilledTicketFields.test.tsx
@@ -1,0 +1,381 @@
+import { renderHook } from "@testing-library/react-hooks";
+import type { TicketFieldObject } from "../ticket-fields/data-types/TicketFieldObject";
+import { usePrefilledTicketFields } from "./usePrefilledTicketFields";
+
+const baseHref = "http://localhost/hc/en-us/requests/new";
+
+const mockLocation = (search: string) => {
+  window.history.replaceState(null, "", `${baseHref}${search}`);
+};
+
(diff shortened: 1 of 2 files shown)


hayden-gray-zd (approved):

hayden-gray-zd (approved):

Fredx87 (approved):
Thanks for adding tests covering also the existing functionality

PR #857, opened 2026-08-17
feat: sticky filters across reloads and tabs

melzreal:
## Description
- Persists applied filters even if page is reloaded or tab is changed
- Ensures filters are persisted, applying new filters overrides the ones in memory, moving back in forth in the browser pops the state appropriately

before

https://github.com/user-attachments/assets/de5c4e17-c500-4693-8fe4-13d4ba324fd6

after

https://github.com/user-attachments/assets/ace3e228-db26-4aaf-9195-b147336de714

## Checklist


src/modules/request-list/components/requests-list/RequestsList.tsx
```diff
@@ -31,20 +33,27 @@ export interface RequestsListProps {
...
 export function RequestsList({
   locale,
   customStatusesEnabled,
   viewRequestsAcrossBrandsEnabled,
 }: RequestsListProps): JSX.Element {
   const { t } = useTranslation();
 
+  const [storedFilters, setStoredFilters] = useLocalStorage<FilterValuesMap>(
+    FILTERS_LOCAL_STORAGE_KEY,
+    {}
```
melzreal: we can omit version but v1 shows in tests since useLocalStorage sets it by default. Not sure what the purpose is for version in this case, maybe this should be v4 as its a v4 theme, but its the first iteration of the storage function so have left it empty which defaults to v1

bin/lighthouse/config.js
```diff
@@ -73,6 +73,12 @@ module.exports = {
           selector: "nav > ol.breadcrumbs > li > a",
         },
       ],
+      "meta-refresh": [
```
Fredx87: This is not related to this PR, but especially it is not the right fix. Lighthouse is complaining that there is a `<meta http-equiv="refresh">` tag in the New Request Page, but if you look at the code, there is none. This means that Lighthouse is not testing the real page, but there is something that prevents it to reach the real page, I suspect Cloudflare Bot Management.

If we set this ignore rule, it means that we will let Lighthouse audit the wrong page

src/modules/request-list/components/requests-list/RequestsList.tsx
Fredx87: The purpose is to store a versioned value. If in the future we want to change the shape of the stored value, we can update the version to `v2` to avoid mixing values with different shapes

src/modules/request-list/utils/isFilterValuesMap.ts
```diff
@@ -0,0 +1,13 @@
...
+import { isFilterValue } from "./deserializeRequestListParams";
+
+export function isFilterValuesMap(value: unknown): value is FilterValuesMap {
+  if (typeof value !== "object" || value === null || Array.isArray(value)) {
+    return false;
+  }
+
+  return Object.values(value).every(
+    (filterValues) =>
+      Array.isArray(filterValues) && filterValues.every(isFilterValue)
```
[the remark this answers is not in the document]
melzreal: added a string check to isFilterValue and respective tests 👍

Fredx87 (commented):
With this change, filters are stored in two places: the URL and local storage. This is fine in the sense that the state in the URL can be used for bookmarking and history management, but I think we need a clearer synchronization model and a single source of truth while the page is open.

Currently, local-storage filters are used for the initial state, and URL filters are applied later by the `useParams` hook. When the URL contains filters, they replace the effective filter map, but they are not written back to local storage. This means the filters displayed in the UI and the filters persisted for the next visit can differ.

 The reverse is also a problem: when the page restores filters from local storage and the URL contains no filters, the restored filters are displayed and used for the request, but the URL is not updated. As a result, the URL no longer represents the current UI state, so copying or bookmarking it will not preserve the filters the user is currently seeing.

 Could we centralize URL/local-storage hydration and synchronization, with a clearly defined precedence? For example, URL filters could win when explicitly present, while local storage is used only as a fallback. After resolving the effective filters, they should be reflected in the URL, using `replaceState` during initial hydration to avoid adding a history entry, and persisted to local storage where appropriate.

Side note, potentially outside the scope of this PR: `useParams` writes history entries using `pushState` but does not listen for `popstate`, so browser Back/Forward can also change the URL without updating the displayed filters and results.

melzreal:
setting back to draft to address the unification of params with localstorage

bin/lighthouse/config.js
melzreal: how should we go around it, instead of having a red pipeline until edge fixes the issue? should we temporarily ignore the /requests/new ? have we raised with edge yet?

melzreal: run an analysis and the AI seems to confirm the Cloudflare is blocking the page by serving an overlay, excluded requests/new from being audited by lighthouse temporarily - not sure what else we could do on our side to prevent a broken pipeline - maybe there is a cloudflare exclusion list somewhere 🤔

src/modules/request-list/components/requests-list/RequestsList.tsx
```diff
@@ -38,17 +29,7 @@ export function RequestsList({
 }: RequestsListProps): JSX.Element {
   const { t } = useTranslation();
 
-  const { params, push } = useParams<RequestListParams>(
-    {
-      query: "",
-      page: 1,
-      sort: { order: "desc", by: "updated_at" },
```
melzreal: note: this test was moved to [src/modules/request-list/hooks/useRequestListParams.test.ts](https://github.com/zendesk/copenhagen_theme/pull/857/changes#diff-5641962c1294000363df0ac6ec12d44b42922a38a5c40ee4e3debb46d19a0f97R141)

src/modules/request-list/components/requests-list/RequestsList.test.tsx
```diff
@@ -123,18 +123,6 @@ describe("Rendering", () => {
     ).toBeInTheDocument();
   }, 10000);
 
-  test("applies a default sort of updated at and descending", async () => {
```
melzreal, line 126: note: this test was moved to [src/modules/request-list/hooks/useRequestListParams.test.ts](https://github.com/zendesk/copenhagen_theme/pull/857/changes#diff-5641962c1294000363df0ac6ec12d44b42922a38a5c40ee4e3debb46d19a0f97R141) as the default params setting was moved to the [hook](https://github.com/zendesk/copenhagen_theme/pull/857/changes?diff=split#diff-8c89dad8d1cc05582bd9a1195aa60a62af6acb3a30dc6d1156d99cbe6f2002faR20)

src/modules/request-list/utils/deserializeRequestListParams.ts
```diff
@@ -71,6 +71,18 @@ export function deserializeRequestListParams(
   return res;
 }
 
+const RECOGNIZED_KEYS: string[] = Object.values(SERIALIZED_KEYS);
```
melzreal, line 74: serialized keys are defined [here](https://github.com/zendesk/copenhagen_theme/blob/melreal/sticky-filters/src/modules/request-list/utils/serializeRequestListParams.ts#L5)

src/modules/request-list/utils/isFilterValuesMap.ts
```diff
@@ -0,0 +1,13 @@
+import type { FilterValuesMap } from "../data-types/FilterValue";
+import { isFilterValue } from "./deserializeRequestListParams";
+
```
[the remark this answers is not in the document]
melzreal: reusing it for consistency, and if it was rewritten it does more then the suggested code

gosiexon-zen (dismissed):
Looks good, I've checked that you applied requested changes from the Gianluca's comment. Now useParams become a bit complex but looks like it needs to be that way :)

src/modules/request-list/hooks/useRequestListParams.ts
```diff
@@ -0,0 +1,97 @@
...
+    localStorage.setItem(
+      FILTERS_LOCAL_STORAGE_KEY,
+      JSON.stringify([FILTERS_LOCAL_STORAGE_VERSION, filters])
+    );
+  } catch (error) {
+    // ignore
+  }
+}
+
+export function useRequestListParams(): UseRequestListParamsResult {
```
luis-almeida: I understand the split of `useParams` but I think the cons outweigh the pros.
- Without looking at the code it's really hard to understand why is there as `useRequestListParams` and a `useParams`
- The whole point with `useParams` was to have a single hook to manage params. With two there's always a risk `useParams` may also be used elsewhere separately
- This change will need to be ported to the other code-bases. A single file change would be much easier to port.

src/modules/request-list/utils/isFilterValuesMap.ts
```diff
@@ -0,0 +1,13 @@
...
+export function isFilterValuesMap(value: unknown): value is FilterValuesMap {
```
luis-almeida: Couldn't thin be inlined since it's only used in one place?
It's also confusing to me that the other util then exports a `isFilterValue` function which then just seems misplaced and getting entangled.

src/modules/request-list/hooks/useRequestListParams.ts
```diff
@@ -0,0 +1,97 @@
...
+  const { params, push: pushParams } = useParams<RequestListParams>(
+    DEFAULT_REQUEST_LIST_PARAMS,
+    serializeRequestListParams,
+    deserializeRequestListParams,
+    { resolve, replaceUrlOnMount: true, syncOnPopState: true }
+  );
+
+  const push = useCallback(
+    (newParams: Partial<RequestListParams>) => {
+      if (newParams.filters) {
```
luis-almeida: Could there be changes "pushed" without filters?
If so those would not be stored and could potentially lead to mismatch.
Why not just always write all params to store?

melzreal: given the story is about "sticky filters" I've scoped the storage to them - I worry what would happen if there was also a search query, we'd be pushing that to storage? 

with filters we also have a narrowly defined set - we can check against specific strings and safeguard we are not storing something outside of those formats.

so storing just filters is intended, it means a smaller change to meet the criteria which is safer - I fear if we store all params we could be expanding the scope of this change into consequences we haven't considered

on the conditional write I'll make the change so push always persists the merged filters, that way storage can't drift even for pushes that don't touch filters

luis-almeida (approved):
Thanks for addressing the feedback!

PR #863, opened 2026-08-26
fix: lighthouse workflow run on new request page

melzreal:
- The edge team recommended using the self-hosted runner to ensure Cloudflare will not display
an overlay on the page and prevent the a11 checks
from running.

- Reverses https://github.com/zendesk/copenhagen_theme/pull/861/changes with the recommended fix.

## Checklist


melzreal:
public workflows cannot run on  [self-hosted, zendesk-stable] so advice from security does not apply here - essentially its not possible to re enable the check for now

